PR #2, opened 2015-02-10
Add testing support package

kortschak:
PTAL @eaburns

### DIFF
--- a/vg/recorder/recorder.go
+++ b/vg/recorder/recorder.go
@@ -1,0 +1,476 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package recorder provides support for vector graphics serialization.
+package recorder
+
+import (
+	"fmt"
+	"image/color"
+	"runtime"
+
+	"github.com/gonum/plot/vg"
+)
+
+var _ vg.Canvas = (*Canvas)(nil)
+
+// Canvas implements vg.Canvas operation serialization.
+type Canvas struct {
+	// Resolution holds the canvas resolution in DPI.
+	Resolution float64
+
+	// Actions holds a log of all methods called on
+	// the canvas.
+	Actions []Action
+
+	// KeepCaller indicates whether the Canvas will
+	// retain runtime caller location for the actions.
+	// This includes source filename and line number.
+	KeepCaller bool
+
+	// c holds a backing vg.Canvas. If c is non-nil
+	// then method calls to the Canvas will be
+	// reflected to c.
+	c vg.Canvas
+
+	// fonts holds a collection of font/size descriptions.
+	fonts map[fontID]vg.Font
+}
+
+type fontID struct {
+	name string
+	size vg.Length
+}
+
+// Action is a vector graphics action as defined by the
+// vg.Canvas interface. Each method of Canvas has a
+// corresponding Action type.
+type Action interface {
+	Call() string
+	ApplyTo(vg.Canvas)
+	callerLocation() *callerLocation
+}
+
+type callerLocation struct {
+	haveCaller bool
+	file       string
+	line       int
+}
+
+func (l *callerLocation) set() {
+	_, l.file, l.line, l.haveCaller = runtime.Caller(3)
+}
+
+func (l callerLocation) String() string {
+	if !l.haveCaller {
+		return ""
+	}
+	return fmt.Sprintf("%s:%d ", l.file, l.line)
+}
+
+// New returns a new Canvas with the specified resolution.
+func New(dpi float64) *Canvas { return &Canvas{Resolution: dpi} }
+
+// NewFrom returns a new Canvas from an existing vg.Canvas. vg.Canvas methods
+// called on the Canvas will also be passed to the provided backing
+// vg.Canvas. The Resolution field is set to the value returned by c.DPI.
+func NewFrom(c vg.Canvas) *Canvas { return &Canvas{Resolution: c.DPI(), c: c} }
+
+// Reset resets the Canvas to the base state.
+func (c *Canvas) Reset() {
+	c.Actions = c.Actions[:0]
+}
+
+// ReplayOn applies the set of Actions recorded by the Recorder onto
+// the destination Canvas.
+func (c *Canvas) ReplayOn(dst vg.Canvas) error {
+	if c.fonts == nil {
+		c.fonts = make(map[fontID]vg.Font)
+	}
+	for _, a := range c.Actions {
+		fa, ok := a.(*FillString)
+		if !ok {
+			continue
+		}
+		f := fontID{name: fa.Font, size: fa.Size}
+		if _, exists := c.fonts[f]; !exists {
+			var err error
+			c.fonts[f], err = vg.MakeFont(fa.Font, fa.Size)
+			if err != nil {
+				return err
+			}
+		}
+		fa.fonts = c.fonts
+	}
+	for _, a := range c.Actions {
+		a.ApplyTo(dst)
+	}
+	return nil
+}
+
+func (c *Canvas) append(a Action) {
+	if c.c != nil {
+		a.ApplyTo(c)
+	}
+	if c.KeepCaller {
+		a.callerLocation().set()
+	}
+	c.Actions = append(c.Actions, a)
+}
+
+// SetLineWidth corresponds to the vg.Canvas.SetWidth method.
+type SetLineWidth struct {
+	Width vg.Length
+
+	l callerLocation
+}
+
+// SetLineWidth implements the SetLineWidth method of the vg.Canvas interface.
+func (c *Canvas) SetLineWidth(w vg.Length) {
+	c.append(&SetLineWidth{Width: w})
+}
+
+// Call returns the method call that generated the action.
+func (a *SetLineWidth) Call() string {
+	return fmt.Sprintf("%sSetLineWidth(%v)", a.l, a.Width)
+}
+
+// ApplyTo applies the action to the given vg.Canvas.
+func (a *SetLineWidth) ApplyTo(c vg.Canvas) {
+	c.SetLineWidth(a.Width)
+}
+
+func (a *SetLineWidth) callerLocation() *callerLocation {
+	return &a.l
+}
+
+// SetLineDash corresponds to the vg.Canvas.SetLineDash method.
+type SetLineDash struct {
+	Dashes  []vg.Length
+	Offsets vg.Length
+
+	l callerLocation
+}
+
+// SetLineDash implements the SetLineDash method of the vg.Canvas interface.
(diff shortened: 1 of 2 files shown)


kortschak:
This is actually looking more like a general purpose vector graphic serialisation protocol intermediate. I think I'm OK with that thought.

vg/recorder/recorder.go
```diff
@@ -0,0 +1,458 @@
...
+)
+
+var _ vg.Canvas = (*Canvas)(nil)
+
+// Canvas implements vg.Canvas operation serialization.
+type Canvas struct {
+	// Resolution holds the canvas resolution in DPI.
+	Resolution float64
+
+	// Actions holds the complete state of the canvas.
```
eaburns: Perhaps: // Actions holds a log of all methods called on the canvas.

vg/recorder/recorder.go
```diff
@@ -0,0 +1,458 @@
...
+	Actions []Action
+
+	// KeepCaller indicates whether the Canvas will
+	// retain caller information for the actions.
```
eaburns: Say what "caller information" is.

vg/recorder/recorder.go
```diff
@@ -0,0 +1,458 @@
...
+// Action is a vector graphics action as defined by the
+// vg.Canvas interface. Each method of vg.Canvas must have
+// a corresponding Action type.
+type Action interface {
+	VgCall() string
+	ApplyTo(vg.Canvas)
+	callRecorder() *callRecorder
+}
+
+type callRecorder struct {
```
eaburns: I'd call it a callRecord since it's a record of a call, not a thing that records calls. I may even just call it a record.

vg/recorder/recorder.go
```diff
@@ -0,0 +1,458 @@
...
+}
+
+func (c callRecorder) String() string {
+	if !c.haveCaller {
+		return ""
+	}
+	return fmt.Sprintf("%s:%d ", c.file, c.line)
+}
+
+// New returns a new Canvas pre-filled with the given base actions.
```
eaburns: I don't understand this. It seems to return a new recorder.Canvas without anything pre-filled.

vg/recorder/recorder.go
```diff
@@ -0,0 +1,458 @@
...
+func (c *Canvas) Reset() {
+	c.Actions = c.Actions[:0]
+}
+
+// ReplayOn applies the set of Actions recorded by the Recorder onto
+// the destination Canvas.
+func (c *Canvas) ReplayOn(dst vg.Canvas) error {
+	fonts := make(map[fontID]vg.Font)
+	for _, a := range c.Actions {
+		if fa, ok := a.(*FillString); ok {
```
eaburns: I prefer to reduce this deep indentation like this:
fa, ok := a.(*FillString)
if !ok {
  continue
}

Up to you.

vg/recorder/recorder.go
```diff
@@ -0,0 +1,458 @@
...
+	callRecorder() *callRecorder
+}
+
+type callRecorder struct {
+	haveCaller bool
+	file       string
+	line       int
+}
+
+func (c *callRecorder) caller() {
```
eaburns: This method name is a noun. When I read it at a call site I expect it to return something, not modify state. Perhaps it should be setCaller?

vg/recorder/recorder.go
```diff
@@ -0,0 +1,458 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package recorder provides support for vector graphics serialization
+// it is intended to be used as a testing helper.
+package recorder
```
eaburns: Recorder is fine, but I slightly prefer calling it a log---it logs the actions performed on it. This does conflict with the standard package log, but Effective Go agrees that we shouldn't worry:
"And don't worry about collisions a priori. The package name is only the default name for imports; it need not be unique across all source code, and in the rare case of a collision the importing package can choose a different name to use locally."

vg/recorder/recorder.go
```diff
@@ -0,0 +1,458 @@
...
+	"image/color"
+	"runtime"
+
+	"github.com/gonum/plot/vg"
+)
+
+var _ vg.Canvas = (*Canvas)(nil)
+
+// Canvas implements vg.Canvas operation serialization.
+type Canvas struct {
```
eaburns: Not sure if this would ever be useful, but this could also have an underlying vg.Canvas to which it forwards calls. That way, it can log calls while actually drawing to another canvas too. If the underlying canvas is nil then it simply doesn't call its methods.

eaburns:
Github doesn't have the best code review interface, apparently. It would be nice if it let me write all my comments and send them all at once instead of sending each individually. That way, I could comment as I go and go back to fix previous comments as I read on.

Anyway, I can't help but wonder if we can greatly simplify this using reflection. I haven't thought about it much, but it seems like you could at least ditch the VgCall method on each Action and instead record the method name and its parameter reflect.Values. If so, then you can probably just record that information in the callRecord and ditch actions all together.

You expressed interest in serializing. Are reflect.Values serializable?

vg/recorder/recorder.go
eaburns: Yes, I think that I prefer record, but maybe it's just because I think that camlCaseLooksUgly, so I try hard to find single-word names.

vg/recorder/recorder.go
```diff
@@ -0,0 +1,458 @@
...
+	// KeepCaller indicates whether the Canvas will
+	// retain caller information for the actions.
+	KeepCaller bool
+}
+
+// Action is a vector graphics action as defined by the
+// vg.Canvas interface. Each method of vg.Canvas must have
+// a corresponding Action type.
+type Action interface {
+	VgCall() string
```
eaburns: How about Call() and Apply()?
It's obviously a "Vg" call since this part of vg. And I think that it's clear that it's going to apply the action to the single argument available to it.

vg/recorder/recorder.go
kortschak: The reason I think recorder is that a log does not play back. This package does that.

vg/recorder/recorder.go
kortschak: I'm glad you brought that up. I was thinking the same thing - NewFrom(c vg.Canvas) *Canvas, which also sets Resolution.

vg/recorder/recorder.go
kortschak: Call is OK, but Apply needs the direction, otherwise it reads as Apply the vg.Canvas.

vg/recorder/recorder.go
kortschak: I'll try and find something better than callRecorder; record is not informative IMO.

vg/recorder/recorder.go
kortschak: Forgot to remove it from previous design - the old API that I use in biogo.graphics retains a special set of initial actions. This design doesn't - this is much nicer.

kortschak:
This is a little repetitious, but I think easier to deal with. In order to deal with the translation of calls to fields, I'd need to keep a map between the two. This is transparent and simple.

vg/recorder/recorder.go
kortschak: Done.

kortschak:
PTAL.

wrt serialisation, If you gob/json encode a Canvas, you can send a plot over the wire - probably far more bandwith efficiently that via svg, ... this was part of the reason for keeping fields public, and encoding font as a string and size.

vg/recorder/recorder.go
kortschak: Done.

kortschak:
@eaburns is this OK to go in?

eaburns:
I was going to look again this weekend. Let me take a peek right now.

vg/recorder/recorder.go
eaburns: OK.

vg/recorder/recorder.go
```diff
@@ -0,0 +1,483 @@
...
+	// the canvas.
+	Actions []Action
+
+	// KeepCaller indicates whether the Canvas will
+	// retain runtime caller location for the actions.
+	// This includes source filename and line number.
+	KeepCaller bool
+
+	// c holds a backing vg.Canvas. Method calls to
+	// the Canvas will be reflected to c.
```
eaburns: If c is non-nil then method calls to...

vg/recorder/recorder.go
```diff
@@ -0,0 +1,483 @@
...
+	fonts map[fontID]vg.Font
+}
+
+type fontID struct {
+	name string
+	size vg.Length
+}
+
+// Action is a vector graphics action as defined by the
+// vg.Canvas interface. Each method of vg.Canvas must have
```
eaburns: The second sentence sounds like an implementation detail. How about, "Each method of Canvas has a corresponding Action type."?

vg/recorder/recorder.go
```diff
@@ -0,0 +1,483 @@
...
+		Resolution: dpi,
+	}
+}
+
+// New returns a new Canvas from an existing vg.Canvas. vg.Canvas methods
+// called on the Canvas will also be passed to the provided backing
+// vg.Canvas.
+func NewFrom(c vg.Canvas) *Canvas {
+	return &Canvas{
+		Resolution: c.DPI(),
```
eaburns: You forgot to set c.

eaburns: Also note in the comment that the DPI of the returned canvas is inherited from c.

vg/recorder/recorder.go
```diff
@@ -0,0 +1,483 @@
...
+// Action is a vector graphics action as defined by the
+// vg.Canvas interface. Each method of vg.Canvas must have
+// a corresponding Action type.
+type Action interface {
+	Call() string
+	ApplyTo(vg.Canvas)
+	call() *location
+}
+
+type location struct {
```
eaburns: caller? callerLocation?

vg/recorder/recorder_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+	replay.Reset()
+	rec.Actions = append(rec.Actions, &FillString{Font: "Foo", Size: 12, X: 0, Y: 10, String: "Bar"})
+	err = rec.ReplayOn(&replay)
+	if !strings.HasPrefix(err.Error(), "Unknown font: Foo.") {
+		t.Errorf("unexpected error: %v", err)
+	}
+}
+
+var want = []string{
```
eaburns: I don't think that this hits every method, does it?
It probably should just to verify that there aren't any big, obvious bugs in one of the Actions.

kortschak: It does now.

kortschak:
OK, all addressed.

eaburns:
LGTM

kortschak:
I don't know what happened with the history here. This PR refers to 395c9bf342ceb823c669af40f59a1ec0bccda814.

PR #3, opened 2015-02-10
plotutil: fix build

fhs:


kortschak:
Fix imports on L10-12 then LGTM.

fhs:
Imports need to be fixed in a lot of other files. I'll send a different PR for import fixes.

kortschak:
OK SGTM

PR #4, opened 2015-02-10
all: update import paths

fhs:
Also run goimports on all files.

### DIFF
--- a/vg/vgsvg/vgsvg.go
+++ b/vg/vgsvg/vgsvg.go
@@ -9,12 +9,13 @@ package vgsvg
 import (
 	"bufio"
 	"bytes"
-	"code.google.com/p/plotinum/vg"
 	"fmt"
-	svgo "github.com/ajstarks/svgo"
 	"image/color"
 	"io"
 	"math"
+
+	svgo "github.com/ajstarks/svgo"
+	"github.com/gonum/plot/vg"
 )
 
 const (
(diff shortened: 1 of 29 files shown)


vg/vgsvg/vgsvg.go
```diff
@@ -9,12 +9,13 @@ package vgsvg
...
 	"bufio"
 	"bytes"
-	"code.google.com/p/plotinum/vg"
 	"fmt"
-	svgo "github.com/ajstarks/svgo"
 	"image/color"
 	"io"
 	"math"
+
+	svgo "github.com/ajstarks/svgo"
```
kortschak: Maybe in a follow-up drop the svgo import rename and use svg? There is no collision.

kortschak:
LGTM

vg/vgsvg/vgsvg.go
fhs: I'm not so sure about this change. It might hurt readability since there is a struct field named svg. I'll defer it to the original author of this code, which looks like @eaburns

kortschak: No worries.

eaburns: I prefer svgo since that's the last element of the import path and well-behaved go packages use the last element name as their package name. I consider it a bug that the package name isn't actually svgo.

kortschak: Yehs, that's what got me.

eaburns: This has come up many times in the past. We need to add a comment explaining why we do this.

PR #8, opened 2015-02-11
fix COVERALLS_TOKEN secure environment variable

jonlawlor:
The wrong environment variable was set, which prevented coveralls.io
from showing a coverage badge.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ go:
 
 env:
   global:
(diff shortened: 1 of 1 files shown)


jonlawlor:
For the life of me I can't figure out why this isn't performing coverage.  I'll have to think about it.

jonlawlor:
I believe that travis doesn't allow secure environment variables here for some reason.  The only case that I know of is when trying to merge a fork; the .travis.yml file is basically identical to the other repos.

PR #9, opened 2015-02-11
Gonum redesign vgdraw

sbinet:
- add vgdraw.Canvas (in replacement of plot.DrawArea)
- introduce vg.Inch, vg.Centimeter and vg.Millimeter
- introduce plot.Ticker interface
- introduce plot.Normalizer interface
- support for persistency

### DIFF
--- a/axis.go
+++ b/axis.go
@@ -10,7 +10,22 @@ import (
 	"math"
 
 	"github.com/gonum/plot/vg"
+	"github.com/gonum/plot/vg/draw"
 )
+
+// Ticker creates Ticks in a specified range
+type Ticker interface {
+	// Ticks returns Ticks in a specified range
+	Ticks(min, max float64) []Tick
+}
+
+// Normalizer rescales values from the data coordinate system to the
+// normalized coordinate system.
+type Normalizer interface {
+	// Normalize transforms a value x in the data coordinate system to
+	// the normalized coordinate system.
+	Normalize(min, max, x float64) float64
+}
 
 // An Axis represents either a horizontal or vertical
 // axis of a plot.
@@ -24,11 +39,11 @@ type Axis struct {
 		Text string
 
 		// TextStyle is the style of the axis label text.
-		TextStyle
+		draw.TextStyle
 	}
 
 	// LineStyle is the style of the axis line.
-	LineStyle
+	draw.LineStyle
 
 	// Padding between the axis line and the data.  Having
 	// non-zero padding ensures that the data is never drawn
@@ -37,10 +52,10 @@ type Axis struct {
 
 	Tick struct {
 		// Label is the TextStyle on the tick labels.
-		Label TextStyle
+		Label draw.TextStyle
 
 		// LineStyle is the LineStyle of the tick lines.
-		LineStyle
+		draw.LineStyle
 
 		// Length is the length of a major tick mark.
 		// Minor tick marks are half of the length of major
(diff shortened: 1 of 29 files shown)


kortschak:
These commits don't seem to happen in order. From what I can see 3465f964e9aef09856a4aa6388a0a8d2e2263a44 should happen before dcf6a66a88c002c7400edb1d331d2f405fe2cc62, but it is presented later in the list. Do you know what is causing that?

sbinet:
some re-ordering might have been caused by the multiple cherry-picking + rebase-fixing/merge-conflict-fixing I had to apply.
that said, the ordered list presented in this PR seems like the correct timeline (at least, the one I recall.)

kortschak:
I'm following the graph. That makes more sense to me.

kortschak:
I think draw would be a better name than vgdraw, but I like these changes over all. Nicely presented too.

sbinet:
I wouldn't mind changing `vgdraw` into `draw` but it could shadow the stdlib package `image/draw` (and the upcoming `x/image/draw`)

sbinet:
PTAL

kortschak:
That should not be an issue. People can rename if they collide.

sbinet:
ok. done.

axis.go
```diff
@@ -196,29 +219,29 @@ func (a *horizontalAxis) draw(da DrawArea) {
...
-	for _, t := range a.Tick.Marker(a.Min, a.Max) {
+	for _, t := range a.Tick.Marker.Ticks(a.Min, a.Max) {
 		if t.IsMinor() {
 			continue
 		}
 		w := a.Tick.Label.Width(t.Label)
 		box := GlyphBox{
 			X:    a.Norm(t.Value),
-			Rect: Rect{Point{X: -w / 2}, Point{X: w}},
+			Rect: draw.Rect{vgdraw.Point{X: -w / 2}, vgdraw.Point{X: w}},
```
kortschak: Missed change. Others at axis.go:327, plot.go:182,207,263.

sbinet: Arf. Right. sed replaces only one occurrence per line.
Will fix and squash.

sbinet:
PTAL

vg/len.go
```diff
@@ -26,42 +13,19 @@ func Points(pt float64) Length {
...
-}
-
-// Millimeters returns a length for the given number of millimeters.
-func Millimeters(mm float64) Length {
-	return Length(mm * ptMillimeter)
-}
+// Common lengths.
+const (
+	Inch       Length = 72
+	Centimeter        = Inch / 2.54
```
kortschak: Missed this, but it's spelled in American in gonum/unit, so we should leave it.

kortschak:
LGTM, but waiting for others.

sbinet:
@eaburns I suspect (hope!) this PR already has your LGTM, right? :)

sbinet:
also: I should probably squash before this gets merged.

eaburns:
LGTM, assuming that it's the same as before. I didn't look at it again (sorry, lots of action on this package at the moment, and I'm trying to squeeze in my reviewing before heading to work).
I agree that you should squash it down before merging.
The commit that adds you to the AUTHORS file should be a separate commit. Don't squash that.
In the near future we will remove this separate AUTHORS file anyway, since there is a global gonum AUTHORS file elsewhere.

sbinet:
yes, it is the same.
I'll squash as you advised.
(I am already in the gonum C+A, IIRC)

sbinet:
squashed and polished.

if nobody beats me to (or disagrees) I'll merge this over the week-end.

PR #10, opened 2015-02-11
Import biogo.graphics palettes

kortschak:
Minor changes to packages in that repo:

* Change from gocheck to testing.
* golint changes.

All AUTHORS of this code are already in gonum/A+C (me and my employer). biogo copyright notices are retained purely for historical reasons.


palette/brewer/brewer.go
```diff
@@ -0,0 +1,174 @@
...
+// The colors defined here are from http://www.ColorBrewer.org/ by Cynthia A. Brewer,
+// Geography, Pennsylvania State University.
+//
+// For more information see:
+// http://www.personal.psu.edu/cab38/ColorBrewer/ColorBrewer_learnMore.html
+//
+package brewer
+
+import (
+	"code.google.com/p/biogo.graphics/palette"
```
sbinet: shouldn't this be `github.com/gonum/plot/palette` ?

kortschak: It should.

kortschak:
PTAL

palette/hsva_test.go
```diff
@@ -0,0 +1,83 @@
...
+import (
+	"image/color"
+	"testing"
+)
+
+func withinEpsilon(a, b, epsilon uint32) bool {
+	d := int64(a) - int64(b)
+	if d < 0 {
+		if d == -d {
+			panic("color: weird number overflow")
```
sbinet: `s/color:/palette:/`
(or `plot`?)

sbinet:
perhaps also add a simple example?

other than that: LGTM

kortschak:
I can't think of any non-facile example, so I'll push this. If you want I can revisit this after discussion.

eaburns:
Is it appropriate to use this in plotutil to replace the various color slices? If so, open an issue so that we don't forget.

kortschak:
That can be done. I'll send a PR today.

kortschak:
Actually, this requires further discussion since there are at least a couple of different approaches, so #11.

PR #12, opened 2015-02-11
palette: add DivergingPalette interface

kortschak:
Formalise the notion already existing in the brewer package.

### DIFF
--- a/palette/palette.go
+++ b/palette/palette.go
@@ -19,6 +19,18 @@ type Palette interface {
 	Colors() []color.Color
 }
 
+// DivergingPalette is a collection of colors ordered into a palette with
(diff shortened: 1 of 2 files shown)


eaburns:
I don't quite understand what a divergent pallet is; probably because I don't know what a critical index is. Mind giving a tiny bit of background?

kortschak:
Where you have values that diverge from a zero/median point a diverging palette is useful. The critical index is the index into the palette for the colour(s) that represent that zero/median value.

kortschak:
This will help (linked in the docs) http://colorbrewer.org/

eaburns:
Got it! Thanks.
LGTM.

Issue #14, opened 2015-02-12
plot: make GlyphBox support plot area-dependent sized glyphs


kortschak:
At the moment, a plot.GlyphBox represents a zero-sized point in the data space surrounded by an pair of offsets in the plot space. There is no way to represent the extent of glyphs that have a size dependent on the size of the plot.

This is due to the fact that the GlyphBoxes method is not passed any information regarding the plot area.

This can be fixed by adding a second pair of normalised x,y coordinates.

```
type GlyphBox struct {
    // The glyph location in normalized coordinates.
    MinX, MinY float64
    MaxX, MaxY float64

    // Rect is the offset of the glyph's minimum drawing
    // point relative to the glyph location and its size.
    Rect
}
```

The details of types etc need to be fleshed out.

PR #15, opened 2015-02-12
plotter: add HeatMap and Contour implementation

kortschak:


plotter/heat.go
```diff
@@ -0,0 +1,144 @@
...
+// DataRange implements the DataRange method
+// of the plot.DataRanger interface.
+func (h *HeatMap) DataRange() (xmin, xmax, ymin, ymax float64) {
+	r, c := h.Matrix.Dims()
+	return -0.5, float64(c) + 0.5, -0.5, float64(r) + 0.5
+}
+
+// GlyphBoxes implements the GlyphBoxes method
+// of the plot.GlyphBoxer interface.
+func (h *HeatMap) GlyphBoxes(plt *plot.Plot) []plot.GlyphBox {
```
kortschak: This does not currently do what I want for this type, but is blocked on #14.

plotter/main.go
```diff
@@ -863,6 +866,30 @@ func Example_stackedBarChart() *plot.Plot {
...
+
+	p, err := plot.New()
+	if err != nil {
+		panic(err)
+	}
+	p.Title.Text = "Heat map"
+
+	p.Add(h)
+
+	p.X.Padding = 0
```
kortschak: It would be nice to have an automated way of dealing with this.

jonlawlor: I can't say that I am terribly familiar with the context, but the insanity of matlab plots has taught me that the answer is probably functional programming via delegates.  What you want is a `func(*plot.Plot)` which deals with the formatting and can be easily chained.  The "undo" functionality that Pike uses would be a nice add.

kortschak: Yes, I mentioned this in the gonum/plot design discussion.

eaburns:
Not sure who's supposed to be reviewing this. We should probably assign people so that it's clear. If it's supposed to be me then I will probably have to wait until this weekend so that I can really look at it. Hope that's OK.

kortschak:
That's fine. I have a partner for it which may go in the same PR - contours. I'm happy to accept any reviewers though.

kortschak:
This is now a WIP.

Running `go test` generates a small svg file that shows what it looks like so far. TODO is actually making the Contour plotter and sorting out level shading (the majority of the contour.go code is needed for that).

@vladimir-ch I have the GridMesh implementation and I can change this to a GridFunc by rearranging the r/c order. PTAL and let me know.

kortschak:
I've decided the GridFunc approach is nicer and more general. The cost of plugging a matrix in is not onerous and we can provide a helper type if needed.

egonelbre:
I did a quick mockup of a way to avoid list.List and code that should be simpler to understand and faster: http://play.golang.org/p/YiLMQ2tdIP.

kortschak:
Thanks. How well does it cope with loop excision? This was the major reason for using lls. I can see that it's just alice tricks, but depending on the data, potentially many.

I think you should be returning point on lines 18 and 19.

kortschak:
It seems to me contour could be more efficiently just two fields of []point. newContour then becomes `return contour{[]point{l.p1}, []point{l.p2}}`.

egonelbre:
Indeed using points instead of lines makes it much simpler. This is my first time writing this kind of code, I'm not quite sure what you mean by "loop excision" -- although I would expect that it can be incorporated.

Current progress so far: https://gist.github.com/egonelbre/5770469d6d7b955a4559. It still has some bugs, but can output something that looks like a contour graph :D.

kortschak:
I had a working version from your play snippet and then I browsed away - I hate mobile devices.

Loop excision is the code at contour.go#212. It is possible to build a path that crosses itself, but if not recognisable as a loop. This is not important at this stage but will be when we want to implement filling.

egonelbre:
Finally have a version where I'm unable to tell the difference visually, although there is some differences... https://gist.github.com/egonelbre/5770469d6d7b955a4559

I'm not sure what to do with the lines that connect to a contour in the middle?

kortschak:
I've added a benchmark for contourPaths and relaxed the test for
contourPaths to accept vg.Paths in either direction for the test data.

kortschak:
I've factored out the operations except for the vg.Path construction. This should make it easier to port across to slices.

vladimir-ch:
I like the GridFunc interface, it is very natural. Just a comment/thought: as it is now (with the Z() method), the interface suggests that it can accomodate more general, parametric surfaces, not only graphs of functions. Perhaps renaming Z() to F() would be better (for heat maps and contour plots) and, depending on the ambitions of the plot package, we could reserve the Z() method for a more general Surface interface.

kortschak:
>  as it is now (with the Z() method), the interface suggests that it can accomodate more general, parametric surfaces, not only graphs of functions

It does handle these.  The volcano example shows that.

btracey:
I wonder if we should rename GridFunc. Throughout gonum (so far at least) a "function" is one of the reals. This is not only in places like optimize, but also in plot itself, i.e. plotter.Function. I would recommend changing to something like GridXYZ, as it's a thing that can return XYZ values on a grid. The natural extension types are DenseGrid, which is a GridXYZ that is backed by a mat64.Dense of values, and GridFunc, which can be constructed with something like

````
GridFunc(f func(float64, float64), xBounds, yBounds [2]float64, sizeX, sizeY int) GridXYZ
````
that allows easy wrapping of 2-D functions.

**Comment by btracey** [comment]
Well, the lazy evaluation is un go-like, so probably GridFunc should compute the values of the function and return the DenseGrid (or something).

**Comment by kortschak** [comment]
I'm happy to call it GridXYZ.

**Comment by kortschak** [comment]
I don't want to add GridFunc or DenseGrid. They are cheap to write and personally, I prefer lazy evaluation in a case like this.

**Comment by vladimir-ch** [comment]
I also prefer to call it GridXYZ, it can represent other surfaces than just function graphs (sorry I overlooked the volcano example before).

**Comment by kortschak** [comment]
I think this can be open for a more complete review now.

When it gets the go-ahead, I'll squash to 32d1c20f39d43b48a48fca0700f4370b7ca1d737 and amend that commit message to include the linked list to slice implementation improvement. I think that is worth keeping in the log.

PTAL

File: plotter/heat.go [hunk 3]
```diff
@@ -0,0 +1,210 @@
...
+	"github.com/gonum/plot/palette"
+	"github.com/gonum/plot/vg"
+	"github.com/gonum/plot/vg/draw"
+)
+
+// GridXYZ describes three dimensional data where the X and Y
+// coordinates are arranged on a rectangular grid.
+type GridXYZ interface {
+	// Dims returns the dimensions of the grid.
+	Dims() (c, r int)
```
kortschak: `Dims` or `Size` or as @vladimir-ch used `GridSize`?

vladimir-ch:
I took a look at the volcano example and I see there has been a misunderstanding. When I referred to "graphs of functions", I meant that in the mathematical sense, graph of a function as a set of points (x, y, z=f(x,y)), not how the data is represented (the volcano data is still a function). For a moment I thought that GridXYZ could represent also other surfaces, but it cannot.

plotter/heat.go
vladimir-ch: I think "grid dimensions" is a commonly used terminology, so I think `Dims()` is the best (in the meantime, I have renamed `GridSize` in `MinimalSurface` to `Dims`).

kortschak: OK.

kortschak:
Sorry, I misunderstood you. What kind of other surfaces are you talking about? Are you thinking of generalised meshes?

vladimir-ch:
I was thinking about parametric surfaces (x(u, v), y(u, v), z(u, v)). The only relation to the current PR is how to name the related types and methods if or when such surfaces are supported, or if they could be brought under the same interface.

kortschak:
Yeah, OK. I think that is probably something for another package to be honest. Certainly not under the same interface. There should be an issue for 3d plotting and where it fits in.

plotter/heat.go
egonelbre: btw. It seems that you didn't get the last version I made: https://gist.github.com/egonelbre/5770469d6d7b955a4559 So you might be able to grab something from there as well. Although I haven't tested it performance wise.

And looking at the loop excision - it seems that it should remove the previous "seen" entries otherwise there might be a problem with:

    """v"""\
       /"v"V
       V\___

(Sorry for the curde graphics)... I.e. when you take out the first loop from the contour line, the next loop would connect to that loop. And, for me, it's not obvious how it will work in that case.

kortschak: Thanks for making me think more about the loop excision algorithm. What I have is incorrect, but not for that reason.

There are three cases for a double cross over (x and y mark cross over points):

1-2x-3-4-5y-6-7-8y-9-10x-11
(expect 1-2-10-11, 2-3-4-5-8-9-10-2 and 5-6-7-8-5)

1-2xy-3-4-5-6-7-8y-9-10x-11
(expect 1-2-10-11, 2-3-4-5-6-7-8-2 and 8-9-10-8)

1-2y-3-4-5x-6-7-8y-9-10x-11
(1-2-8-9-10-11, 2-3-4-5-6-7-8-2 and 5-6-7-8-9-10-5, but will get 1-2-8-9-10-11, 2-3-4-5-6-7-8-2 and 10 which is not valid since has only one point).

The current code handles the first two cases correctly, and case 2 would be broken by removing elements from seen. I suspect the last case is rare, but I can't see why it should not happen in theory.

I'm not entirely sure how to handle case three. Obviously the working list can't be altered in these cases or you lose the capacity to make the second closing loop. It looks like a sequence reconstruction problem, so people at work might have ideas.

I will add some tests for these cases and others.

egonelbre: My version should give:

    1-2x-3-4-5y-6-7-8y-9-10x-11
    	5 6 7 8
    	2 3 4 5 9 10
    	1 2 11
    
    1-2xy-3-4-5-6-7-8y-9-10x-11
    	2 3 4 5 6 7 8
    	2 9 10
    	1 2 11
    
    1-2y-3-4-5x-6-7-8y-9-10x-11
    	2 3 4 5 6 7 8
    	1 2 9 10 11

I.e. once it detects a loop, it extracts the loop from the content as if it never happened and ignores it for the rest of processing. This should create good enough paths - but depending on the place where you start iterating you may get different results... i.e. the last case would give different results if the sequence were reversed. Also I'm not sure whether it's worth the effort to fix that... I assume this would mean constructing all possible loops, evaluating based on some metric and start removing them one-by-one.

egonelbre: I imagined the algorithm as drawing the line by hand and once you connect to a previous point, you remove that loop from your drawing.

kortschak: Yes, sorry, delete the 10s from the first string of the first two cases.

sbinet: FYI, I used Tarjan's algorithm to detect cycles in a dependency graph:
https://github.com/go-hep/fwk/blob/master/utils/tarjan/tarjan.go
http://en.wikipedia.org/wiki/Tarjan’s_strongly_connected_components_algorithm

hth,
-s

kortschak: Seems like overkill, but I think it's actually what is needed.

plotter/contour.go
```diff
@@ -0,0 +1,425 @@
...
+	// Palette is the color palette used to render
+	// the heat map.
+	Palette palette.Palette
+
+	// Min and Max define the dynamic range of the
+	// heat map.
+	Min, Max float64
+}
+
+// NewContour creates as new contour plotter for the given data,
```
btracey: It would be nice if a default set of levels could be allocated if levels == nil (10 contour lines at the 5, 10, ..., 95 percentile)

plotter/contour.go
```diff
@@ -0,0 +1,425 @@
...
+	// LineStyle is the style of the contour lines.
+	draw.LineStyle
+
+	// Palette is the color palette used to render
+	// the heat map.
+	Palette palette.Palette
+
+	// Min and Max define the dynamic range of the
+	// heat map.
+	Min, Max float64
```
btracey: Do Min and Max actually do anything for Contour? I see them set in NewContour, but I don't see them used in any of the methods. I'm not sure why they'd be necessary since the desired behavior is set with Levels.

plotter/contour.go
```diff
@@ -0,0 +1,425 @@
...
+		GridXYZ:   g,
+		Levels:    levels,
+		LineStyle: DefaultLineStyle,
+		Palette:   p,
+		Min:       min,
+		Max:       max,
+	}
+}
+
+const naive = false
```
btracey: I'm not sure why this is a constant. It seems like there should be a TODO here with some description.

plotter/contour.go
kortschak: Calculating percentiles is quite costly in terms of allocation (all elements need to be allocated again in a slice to do the sort), but it can be done easily. What default for the chosen contours: 5, 25, 50, 75, 95?

plotter/contour.go
kortschak: Yes, that was left over. If a default level slice is created they become useful again.

plotter/contour.go
kortschak: This is a debugging constant (see the comment over the conditional that uses it). I can add a comment here too.

plotter/contour.go
btracey: That sounds like a good set of defaults.

plotter/contour.go
btracey: Why would they be useful again? The proposed default behavior was to use quantiles which don't need Min or Max.

kortschak: Say you have the defaults, but you want to see which cotnoutrs are above some absolute threshold. You set the appropriate field or Min or Max and the (deleted but to be readded) fields Overflow and Underflow colours and they are drawn in the approriate colours.

kortschak:
@btracey PTAL

plotter/volcano
```diff
@@ -0,0 +1,84 @@
...
+		if err := p.Save(4, 4, name+ext); err != nil {
+			panic(err)
+		}
+	}
+}
+
+// Data extracted from RDatasets volcano data for the Maunga Whau volcano topographic data.
+var volcano = deciGrid{mat64.NewDense(87, 61, []float64{
+EOF
+R -q -e 'write.table(as.data.frame(volcano), file="volcano_example.go", sep=", ", eol=",\n", col.names=FALSE, row.names=FALSE, append=TRUE)'
```
btracey: Do I need R installed to run this example?

Can we have a different example? I feel like this is an unidiomatic use of 'go generate'. In general, I don't think go generate is intended to be used by library users, instead just by authors. Instead, could we establish something like gonum/plot/examples/examplename where they are all compileable examples? I don't know what the licensing is on the volcano data, but we could have the data marshaled into a *mat64.Dense directly. It's fine to use generate to do that construction (which is what it looks like this execution does), but the actual published volcano_example should be able to be compiled and ran without using go generate.

plotter/contour.go
```diff
@@ -0,0 +1,488 @@
...
+	Levels []float64
+
+	// LineStyle is the style of the contour lines.
+	draw.LineStyle
+
+	// Palette is the color palette used to render
+	// the heat map.
+	Palette palette.Palette
+
+	// Underflow and Overflow are colors used to draw
```
btracey: This all seems weird to me. The purpose of Contour is to plot the iso-surfaces of a function. These isosurfaces are specified by Levels. What does it matter if the values lie outside of Min and Max? That has no effect on what the contours are.

plotter/contour.go
```diff
@@ -0,0 +1,488 @@
...
+
+	"github.com/gonum/plot"
+	"github.com/gonum/plot/palette"
+	"github.com/gonum/plot/vg"
+	"github.com/gonum/plot/vg/draw"
+)
+
+// Contour implements the Plotter interface, drawing
+// a contour plot of the values in the GridXYZ field. The
+// order of rows is from high index to low index down.
```
btracey: Instead of "the order of rows is high index to low index down", can't we just say "row index is from top to bottom"?

plotter/volcano
kortschak: Yes, you need R.

I was avoiding having to understand the legal implications of the GPL, but this example is widely used and I felt it justified to include. If someone can clarify that copyright does not cover the data then I can include this as it is generated.

plotter/contour.go
kortschak: They provide a mechanism to illustrate when the field is outside the expected range without having to fiddle with the palette.

Say I'm doing a densitometric analysis of an image and I am aware of the dynamic range of the instrument making the measurements, I can still see the contours, but they are set to (e.g.) warning colours. This does not alter the colour scaling of the within range contours and can be done without altering the palette in any way. It is a convenience.

plotter/contour.go
kortschak: Done.

plotter/contour.go
```diff
@@ -0,0 +1,488 @@
...
+// Contour implements the Plotter interface, drawing
+// a contour plot of the values in the GridXYZ field. The
+// row index is from top to bottom.
+type Contour struct {
+	GridXYZ GridXYZ
+
+	// Levels describes the contour heights to plot.
+	Levels []float64
+
+	// LineStyle is the style of the contour lines.
```
btracey: I don't think LineStyle should be embedded. Contour.LineStyle.Width is much clearer than Contour.Width

kortschak: Sure. I was following the style in other plotters.

Perhaps there can be a sweep across them all together.

btracey: SGTM

plotter/contour.go
```diff
@@ -0,0 +1,488 @@
...
+	Underflow color.Color
+	Overflow  color.Color
+
+	// Min and Max define the dynamic range of the
+	// heat map.
+	Min, Max float64
+}
+
+// NewContour creates as new contour plotter for the given data,
+// using the provided palette.
```
btracey: There needs to be a description of the MinMaxer interface.

plotter/contour.go
```diff
@@ -0,0 +1,488 @@
...
+	switch g := g.(type) {
+	case minMaxer:
+		min, max = g.Min(), g.Max()
+	default:
+		min, max = math.Inf(1), math.Inf(-1)
+		c, r := g.Dims()
+		for i := 0; i < c; i++ {
+			for j := 0; j < r; j++ {
+				v := g.Z(i, j)
+				if math.IsNaN(v) {
```
btracey: Other plotters return an error if there are NaN values (scatter). It doesn't need to be fixed now, but we should be consistent about it.

btracey: Or rather, we should have a discussion on what our general policy is on NaN

plotter/contour.go
```diff
@@ -0,0 +1,488 @@
...
+		Palette:   p,
+		Min:       min,
+		Max:       max,
+	}
+}
+
+// Default quantiles for case where levels is not explicitly set.
+var defaultQuantiles = []float64{0.01, 0.05, 0.25, 0.5, 0.75, 0.95, 0.99}
+
+// quantilesR7 returns the pth quantiles of the data in g according the th R-7 method.
```
btracey: My opinion is that we should be using gonum/stat here, and if gonum/stat isn't currently sufficient we should improve it. Certainly that would imply coming up with a public name for R7, and possibly implementing a Quantiles function that efficiently finds a set of sorted quantiles instead of just one. Thoughts?

plotter/contour.go
kortschak: Agreed.

plotter/contour.go
kortschak: You think so?

plotter/contour.go
kortschak: This is small enough it can be done here. If it were in stat, I imagine it would be `quantileR7(v []float64, p float64) float64` which is not what I want here.

plotter/contour.go
btracey: Well, we allow custom implementation of GridXYZ. It should be documented somewhere that having a Min and Max function could affect the behavior of plotting.

kortschak: Mmm. I guess so. I would have thought it was pretty obvious, but I'll add a comment. I don't think it warrants an additional type in the godoc.

plotter/contour.go
```diff
@@ -0,0 +1,488 @@
...
+// http://en.wikipedia.org/wiki/Quantile#Estimating_the_quantiles_of_a_population
+func quantilesR7(g GridXYZ, p []float64) []float64 {
+	c, r := g.Dims()
+	data := make([]float64, 0, c*r)
```
btracey: stat.Quantile returns NaN if any of the values are NaN. I should have left a better comment, but I'm pretty sure this is because the data is not sorted if there are NaN values. After the call to sort, I think the NaN values can be anywhere, making quantiles inconsistent for the same data. It's probably better to not append NaN values so at least the same quantiles are found

plotter/contour.go
btracey: We'd add a new CumulantKind rather than a new function.

It isn't what you want because of the slice, or is there something I'm missing? I'm suggesting maybe we should also have Quantiles(p []float64, c CumulantKind, x, weights []float64) []float64

plotter/contour.go
```diff
@@ -0,0 +1,488 @@
...
+	}
+	sort.Float64s(data)
+	v := make([]float64, len(p))
+	for j, q := range p {
+		if q == 1 {
+			v[j] = data[len(data)-1]
+		}
+		h := float64(len(data)-1) * q
+		i := int(h)
+		v[j] = data[i] + (h-math.Floor(h))*(data[i+1]-data[i])
```
btracey: This is another place where NaN can corrupt.

plotter/contour.go
```diff
@@ -0,0 +1,488 @@
...
+type Contour struct {
+	GridXYZ GridXYZ
+
+	// Levels describes the contour heights to plot.
+	Levels []float64
+
+	// LineStyle is the style of the contour lines.
+	draw.LineStyle
+
+	// Palette is the color palette used to render
```
btracey: Could you add a comment that if len(Palette) == 0 then the linestyle color will be used?

plotter/contour.go
kortschak: The slice is it. OK. I can see what you are thinking. I prefer this simplicity - I can look at this code and immediately see what it is doing, but I suppose I'd be happy to accept that change in a later PR after stat supports it.

plotter/contour.go
```diff
@@ -0,0 +1,488 @@
...
+// Plot implements the Plot method of the plot.Plotter interface.
+func (h *Contour) Plot(c draw.Canvas, plt *plot.Plot) {
+	var pal []color.Color
+	if h.Palette != nil {
+		pal = h.Palette.Colors()
+	}
+	if len(pal) == 0 {
+		pal = []color.Color{h.LineStyle.Color}
+	}
+
```
btracey: I think the code eventually panics if len(pal) < len(h.Levels). There should be an explicit check / documentation

plotter/contour.go
btracey: I don’t think hidden interface testing is ever obvious

It doesn’t need an extra type.

plotter/contour.go
btracey: SGTM

plotter/contour.go
kortschak: I shouldn't. Try setting a nil palette and running `go test -visual`.

plotter/contour.go
kortschak: The behaviour for nan in sort.Float64Slice is well defined: http://golang.org/src/sort/sort.go?s=5740:5767#L233, but yes, I should ignore them in one way or another.

Done.

plotter/contour.go
kortschak: Done.

plotter/contour.go
```diff
@@ -0,0 +1,494 @@
...
+		i := int(h)
+		v[j] = data[i] + (h-math.Floor(h))*(data[i+1]-data[i])
+	}
+	return v
+}
+
+// naive is a debugging constant. If true, Plot performs no contour path
+// reconstruction, instead rendering wach path segment individually.
+const naive = false
+
```
btracey: Sorry, this function was a little hard for me to work through. Could you do the following variable name substitutions
pal --> palette
cp --> cps    (bike shedding, I know, but it makes it clearer that there are multiple of them)
pa --> path
ps --> ?   I'm not sure what this does yet 

I was expecting to see that level i would get plotted with color palette[i], but it seems instead that it gets plotted with some percentage of the palette? Why is that the right behavior?

plotter/contour.go
```diff
@@ -0,0 +1,494 @@
...
+		if len(h.Levels) == 1 {
+			ps = 0
+		}
+
+		for _, z := range h.Levels {
+			if math.IsNaN(z) {
+				continue
+			}
+			for _, pa := range cp[z] {
+				if isLoop(pa) {
```
btracey: Sorry, I don't understand this (probably due to not really understanding vg.Path). If the path starts and ends at the same place, why does it need to have a line segment attaching the point to itself?

kortschak: When we want to do fills, this is necessary AFAIK.

plotter/contour.go
btracey: I see your point, but it feels to me like this is the wrong layer to put it at. A palette just defines a sequence of colors, but if we had
````
type Colormap func(float64) color.Color  // TODO: better name
````
then the warning mechanism follows through cleanly. The colormap associates z-axis numbers to colors, which could include producing warning colors at unexpected data values.

File: plotter/contour.go (hunk 18 above)
Comment (kortschak) [inline plotter/contour.go]: ps is palette scale. This allows spreading of a palette across a set of levels where the numbers don't exactly match. There is no definitive correct behaviour, this is a reasonable approach that works well in practice.

I don't think those changes are helpful; pa is what vg.Paths are normally named, cp is created immediately and the range says cp[z], pal is actually not really a palette it's a []color.Color. I can rename ps or add a comment.

File: plotter/contour.go (hunk 8 above)
Comment (kortschak) [inline plotter/contour.go]: Yeah. I understand. This is something I spent a fair amount of time
working through when I was designing palette and I'm still not entirely
happy with it. Can you file an issue to have a look at this again later.

Comment (btracey) [inline plotter/contour.go]: Filed #20

File: plotter/contour.go (hunk 18 above)
Comment (btracey) [inline plotter/contour.go]: Please add a comment with ps

File: plotter/heat.go [hunk 20]
```diff
@@ -0,0 +1,212 @@
...
+	// X returns the coordinate for the column at the index x.
+	// It will panic if c is out of bounds for the grid.
+	X(c int) float64
+
+	// Y returns the coordinate for the row at the index r.
+	// It will panic if r is out of bounds for the grid.
+	Y(r int) float64
+}
+
+// HeatMap implements the Plotter interface, drawing
```
btracey: Same comments here as for Contour (top down, length of palette)

plotter/contour.go
```diff
@@ -0,0 +1,499 @@
...
+	c.SetLineStyle(h.LineStyle)
+
+	trX, trY := plt.Transforms(&c)
+
+	// Collate contour paths and draw them.
+	//
+	// The alternative naive approach is to draw each line segment as
+	// conrec returns it. The integrated path approach allows graphical
+	// optimisations and is necessary for contour fill shading.
+	if !naive {
```
eaburns: Can the two branches of this condition each be a s separate function or method?

kortschak: Sure. I'll pull the naive out into plotNaive and make a check for naive in a preamble in Plot.

kortschak: Done.

kortschak:
@gonum/developers PTAL

I think this is ready (except for the volcano example issue which I have skirted by not even mentioning go generate - when/if someone can demonstrate that the data are not subject to the GPL a PR can provide the example complete).

plotter/volcano
kortschak: I have looked at the NZ and US copyright acts and they both exclude facts/discoveries (by various names), so the data from RDatasets we use here should be free from copyright - I have built the complete example and committed it. caveat: not lawyer.

plotter/conrec.go
```diff
@@ -0,0 +1,177 @@
...
+		h      [5]float64
+		sh     [5]int
+		xh, yh [5]float64
+
+		im = [4]int{0, 1, 1, 0}
+		jm = [4]int{0, 0, 1, 1}
+
+		cases = [3][3][3]int{
+			{{0, 0, 8}, {0, 2, 5}, {7, 6, 9}},
+			{{0, 3, 4}, {1, 0, 1}, {4, 3, 0}},
```
btracey: The middle point of this is "3" in the conrec.c algorithm, but is 0 here. Is that intentional?

plotter/conrec.go
```diff
@@ -0,0 +1,177 @@
...
+			{{9, 6, 7}, {5, 2, 0}, {8, 0, 0}},
+		}
+	)
+
+	c, r := g.Dims()
+	for i := 0; i < c-1; i++ {
```
btracey: The other thing I notice about the translation is that the loops seem to go in another direction from the original code. I can't see that it matters, but thought I'd point it out.

plotter/conrec.go
kortschak: Yes. I wrote to Paul about that, but haven't had a reply.

I believe the 3 in that position is incorrect when considering the commentary in the paper.

axiom: When we have a section of the grid where all the Z values are equal, and equal to a contour height we would expect to have no internal segments to draw.

This is covered by case g) in his description of the algorithm (a triangle with three vertices the lie on the contour level). He says "... case g above has no really satisfactory solution and fortunately will occur rarely with real arithmetic." and then goes on to show the following:

conrec cases

which shows case g) in the set where no edge is drawn.

However, in the iteration over sh at conrec.c +44 (conrec.go +74), a triangle with all vertices on the plane is given sh = {0,0,0,0,0} and then when the switch happens, castab/cases resolves that to case 3 for all values of m. This is fixed by replacing castab/cases[1][1][1] with 0.

plotter/contour.go
```diff
@@ -0,0 +1,659 @@
...
+				c.SetLineStyle(style)
+				c.SetColor(col)
+				c.Stroke(pa)
+			}
+		}
+	}
+}
+
+// naivePlot implements the a naive rendering approach for contours.
+// It is here as a debugging mode since it sinply draw line segments
```
btracey: s/sinply/simply

Something is wrong with "draw line segments returned generated"

plotter/conrec.go
kortschak: Yeah, I had no idea why he was doing that, it makes not difference to the outcome. I chose forward looping for ease - at one point I was ranging (before GridXYZ got a Dims() method).

plotter/contour.go
kortschak: Indeed something is.

plotter/contour.go
```diff
@@ -0,0 +1,659 @@
...
+
+// endMap holds a working collection of available ends.
+type endMap map[point]*contour
+
+// paths extends a conrecLine function to build a set of conts that represent
+// paths along contour lines. It is used as the engine for a closure where ends
+// and conts are closed around in a conrecLine function, and l and z are the
+// line and height values provided by conrec. At the end of a conrec call,
+// conts will contain a map keyed on the set of paths with each value being
+// the height of the contour.
```
kortschak: This comment is incorrect - fixed in pending commit.

btracey:
I don't fully comprehend a lot of the contour code, but otherwise LGTM

kortschak:
Are there sections that are a concern in particular wrt comprehension. I would like to make this as clear as possible. If comments in any sections would help, I'll add them.

kortschak:
@egonelbre would you please take a quick look at this?

egonelbre:
I took a quick look and based on that quick look LGTM. *Basically, didn't notice anything sticking out.*

kortschak:
Thank you everyone.

btracey:
Thank you for putting that together! Contour maps are very useful (now I feel worse about the lack of a quadrature package).

PR #16, opened 2015-02-12
vgx11: an X11-based backend

sbinet:
- adds support for `X11` on unix (ie: `!windows`)

needs #9

### DIFF
--- a/vg/vgimg/vgimg.go
+++ b/vg/vgimg/vgimg.go
@@ -53,13 +53,22 @@ func New(width, height vg.Length) *Canvas {
 // minimum point of the given image
 // should probably be 0,0.
 func NewImage(img draw.Image) *Canvas {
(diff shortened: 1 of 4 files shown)


eaburns:
If I understand, this needs to be rebased once #9 goes in. Then you need to squash it. Besides that, and assuming that it's the same as what I've seen before: LGTM.

sbinet:
yes, it is the same. will rebase when #9 lands.

sbinet:
squashed.

sbinet:
do I need another LGTM or can I merge?
(sorry for the nitpicking, this kind of "little detail" should perhaps be put in our general "how to contribute" guidelines...)

kortschak:
If it has made no changes since the LGTM and the tests pass (which they appear to), I think you can merge this. Please do a fast forward merge.

sbinet:
ok, done.

PR #22, opened 2015-02-26
Fix algorithm origin comment

kortschak:
This was raised in code review for gonum/graph#34. The code was changed, but the comment was left. Now the comment points to the correct WP revision.


sbinet:
LGTM

PR #23, opened 2015-03-13
vgimg dpi, vgsvg origin, and legend alignment

ctessum:
Hi,

In these changes I fixed a bug in the alignment between the legend keys and their respective thumbnails, and changed the origin in vgsvg to be the middle of the document so that it is consistent with vgimg. When the origin was at the bottom left corner as before, using the draw.Canvas.Crop() method to create subfigures produced incorrect results.

Also, in vgimg, 96 dpi is not sufficient for certain applications, such as print publication, so making 96 dpi a constant limits the usefulness. Instead, I've made dpi an input argument for creating a new canvas, but if backwards compatibility is desired perhaps there could be separate functions for creating a new canvas, some that require dpi to be specified and others that don't.

### DIFF
--- a/legend.go
+++ b/legend.go
@@ -114,6 +114,7 @@ func (l *Legend) draw(c draw.Canvas) {
 		yoffs := (enth - l.TextStyle.Height(e.text)) / 2
 		c.FillText(l.TextStyle, textx, icon.Min.Y+yoffs, xalign, 0, e.text)
 		icon.Min.Y -= enth + l.Padding
+		icon.Max.Y -= enth + l.Padding
 	}
 }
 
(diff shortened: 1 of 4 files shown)


kortschak:
Take a look at http://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis for an approach to optional attributes. This is the way we want to go more egenrally, so this might be a good place to start.

Please separate the two issues into different PRs.

eaburns:
As for DPI, I think only the most general New function should take it as a parameter; the rest should default to 96. Those who want a special DPI can call the most general New, others needn't worry about it.

kortschak:
if we move to functional options, the point of generality becomes moot.

ctessum:
I took a stab at the functional options (PR #24), but all the different New... functions makes it a little messy. I'll follow up with the other changes separately.

PR #24, opened 2015-03-13
added dpi option to vgimg

ctessum:
As suggested...


kortschak:
Define vgimg.Option as func(*Canvas), then the label can be 'o' instead of 'options' and the signatures become shorter and don't need to break across lines.

Also, please put the initialized field in a separate block and fix up the comment to be a complete sentence.

Other than that LGTM (but wait for @eaburns)

@jonlawlor, do we know why tests aren't running.

vg/vgimg/vgimg.go
```diff
@@ -40,46 +44,73 @@ type Canvas struct {
...
+	c.w = vg.Length(w/float64(c.dpi)) * vg.Inch
+	c.h = vg.Length(h/float64(c.dpi)) * vg.Inch
+	c.color = []color.Color{color.Black}
 	vg.Initialize(c)
+	c.initialized = true
 	return c
 }
 
+// DPI sets the dots per inch of a canvas. It should only be
+// used as an "option" argument when initializing a new canvas.
```
eaburns: No need to quote option.

vg/vgimg/vgimg.go
```diff
@@ -40,46 +44,73 @@ type Canvas struct {
...
+func DPI(dpi int) func(*Canvas) {
+	return func(c *Canvas) {
+		if c.initialized {
```
eaburns: I think that you are trying to avoid someone doing:
```c := vgimg.New(...)
DPI(12345)(c)```
which is invalid because it changes the DPI without changing it on the underlying GraphicsContext. A different way to prevent this is to make options opaque:
```type option func(*Canvas)
func DPI(dpi int) option { ... }```
The caller can still create an option via `DPI()` and pass it to `New`, but now they don't know what it is and they cannot call it themselves, and you don't need the initialized flag or this panic.

Comment (kortschak) [inline vg/vgimg/vgimg.go]: The unexported option type can still be called, but by making it unexported, we lose its documentation.

Comment (kortschak) [inline vg/vgimg/vgimg.go]: ... losing the documentation is not necessarily a bad thing.

File: vg/vgimg/vgimg.go [hunk 3]
```diff
@@ -40,46 +44,73 @@ type Canvas struct {
...
+	c.dpi = defaultDPI
+	for _, option := range options {
+		option(c)
+	}
+	w := width / vg.Inch * vg.Length(c.dpi)
+	h := height / vg.Inch * vg.Length(c.dpi)
 	img := image.NewRGBA(image.Rect(0, 0, int(w+0.5), int(h+0.5)))
 
-	return NewImage(img)
+	return NewImage(img, options...)
```
eaburns: I don't like how this evaluates the options multiple times. I realize that it's not causing a problem here, but I still think that it would be better not to get into this habit. Instead, we should factor out code that's currently shared by delegating to the next `New` call into separate functions that are called by each `New`:
```
func makeGraphicsContext(img draw.Image, dpi int) draw2d.GraphicsContext {
  gc := draw2d.NewGraphicsContext(img)
  gc.SetDPI(dpi)
  gc.Scale(1, -1)
  gc.Translate(0, -h)
  return gc
}

func initCanvas(c *Canvas, img draw.Image, gc draw2d.GraphicsContext) {
  w := float64(img.Bounds().Max.X - img.Bounds().Min.X)
  h := float64(img.Bounds().Max.Y - img.Bounds().Min.Y)	
  draw.Draw(img, img.Bounds(), image.White, image.ZP, draw.Src)
  c.w = vg.Length(w/float64(c.dpi)) * vg.Inch
  ...
}
```

ctessum:
I've made the requested changes (I think).

vg/vgimg/vgimg.go
```diff
@@ -40,46 +43,77 @@ type Canvas struct {
...
 // nearest pixel.
-func New(width, height vg.Length) *Canvas {
-	w := width / vg.Inch * dpi
-	h := height / vg.Inch * dpi
-	img := image.NewRGBA(image.Rect(0, 0, int(w+0.5), int(h+0.5)))
-
-	return NewImage(img)
+func New(width, height vg.Length, o ...option) *Canvas {
+	c := initCanvas(o)
+	return NewImage(c.newImage(width, height), o...)
```
eaburns: This still calls initCanvas 3 times. I would simply not call NewImage, but call the functions that it calls here. This probably means some code from NewImageWithContext into a separate function so that you don't need 3 copies of it.

Does that make sense?

ctessum:
Since we're kind of dwelling here, when I was trying to figure out how to use this I found "NewImage" and "NewImageWithContext" to be super confusing. Since I was trying to make an image, I figured NewImage was the easiest option, but that wasn't the case. Instead I would suggest NewFromImage and NewFromImageWithContext, but since we're already the functional options route, why not just have the one "New" and then make UseImage(img draw.Image) and UseImageWithContext(...) as functional options? I think to implement it, New would have to first make a new image and context and then overwrite them if the options required, but to me this would be a small price to pay for the decrease in confusion.

kortschak:
I think folding these into a set of functional options makes enourmous sense - I was thinking this before, so thanks for suggesting it. This probably requires some sketching to get things right; can you outline what your plan of attack would be?

ctessum:
After thinking about it, one issue is that the options would not be independent: one option would be to set the dpi, but the other options would also set or depend on the dpi, so the result would be dependent upon the order in which the options were passed. Instead, it might work to do something like this:

New(i Init) *Canvas {
...
}

type Init func(*Canvas)

func FromWH(w, h) Init {
...
}

func FromWHDpi(w, h, dpi) Init {
...
}

func FromImage(img) Init {
...
}

func FromImageWithDpi(img, dpi) Init {
...
}

func FromImageWithContext(img, GC) Init {
...
}

Init might be better than init because I think that godoc will group the functions if Init is exported but it might not otherwise. Let me know what you think.

kortschak:
This just brings us back to where we were, but with one extra level of indirection. You are right though about the non-independence. I was just expecting that inconsistent options would panic or error.

kortschak:
The sketch of what I was suggesting is that option is func(*Canvas) uint32 (more than we need) where the return value sets flags of what actions were taken.

    var g uint32
    for _, opt := range o {
        f := o(c)
        if g&f != 0 {
            panic("incompatible options")
        }
        g |= f
    }

We define

    const (
        setsDPI = 1<<iota
        setsSize
    )

Then return the appropriate const for each option.

Necessarily non-nil fields that are nil after the loop are then created from the settings that have been chosen in the loop.

ctessum:
Is this what you had in mind? I added a test but this change probably breaks things elsewhere.

eaburns:
I feel like this is getting too complex for simply adding a DPI option. Now we are adding various, possibly conflicting, options. The caller needs to take special care to use the right combination

What if instead we keep the original API and document that New and NewImage use the default of 96 DPI and NewImageWithContext uses the DPI from the context. If you want to set the DPI then you have to create your own image and your own context. That doesn't seem too arduous for the very rare case of needing a special DPI.

ctessum:
I guess the question becomes, what are people going to use this package for? As far as I know, there are three main venues for plots: the web, powerpoint, and printed documents (I would bundle pdf documents with printed documents because they're designed to be printed). Everyone has a different background, but the people that I know usually use all three (students, scientists, technical employees, etc). 96 dpi is not suitable for printed documents:

http://www.sciencemag.org/site/feature/contribinfo/prep/prep_revfigs.xhtml

eaburns:
Raster images, like those generated by vgimg are meant for computer screens. They don't scale and they look very bad on paper. Pdf and eps are there for print. Both can be easily embedded into a LaTeX doc. I suspect that Word or whatever else can handle PDF too, no? I would encourage you not to use raster images for print.

That said, it doesn't mean we shouldn't support it, but I don't think that we should make an overly complex API to cater to it.

btracey:
My use case is to simulate being able to "zoom in" on plots with lots of
data. I could make the image bigger, but it's a fair amount of effort to
scale all the other parts of the plot to keep the size proportions.

eaburns:
I'm not sure I understand your comment about scaling ask other parts of the plot. But if you want to zoom in and still have the plot look good, vgimg is exactly *not* the backend to use. It's the only one that doesn't use a vector graphics representation of the image, so it's the only one for which you cannot zoom without getting pixel artifacts. Increasing DPI can defer the issue, but it'll always be there at some zoom level.

Anyway, I won't argue any further. You're all the ones using the package, so you know better what you want from it.

kortschak:
Zoomable plots is exactly where svg and eps shine. Don't use raster graphicsmfor that.

ctessum:
Sometimes, vector images end up being too large to include in a reasonably sized document, like if there are scatter plots with many data points. Then you have to revert to raster graphics. Ms word doesn't take vector graphics, nor does google docs.

ctessum:
Does this work? Now there is still a New(w, h), like all of the other backends, but the alternatives are combined into a NewFromOptions(o ...option).

vg/vgimg/vgimg.go
```diff
@@ -22,63 +22,138 @@ import (
...
-	h := height / vg.Inch * dpi
-	img := image.NewRGBA(image.Rect(0, 0, int(w), int(h)))
+// These constants are used to ensure that the options
+// used when initializing a canvas are compatible with
+// each other.
+const (
+	setsDPI = 1 << iota
+	setsSize
+
+	setsDPIAndSize = setsDPI | setsSize
```
eaburns: I would just or these together in the one place where they are used. No need to introduce a new name for it.

vg/vgimg/vgimg.go
```diff
@@ -22,63 +22,138 @@ import (
...
+)
+
+const (
+	// defaultDPI is the default number of dots per inch.
+	defaultDPI = 96
+
+	// defaultWidth and defaultHeight are the default canvas
+	// dimensions.
+	defaultWidth  = 4 * vg.Inch
```
eaburns: If we want to take this approach then I think we should export these so they show up in the docs.

eaburns:
The thing that I dislike about this approach is that it's overly complex with little gain, not that it would get rid of `New(w, h)`. We've determined that there are only very few, specific cases where one wants to set DPI (scatter plots with too many points for a vector format, where one can't use random sampling to reduce the number of points, and one needs to print it). It's not so bad for such users to create their own image and context in this peculiar situation. It's 10s of lines of code that they can easily wrap in a function within their data analysis code.

That said. I don't want to block this any further, so if it LGTO (looks good to others :-D) then it's fine by me.

eaburns:
Oh! We also need to improve the comment on NewWithOptions. It needs to say that it uses the defaults if they are not overridden. It also needs to explain option compatibility; namely, which are not compatible and what happens when you pass incompatible options.

kortschak:
I'm happy with this - we're still quite fluid with API breakage at the moment, so if we decide we don't like it we can change it.

I would prefer NewWithOptions to be called NewUsing or NewWith.

ctessum:
PTAL

eaburns:
@kortschak, you seemed happy with this approach, did you want to take a peek at this and LGTM it?
At very least, it looks like it needs to be rebased.

kortschak:
I think this is still OK, but I suspect it will break things now. I'll have a closer look at what needs to be fixed up if anything later today.

kortschak:
This all LGTM after a rebase (and possibly a squash).

Can you do this @ctessum?

ctessum:
Yes, I can probably get to it next week.

ctessum:
@kortschak I've rebased it; want to walk me through again how to correctly merge it with the master? It's something other than just clicking the green "merge pull request" button, right?

kortschak:
You'll need to rebase it onto gonum's master as well.

ctessum:
So, on my local computer, I pull from github.com/ctessum/plot to github.com/gonum/plot/, rebase github.com/gonum/plot, and then push github.com/gonum/plot to this website? Won't that break the link between the changes and this code review, though?

kortschak:
You are right, that will break the link. Fetch gonum master and rebase onto that then push to your PR branch. After that's done I can fast forward merge into master.

ctessum:
OK I already did that (I believe).

kortschak:
No, I don't think so. If you look at the network graph you can see the parent of this is a long way back.

The history is not linear at the moment though, so we may as well just click the big green button.

eaburns:
Let's not give up. Can we please try to rebase this one more time please?

ctessum:
I think I've fixed it. Let me know if I should push the big green button (or do something else).

kortschak:
You have fixed it.

Are you tracking gonum master? If so, checkout your local branch that does that and then git merge. It should say it did it fast forward (or you can say `--ff-only` and it won't do anything if there is a problem).

When that is done cleanly (you can check this by doing git diff <yourfork>/dpi and it should print nothing), you can push to gonum master.

PR #26, opened 2015-03-14
Fixed cropping bug

ctessum:
There is an issue with the draw.Canvas.Crop() function. I've added a test to document it. The test does cropping two different ways, the rectangles of which should be the same, and produces two output images that should also be the same. In the current version the two methods yield different results. The change that I made fixes this bug.

### DIFF
--- a/vg/draw/draw_test.go
+++ b/vg/draw/draw_test.go
@@ -1,0 +1,58 @@
+package draw
+
+import (
+	"image/color"
+	"reflect"
+	"testing"
+
+	"github.com/gonum/plot/vg"
+	"github.com/gonum/plot/vg/recorder"
+)
+
+func TestCrop(t *testing.T) {
+	ls := LineStyle{
+		Color: color.NRGBA{0, 20, 0, 123},
+		Width: 0.1 * vg.Inch,
+	}
+	r1 := recorder.New(96)
+	c1 := NewCanvas(r1, 6, 3)
+	c11 := c1.Crop(0, 0, -3, 0)
+	c12 := c1.Crop(3, 0, 0, 0)
+
+	r2 := recorder.New(96)
+	c2 := NewCanvas(r2, 6, 3)
+	c21 := Canvas{
+		Canvas: c2.Canvas,
+		Rectangle: Rectangle{
+			Min: Point{0, 0},
+			Max: Point{3, 3},
+		},
+	}
+	c22 := Canvas{
+		Canvas: c2.Canvas,
+		Rectangle: Rectangle{
+			Min: Point{3, 0},
+			Max: Point{6, 3},
+		},
+	}
+	str := "unexpected result: %+v != %+v"
+	if c11.Rectangle != c21.Rectangle {
+		t.Errorf(str, c11.Rectangle, c21.Rectangle)
+	}
+	if c12.Rectangle != c22.Rectangle {
+		t.Errorf(str, c11.Rectangle, c21.Rectangle)
+	}
+
+	c11.StrokeLine2(ls, c11.Min.X, c11.Min.Y,
+		c11.Min.X+3*vg.Inch, c11.Min.Y+3*vg.Inch)
+	c12.StrokeLine2(ls, c12.Min.X, c12.Min.Y,
+		c12.Min.X+3*vg.Inch, c12.Min.Y+3*vg.Inch)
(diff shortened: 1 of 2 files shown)


kortschak:
The fix is correct, but the tests are dependent on manual inspection. Rather than using vgimg and writing outthe resulting png, use a vg/recorder.Canvas and compare the resulting sets of vg operations mechanically (you could write the png to a bytes.Buffer, but I would prefer to be able to see that both things are sane).

vg/draw/draw_test.go
```diff
@@ -0,0 +1,71 @@
+package draw_test
+
+import (
+	"fmt"
```
eaburns: Imported and not used.

eaburns:
Chris, thanks for this and for your other changes too. Could you please explain this bug in a bit more detail (in fact, could you please add the explanation to the commit message)? I don't quite understand what's happening here, but I also don't understand why I started the canvas at -w/2, -h/2 either :)

kortschak:
It looks like it was an error in Seb's change to rect. The original code used an implicitly zero Min field. For some reason that got translated into what exists now.

eaburns:
LGTM, once Dan's suggestions are in.

ctessum:
I fixed the test to not require manual inspection

kortschak:
On mobile, so I'll do this properly later, but to get you started: you can reflect.DeepEqual-compare the recorder actions; now that it's not dependent on being visible, the inches can go away; obvious cleanup of commented out sections; and what is the reason for the change to plotter/main.go?

vg/draw/draw_test.go
```diff
@@ -0,0 +1,58 @@
...
+		c12.Min.X+3*vg.Inch, c12.Min.Y+3*vg.Inch)
+	c21.StrokeLine2(ls, c21.Min.X, c21.Min.Y,
+		c21.Min.X+3*vg.Inch, c21.Min.Y+3*vg.Inch)
+	c22.StrokeLine2(ls, c22.Min.X, c22.Min.Y,
+		c22.Min.X+3*vg.Inch, c22.Min.Y+3*vg.Inch)
+
+	c1Actions := c1.Canvas.(*recorder.Canvas).Actions
+	c2Actions := c2.Canvas.(*recorder.Canvas).Actions
+	if !reflect.DeepEqual(c1Actions, c2Actions) {
+		t.Errorf("%#v != %#v", c1Actions, c2Actions)
```
kortschak: Change the format string to "unexpected result: %+v != %+v".

vg/draw/draw_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/gonum/plot/vg/recorder"
+)
+
+func TestCrop(t *testing.T) {
+	ls := LineStyle{
+		Color: color.NRGBA{0, 20, 0, 123},
+		Width: 0.1 * vg.Inch,
+	}
+
+	c1 := NewCanvas(recorder.New(96), 6, 3)
```
kortschak: Keep the recorders as r1 and r2, then you don't need the assertion below.

vg/draw/draw_test.go
```diff
@@ -0,0 +1,58 @@
...
+	c12.StrokeLine2(ls, c12.Min.X, c12.Min.Y,
+		c12.Min.X+3*vg.Inch, c12.Min.Y+3*vg.Inch)
+	c21.StrokeLine2(ls, c21.Min.X, c21.Min.Y,
+		c21.Min.X+3*vg.Inch, c21.Min.Y+3*vg.Inch)
+	c22.StrokeLine2(ls, c22.Min.X, c22.Min.Y,
+		c22.Min.X+3*vg.Inch, c22.Min.Y+3*vg.Inch)
+
+	c1Actions := c1.Canvas.(*recorder.Canvas).Actions
+	c2Actions := c2.Canvas.(*recorder.Canvas).Actions
+	if !reflect.DeepEqual(c1Actions, c2Actions) {
```
kortschak: After you have r1 and r2, this becomes `!reflect.DeepEqual(r1.Actions, r2.Actions)` and the 2 lines above go away.

kortschak:
LGTM after minor nits.

Then rebase onto gonum/plot master if necessary and squash.

kortschak:
Thanks, Chris.

dgryski:
This change broke all the example images generated by plotter/main.go.
example_barchart

PR #158, opened 2015-03-16
Clean up gob tests and move into plot/gob

kortschak:
It was getting in the way of other tests in plot and it is really a gob test, not a plot test.

Also added header and made organisation clearer.

@sbinet @eaburns @gonum/developers


sbinet:
LGTM

eaburns:
LGTM

PR #179, opened 2015-03-20
Fixed vgimg origin to match other backends

ctessum:


kortschak:
Is there a bug this is fixing?

On line 60 s/0./0/

dgryski:
@kortschak bug introduced in https://github.com/gonum/plot/pull/26

kortschak:
Thanks, @dgryski.

@ctessum, please add to the commit message that this fixes a bug that was uncovered by #26.

Some kind of test would be good too.

ctessum:
I'm not sure how to add a test that doesn't require visual inspection for this.

kortschak:
We can cope with a small png.

kortschak:
Closing because we have a competing proposal with a test in #182.

PR #180, opened 2015-03-21
Fix use of RW mutex

kortschak:
Concurrent add operations would have potentially resulted in memory corruption.


kortschak:
@gonum/developers

eaburns:
(Github seems to be down for me. Sorry if you get this twice.)
LGTM. How'd that happen?

kortschak:
It was provided code see [issue plotinum issue 148](https://code.google.com/p/plotinum/issues/detail?id=148).

PR #183, opened 2015-03-25
plot: make Save take locale agnostic units

kortschak:
Fixes #135.

@eaburns

PR #184, opened 2015-03-26
Make example code match Save unit semantics

kortschak:
Semantically the same as prior to #183.

@btracey @eaburns


btracey:
LGTM

PR #185, opened 2015-03-26
plotter: reduce repetition in example main

kortschak:


plotter/main.go
```diff
@@ -22,74 +24,47 @@ import (
...
+	{"example_points", Example_points()},
+	{"example_errBars", Example_errBars()},
+	{"example_bubbles", Example_bubbles()},
+	{"example_histogram", Example_histogram()},
+	{"example_barChart", Example_barChart()},
+	{"example_stackedBarChart", Example_stackedBarChart()},
+	{"example_heatMap", Example_heatMap()},
+}
+
+var formats = []string{
```
eaburns: I like using [...] to make this an array. It's a simpler type. Though it really won't matter at all in practice, so feel free to ignore me.

eaburns:
LGTM

plotter/main.go
kortschak: It may or may not be simpler in the runtime depending on the degree of optimisation - if it's an array and the compiler doesn't notice that its a literal, or doesn't use that information then the array will be copied prior to executing the for loop.

eaburns: I wasn't worried about optimization. It's fine.

PR #182, opened 2015-03-25
Fix vgimage translate.

btracey:
The old code created scatter plots which were incorrectly offset on the canvas. This makes the correct offset.

Fixes Issue 179.

PR #186, opened 2015-03-26
vg/...: don't stroke paths with width <= 0

### LINKED ISSUE #143: vg with negative widths and lengths
Original [issue 116](https://code.google.com/p/plotinum/issues/detail?id=116) created by eaburns on 2013-02-21T15:28:55.000Z:

vg shouldn't draw things when the line width or dash pattern has negative values.  This should be documented.  The current back-ends should be updated to handle it correctly.

kortschak:
Also improve vg.Canvas documentation.

Fixes #143.

vg/recorder is left untouched since it is also included as a debugging tool and negative widths may be something people want to see.

Wondering about tests?

@eaburns


vg/vg.go
```diff
@@ -15,24 +15,21 @@ import (
 // graphics.  The origin is in the bottom left corner.
 type Canvas interface {
 	// SetLineWidth sets the width of stroked paths.
-	// If the width is set to 0 then stroked lines are
-	// not drawn.
+	// If the width is zero or less then stroked lines
```
eaburns: not positive

vg/vg.go
```diff
@@ -42,15 +39,15 @@ type Canvas interface {
 	// Rotate applies a rotation transform to the
 	// context.  The parameter is specified in
 	// radians.
-	Rotate(float64)
+	Rotate(rad float64)
```
eaburns: Thanks!

eaburns:
LGTM

vg/vg.go
kortschak: Was avoiding double negatives.

kortschak:
Tests blocked on #182.

btracey:
Sorry, I thought I had merged it.

kortschak:
PTAL

vg/gen_width.go
```diff
@@ -0,0 +1,63 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//+build ignore
+
```
eaburns: Add a comment describing what this program does.

vg/vg.go
```diff
@@ -15,24 +15,21 @@ import (
...
-	// If the width is set to 0 then stroked lines are
-	// not drawn.
+	// If the width is not positive then stroked lines
+	// are not drawn.
 	//
 	// The initial line width is 1 point.
 	SetLineWidth(Length)
 
 	// SetLineDash sets the dash pattern for lines.
-	// The first argument is the pattern (units on,
```
eaburns: Why remove this part of the comment?

vg/vg_test.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/gonum/plot/vg/vgpdf"
+	"github.com/gonum/plot/vg/vgsvg"
+)
+
+// TestLineWidth tests output against test images generated by
+// gen_width.go - these images have been visually inspected and
+// validated.
+func TestLineWidth(t *testing.T) {
+	formats := []string{
+		// TODO: Add logic to cope with run to run eps differences.
```
eaburns: I don't understand this TODO, could you please clarify?

vg/vg_test.go
```diff
@@ -0,0 +1,119 @@
...
+		"tiff",
+		"jpg",
+	}
+
+	const (
+		width  = 100
+		height = 100
+	)
+
+	for _, w := range []vg.Length{-1, 0, 1} {
```
eaburns: How would you feel about parsing the width and format out of the testdata file name? Then if you add new formats you don't have to modify both the gen script and this file.

eaburns: Nevermind. I didn't see what you had to do below.

vg/vg_test.go
```diff
@@ -0,0 +1,119 @@
...
+			}
+
+			var c interface {
+				vg.Canvas
+				Size() (w, h vg.Length)
+				io.WriterTo
+			}
+
+			name := fmt.Sprintf("width_%v.%s", w, typ)
+			switch typ {
```
eaburns: We need to open up an issue to get this logic into a method in plot so that we can replace plot.Save with plot.Write. Then this test can call that method instead.

eaburns:
I'm a bit sad about the duplication between the test code and the gen script. Can we think of a way to reduce that? If not, fine. Still LGTM.

kortschak:
Yeah, I thought about that. The solution I came up with other than this one were to have a build tag that would cause the test data to be created prior to running tests - thus guaranteeing a pass on that run. I can do that if you prefer.

vg/vg_test.go
kortschak: I think that is already an issue, but if not, sure. Note that this sort of conflicts with a comment in the redesign wiki:

>Additionally, we may want to move the Plot.Save function to the plotutil, as it does a form of run-time checking in that it looks at the file extension to determine which vg.Canvas should be used to save the plot.

vg/vg_test.go
kortschak: eps includes creation time data.

The logic would involve stripping that from the testdata file and the buffered bytes.

vg/vg.go
kortschak: It is now implicit in the method arguments. With both it read as a stutter.

vg/gen_width.go
kortschak: Done.

kortschak:
PTAL

If you prefer the second commit state, I'll squash and we go with that. Otherwise I'll drop that and go with the original.

vg/vg_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+			default:
+				panic("bad typ string")
+			}
+
+			p.Draw(draw.New(c))
+
+			var buf bytes.Buffer
+			if _, err = c.WriteTo(&buf); err != nil {
+				t.Errorf("failed to write plot for %v:%s: %v", w, typ, err)
```
kortschak: This should be Fatalf.

vg/vg.go
eaburns: I guess that the part I was concerned with was the description of the pattern: units on first, units off second. But perhaps that's obvious?

vg/vg_test.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/gonum/plot/plotter"
+	"github.com/gonum/plot/vg"
+	"github.com/gonum/plot/vg/draw"
+	"github.com/gonum/plot/vg/vgeps"
+	"github.com/gonum/plot/vg/vgimg"
+	"github.com/gonum/plot/vg/vgpdf"
+	"github.com/gonum/plot/vg/vgsvg"
+)
+
+var generateTestData = flag.Bool("good", false, "Specifies that the current state is good.")
```
eaburns: --regen, "Uses the current state to regenerate the test data."

I don't care about the details of the wording, but it needs to say that it regenerates the test data.

vg/vg_test.go
```diff
@@ -0,0 +1,129 @@
...
+				c = vgimg.PngCanvas{Canvas: vgimg.New(width, height)}
+
+			case "svg":
+				c = vgsvg.New(width, height)
+
+			case "tiff":
+				c = vgimg.TiffCanvas{Canvas: vgimg.New(width, height)}
+
+			default:
+				panic("bad typ string")
```
eaburns: "bad typ string: " + typ

vg/vg_test.go
```diff
@@ -0,0 +1,129 @@
...
+				err = p.Save(width, height, name)
+				if err != nil {
+					log.Fatalf("failed to save %q: %v", name, err)
+				}
+			}
+
+			f, err := os.Open(name)
+			if err != nil {
+				t.Fatalf("failed to open test image: %v", err)
+			}
```
eaburns: defer f.Close()

eaburns:
LGTM

vg/vg_test.go
kortschak: defered close is no good here - in a loop and early exits are exits, not returns.

I'll add f.Close after the ReadAll call.

eaburns: It's fine to defer close in a loop and it keeps with the idiom, but closing
after the read is just great. The point is: don't forget to close.

kortschak: In this case I'm not actually all that fussed about it. We create 15 files and the test exit - the fds will be recovered when the test exits. But sure, it's nicer to be clean.

kortschak:
PTAL

eaburns:
Yes. LGTM.
(In the future, I'll be more conservative with "LGTM"ing code with unaddressed comments.)

PR #187, opened 2015-03-27
plot: add WriterTo method to Plot

kortschak:
This is a first stab at implementing a generalised plot writing method as suggested [here](https://github.com/gonum/plot/pull/186#discussion_r27268402).

I'm not completely happy with it for a couple of reasons: 1. we lose the title in eps (I'm less concerned about this) and 2. it's not extensible.

Addressing the second point. If we have a map[string]func(w, h vg.Length) interface { vg.CanvasSizer; io.WriterTo } that can be registered to, then this problem goes away. This map must be in vg, and referenced from the WriterTo method. Then each vg implementation that can make a WriterTo will register in an init.

### DIFF
--- a/plot.go
+++ b/plot.go
@@ -439,49 +439,71 @@ func (p *Plot) NominalY(names ...string) {
 	p.Y.Tick.Marker = ConstantTicks(ticks)
 }
 
+// WriterTo returns an io.WriterTo that will write the plot as
+// the specified image format.
+//
+// Supported formats are:
+//
+//  eps, jpg|jpeg, pdf, png, svg, and tif|tiff.
+func (p *Plot) WriterTo(w, h vg.Length, format string) (io.WriterTo, error) {
+	var c interface {
+		vg.CanvasSizer
+		io.WriterTo
+	}
+	switch format {
+	case "eps":
+		c = vgeps.New(w, h)
+
+	case "jpg", "jpeg":
+		c = vgimg.JpegCanvas{Canvas: vgimg.New(w, h)}
+
+	case "pdf":
+		c = vgpdf.New(w, h)
+
+	case "png":
+		c = vgimg.PngCanvas{Canvas: vgimg.New(w, h)}
+
+	case "svg":
+		c = vgsvg.New(w, h)
+
+	case "tif", "tiff":
+		c = vgimg.TiffCanvas{Canvas: vgimg.New(w, h)}
+
+	default:
+		return nil, fmt.Errorf("unsupported format: %q", format)
+	}
+	p.Draw(draw.New(c))
+
+	return c, nil
+}
+
 // Save saves the plot to an image file.  The file format is determined
 // by the extension.
 //
 // Supported extensions are:
 //
-//  .eps, .jpg, .jpeg, .pdf, .png, .svg, and .tiff.
+//  .eps, .jpg, .jpeg, .pdf, .png, .svg, .tif and .tiff.
 func (p *Plot) Save(w, h vg.Length, file string) (err error) {
-	var c interface {
-		vg.Canvas
-		Size() (w, h vg.Length)
-		io.WriterTo
-	}
-	switch ext := strings.ToLower(filepath.Ext(file)); ext {
-
-	case ".eps":
-		c = vgeps.NewTitle(w, h, file)
-
-	case ".jpg", ".jpeg":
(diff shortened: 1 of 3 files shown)


kortschak:
@eaburns

plot.go
```diff
@@ -439,48 +439,65 @@ func (p *Plot) NominalY(names ...string) {
...
+// by the extension.
+//
+// Supported extensions are:
+//
+//  .eps, .jpg, .jpeg, .pdf, .png, .svg, .tif and .tiff.
+func (p *Plot) Save(w, h vg.Length, file string) error {
 	f, err := os.Create(file)
 	if err != nil {
 		return err
 	}
```
sbinet: perhaps add:
```go
defer f.Close()
```
after we checked the call to ``os.Create`` went fine.
so no file descriptor will ever be leaked.

otherwise, an error during ``p.WriterTo`` or ``c.WriteTo`` would leak `f` descriptor.

if everything went fine, ``return f.Close()`` will run correctly (returning a possibly nil error) and the defer will run - fail as closing a file twice is an error - but won't overwrite the error.

plot.go
```diff
@@ -439,49 +439,71 @@ func (p *Plot) NominalY(names ...string) {
...
+// WriterTo returns an io.WriterTo that will write the plot as
+// the specified image format.
 //
-// Supported extensions are:
+// Supported formats are:
 //
-//  .eps, .jpg, .jpeg, .pdf, .png, .svg, and .tiff.
-func (p *Plot) Save(w, h vg.Length, file string) (err error) {
+//  eps, jpg|jpeg, pdf, png, svg, and tif|tiff.
+func (p *Plot) WriterTo(w, h vg.Length, format string) (io.WriterTo, error) {
```
eaburns: An alternative would be to just return an io.WriterTo, but in the case of the error return:
type errWriterTo error
func (e errWriterTo) WriteTo(w io.Writer) (int64, error) { return 0, e }

Maybe it's a bit round-about, but I thought that I'd mention it.

plot.go
```diff
@@ -439,49 +439,71 @@ func (p *Plot) NominalY(names ...string) {
...
+//
+//  .eps, .jpg, .jpeg, .pdf, .png, .svg, .tif and .tiff.
+func (p *Plot) Save(w, h vg.Length, file string) (err error) {
 	f, err := os.Create(file)
 	if err != nil {
 		return err
 	}
-	if _, err = c.WriteTo(f); err != nil {
+	defer func() {
+		e := f.Close()
```
eaburns: This could me the initializing statement of the following if.

eaburns:
Let me think about this a bit.

plot.go
kortschak: I'm not a fan of `if exp; cond {` unless it's necessary.

plot.go
kortschak: We signal early this way. Are you suggesting this so we can `p.WriterTo(...).WriteTo(w)`?

kortschak:
No worries. I'll pick this up again tomorrow.

plot.go
eaburns: Yeah. It was just a brainstorm, but basically it avoids an extra error check:
```wt, err := p.WriterTo(...)
if err != nil {
    whatever
}
if _, err := wt.WriteTo(w); err != nil {
   whatever else
}```

becomes what you wrote:
```
if _, err := p.WriterTo(...).WriteTo(w); err != nil {
    whatever
}
```

plot.go
eaburns: I like it, but I can understand why you may not. I won't bug about it in future PRs.

plot.go
kortschak: The fluidity is nice, but you lose flexibility over the order of your priorities.

In the case of Save, I decide that if the file can't be openned, there is no point doing the plot work, but I could under other circumstances decide that there is no point getting a writer if the plot work fail. That option is lost.

I'm not sure which is more important here. I'll sleep on it and hear any views you have in the morning.

kortschak:
Ping

eaburns:
Sorry. I'll certainly look at it by this weekend.

eaburns:
Sorry for the delay.
If you are happy with WriterTo returning an error then so am I.
If something comes up that makes us change our minds then we can always change it.
LGTM.

PR #188, opened 2015-03-27
vgx11: fix translate offset

sbinet:
Address issue #179  by translating the image to (0,-h) so (0,0) is at the
bottom-left corner for the X11 backend.

``go test ./...`` will skip the X11 test if ``$DISPLAY`` is not defined.

### DIFF
--- a/vg/vgx11/canvas.go
+++ b/vg/vgx11/canvas.go
@@ -4,8 +4,7 @@
 
 // +build !windows
 
-// The vgx11 package uses xgbutil (github.com/BurntSushi/xgbutil)
-// as a X11-display backend for vg.
+// Package vgx11 implements X-Window vg support.
 package vgx11
 
 import (
@@ -69,7 +68,7 @@ func NewImage(img draw.Image, name string) (*Canvas, error) {
 	gc := draw2d.NewGraphicContextWithPainter(ximg, painter)
 	gc.SetDPI(dpi)
 	gc.Scale(1, -1)
-	gc.Translate(+0.5*w, -0.5*h)
+	gc.Translate(0, -h)
 
 	wid := ximg.XShowExtra(name, true)
 	go func() {
--- a/vg/vgx11/canvas_test.go
+++ b/vg/vgx11/canvas_test.go
@@ -1,0 +1,69 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build !windows
+
+package vgx11
+
+import (
+	"bytes"
+	"image/jpeg"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/gonum/plot"
+	"github.com/gonum/plot/plotter"
+	"github.com/gonum/plot/vg"
+	"github.com/gonum/plot/vg/draw"
+)
+
+// hasX11 checks if X11 is available by looking at the DISPLAY environment variable
+func hasX11() bool {
+	return os.Getenv("DISPLAY") != ""
+}
+
+func TestIssue179(t *testing.T) {
+	if !hasX11() {
+		t.Skip("no X11 environment")
+	}
+
+	scatter, err := plotter.NewScatter(plotter.XYs{{1, 1}, {0, 1}, {0, 0}})
+	if err != nil {
+		t.Fatalf("error: %v\n", err)
+	}
+	p, err := plot.New()
+	if err != nil {
+		t.Fatalf("error: %v\n", err)
+	}
+	p.Add(scatter)
+	p.HideAxes()
+
+	c, err := New(5.08*vg.Centimeter, 5.08*vg.Centimeter, "test issue179")
+	if err != nil {
+		t.Fatalf("error: %v\n", err)
+	}
+
+	p.Draw(draw.New(c))
+	b := bytes.NewBuffer([]byte{})
+	err = jpeg.Encode(b, c.ximg, nil)
+	if err != nil {
+		t.Error(err)
+	}
+
+	f, err := os.Open(filepath.Join("..", "vgimg", "testdata", "issue179.jpg"))
+	if err != nil {
+		t.Error(err)
+	}
+	defer f.Close()
+
+	want, err := ioutil.ReadAll(f)
+	if err != nil {
+		t.Error(err)
+	}
+	if !bytes.Equal(b.Bytes(), want) {
+		t.Error("Image mismatch")
+	}
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,10 +10,12 @@ go:
 env:
   global:
    - secure: "GYPf9iqbY/5YLmorhVxH2Q4Y8/kxF3MWxDau8Sb2QnineLAJZA2HKi4aWdZIoK8AMAJgXdHUC4Pw2jywp9nC2Q+TW1+57tp0uLLIdghuwif78oyqykhlt11ny+SUk5PAkqYmm+dCbA0kxShg1hFFfHGTTiDuWklTWomxGGICSPc="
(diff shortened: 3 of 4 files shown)


vg/vgx11/canvas_test.go
```diff
@@ -0,0 +1,58 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build !windows, x11
```
kortschak: No space after comma.

vg/vgx11/canvas_nox11.go
```diff
@@ -0,0 +1,11 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build windows
+
+// The vgx11 package uses xgbutil (github.com/BurntSushi/xgbutil)
```
kortschak: Package vgx11...

(also in canvas.go).

kortschak:
I feel unsatisfied by the fact that the test is there but not run. It seems to me that a better approach would be to more heavily rely on vgimg (possibly extending it) to create the vgimg.Canvas and thereby trust the test for its creation in vg.

vg/vgx11/canvas_test.go
sbinet: done

vg/vgx11/canvas_nox11.go
sbinet: done

sbinet:
alternatively, I could try to have the travis-ci actually run the X11 test.

it seems to me like the approach you're advising would be more "testing vgimg" or "testing how xgbutil is using image" rather than actually testing how an X11 window is being displayed.

kortschak:
I'm happy if you can get travis to test the X11 behaviour.

sbinet:
PTAL.

now the test is run if ``$DISPLAY`` is defined (which should work as a poor man's way to detect whether X11 is available. more refined ways could be devised if/when deemed necessary)

kortschak:
Do we need the windows build now? It seems not.

sbinet:
you mean the windows tag in the `canvas_test.go` file?
probably not.

kortschak:
The entire canvas_nox11.go file.

kortschak:
The !windows constraint does need to be there - not for us, but for people on Windows who do go test themselves.

sbinet:
the `canvas_nox11.go` file still needs to be there I think.
at least, when I tested on my linux+darwin machines with:
``$ go test -v -tags=windows ./...``
it would fail (w/o that canvas_nox11 file) complaining there were no source files for package vgx11 (paraphrasing.)

you are right though that even with the `hasX11` function, the ``!windows`` build tag is still necessary (otherwise, compilation of the test binary would fail. d'oh!)

kortschak:
If there is only source for a non windows build environment the directory will be skipped when GOOS=windows.

sbinet:
ok. I wasn't indeed completely sure this wasn't an artifact of my unorthodox way of simulating a Windows environment.

kortschak:
Nope, you're right. That does not seem reasonable. But it is what it is.

sbinet:
ok. removed

kortschak:
I going to have a look at build and see why this is how it behaves.

vg/vgx11/canvas.go
```diff
@@ -4,7 +4,7 @@
 
 // +build !windows
 
-// The vgx11 package uses xgbutil (github.com/BurntSushi/xgbutil)
+// Package vgx11 uses xgbutil (github.com/BurntSushi/xgbutil)
```
kortschak: Sorry, I'd really like this to be "Package vgx11 implements X11 vg support." or words to that effect. The basis for the implementation is secondary and should not be in the first sentence.

kortschak:
I think it is a result of doing -tags=windows. When I change the build constraints to windows (without the bang), I get that error if the test is invoked in vgx11, but not if I invoke go test ./... in plot. I think we are good.

LGTM apart from minor comment. Please wait for @eaburns though.

vg/vgx11/canvas.go
sbinet: done. (I was just following the other vg-foo backends description)

vg/vgx11/canvas_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+	"github.com/gonum/plot"
+	"github.com/gonum/plot/plotter"
+	"github.com/gonum/plot/vg"
+	"github.com/gonum/plot/vg/draw"
+)
+
+// hasX11 checks if X11 is available by looking at the DISPLAY environment variable
+func hasX11() bool {
+	d := os.Getenv("DISPLAY")
```
eaburns: return os.Getenv("DISPLAY") != ""

vg/vgx11/canvas_test.go
```diff
@@ -0,0 +1,68 @@
...
+		t.Fatalf("error: %v\n", err)
+	}
+	p, err := plot.New()
+	if err != nil {
+		t.Fatalf("error: %v\n", err)
+	}
+	p.Add(scatter)
+	p.HideAxes()
+
+	c, err := New(5.08*vg.Centimeter, 5.08*vg.Centimeter, "test issue179")
```
eaburns: I get the joke, but we don't have to do it everywhere.

vg/vgx11/canvas_test.go
```diff
@@ -0,0 +1,68 @@
...
+	err = jpeg.Encode(b, c.ximg, nil)
+	if err != nil {
+		t.Error(err)
+	}
+
+	f, err := os.Open(filepath.Join("..", "vgimg", "testdata", "issue179.jpg"))
+	if err != nil {
+		t.Error(err)
+	}
+	want, err := ioutil.ReadAll(f)
```
eaburns: defer f.Close()

vg/vgx11/canvas_test.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	p.Draw(draw.New(c))
+	b := bytes.NewBuffer([]byte{})
+	err = jpeg.Encode(b, c.ximg, nil)
+	if err != nil {
+		t.Error(err)
+	}
+
+	f, err := os.Open(filepath.Join("..", "vgimg", "testdata", "issue179.jpg"))
```
eaburns: It looks like we are going to be doing this boilerplate `reading from a file and compare` in various tests. I think that it's time for a vgtest package that exports this function.

eaburns:
LGTM.
I assume that you two worked out all of the build issues.

vg/vgx11/canvas_test.go
kortschak: In this case we actually do - Seb is using the same image that Brendan created.

eaburns: Haha. Fair enough.

kortschak:
I'm satisifed that it's correct.

vg/vgx11/canvas_test.go
sbinet: done.

vg/vgx11/canvas_test.go
sbinet: done.

vg/vgx11/canvas_test.go
sbinet: yeah... at some point I was tinkering with the idea of folding vgx11 into vgimg directly so it could be handled like the JpegCanvas and co. (so the test could also be folded into vgimg_test...)

vg/vgx11/canvas_test.go
sbinet: (also fixed in vgimg_test)

sbinet:
I'll rebase, squash and make a nice commit message. (tomorrow)

Issue #160, opened 2015-03-17
Only error about missing fonts if text is actually drawn


eaburns:
Original [issue 132](https://code.google.com/p/plotinum/issues/detail?id=132) created by eaburns on 2013-07-08T11:26:19.000Z:

Currently, plot.New returns an error if it cannot find the default font.  There are two concerns here: 1) if the default font will remain unexported then this might as well be a panic, and 2) we should support drawing plots that do not contain any text if the fonts aren't actually installed.

My plan is: export the defaultFont variable, leave the error return on plot.New, and change plot and vg so that, if the fonts aren't installed, they only error if/when text is actually drawn.

PR #190, opened 2015-03-28
Plot with sticky-error

sbinet:
- adds an Error-list field to Plot
- make plot.New only return a `*Plot`
- check sticky error only during `plot.Plot.Save`

First stab at fixing #160

### DIFF
--- a/gob/gob_test.go
+++ b/gob/gob_test.go
@@ -45,14 +45,14 @@ func TestPersistency(t *testing.T) {
 	// Make a scatter plotter and set its style.
 	s, err := plotter.NewScatter(scatterData)
 	if err != nil {
-		panic(err)
+		t.Fatalf("error creating scatter plotter: %v\n", err)
 	}
 	s.GlyphStyle.Color = color.RGBA{R: 255, B: 128, A: 255}
 
 	// Make a line plotter and set its style.
 	l, err := plotter.NewLine(lineData)
 	if err != nil {
-		panic(err)
+		t.Fatalf("error creating line plotter: %v\n", err)
 	}
 	l.LineStyle.Width = vg.Points(1)
 	l.LineStyle.Dashes = []vg.Length{vg.Points(5), vg.Points(5)}
@@ -61,7 +61,7 @@ func TestPersistency(t *testing.T) {
 	// Make a line plotter with points and set its style.
 	lpLine, lpPoints, err := plotter.NewLinePoints(linePointsData)
 	if err != nil {
-		panic(err)
+		t.Fatalf("error creating line-points plotter: %v\n", err)
 	}
 	lpLine.Color = color.RGBA{G: 255, A: 255}
 	lpPoints.Shape = draw.PyramidGlyph{}
--- a/gob/gob_test.go
+++ b/gob/gob_test.go
@@ -32,11 +32,7 @@ func TestPersistency(t *testing.T) {
 	lineData := randomPoints(n)
 	linePointsData := randomPoints(n)
 
-	p, err := plot.New()
-	if err != nil {
-		t.Fatalf("error creating plot: %v\n", err)
-	}
-
+	p := plot.New()
 	p.Title.Text = "Plot Example"
 	p.X.Label.Text = "X"
 	p.Y.Label.Text = "Y"
--- a/plot.go
+++ b/plot.go
@@ -64,8 +64,8 @@ type Plot struct {
 	// Legend is the plot's legend.
 	Legend Legend
 
-	// Errs is a list of encountered errors, if any
-	Errs Errors
+	// Err is the first encountered error, if any.
+	Err error
 
 	// plotters are drawn by calling their Plot method
 	// after the axes are drawn.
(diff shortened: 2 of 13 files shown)


sbinet:
This CL allows what was discussed at some point in the redesign document:

```go
p := plot.New()
```

instead of:
```go
p, err := plot.New()
handle(err) 
```

defering the error handling related with (mainly?) the loading of fonts to the ``plot.Plot.Save(...)`` call.
FTR, the same type of handling should probably done during ``plot.Plot.WriteTo`` if #187 is accepted.

kortschak:
I'm not conviced that Plot.Save needs this. There is only one case when an error will collide and that is when both WriterTo failed (arguably a panickable failure) and the file Write or Close failed. If the first errors, then the user should not be interested in the subsequent error. Collecting errors is an excellent way to overload the client (see C++ compilation errors as a case in point).

kortschak:
Further, I'm wondering if we need to keep a list of errors here rather than just returning with the sticky error. What does the list gain us?

I think a preferable approach would be to have a set of flags on Plot that indicate what initialisation operations have been successful and error when a subsequent operation depends on an unsuccessful initialisation op. This gets us the behaviour we want, don't fail completely unless the client has requested a behaviour we can't satisfy.

sbinet:
updated to only keep the first error.

sbinet:
wrt the error+flag: I am not completely sold. that would make the internal state of plot.Plot very finicky and complicated.

kortschak:
Not particularly complicated, but consider the driver for this issue:

We have a client who want to create a plot that has no text fo whatever reason (I think this was the actual request), and doesn't have the required fonts installed. Currently, this fails. With the change here, AFAICS, on return from plot.New, the Plot will have a non-nil Err field. The client can look at that and potentially choose to set it to nil, but this seems like we load the work onto them. If they don't set it to nil, we have the same behaviour as currently exists.

I can guarantee this will be a cause for repeated questions along the lines of "How do I prevent plot from worrying about fonts when I don't want them?" and misguided replies to set the Err field to nil, which will lead to other difficult to debug issues. In the long run, that will be more work.

sbinet:
another way to tackle this is to keep a list of errors of type:
```go
type ErrFlag struct {
 Err error
 Type ErrType
}
```
where `ErrType` is ``ErrTitle | ErrXAxis | ErrYAxis | ErrLegend``

and then erase these errors during `plot.Plot.Save` if the corresponding `Title.Text`, `X.Label.Text`, ... are the empty string. (and if a non-nil error remains, return early)

yet another way to tackle this, is to modify `vg.Font` to lazily load the `truetype.Font` and forcibly load it during `plot.Plot.Save`, only for `draw.TextStyle` associated with non-empty strings.
the two could be packaged into `vg/draw` as:
```go
type Text struct {
 Value string
 Style TextStyle
}
```

kortschak:
TBH, I prefer the lazy loading (though this would be during Plot.WriterTo), but I'd like to hear @eaburns view on this.

kortschak:
A third is to add an error return to Plotter.Plot.

eaburns:
I need to think about this, but my intuition is that we want to do lazy font loading. This probably means returning the font error from Plot.Plot.

sbinet:
ideally, if we could bundle a BSD-compliant truetype font with gonum/plot, then we could be sure that the (new) default font would always load correctly, do it during `init()` and allow to pass an optional `vg.Font` to `plot.New(...)`.

kortschak:
It would still be an error case if the font requested by the user id not exist or could not be loaded.

sbinet:
yes, but the error would have to be handled before passing `vg.Font` to `plot.New(...)`.
e.g.:

```go
fnt, err := vg.MakeFont("my-font", 12*vg.Point)
if err != nil { panic(err) }

p := plot.New(plot.SetFont(fnt))
p := plot.New(plot.SetTitleFont(fnt), plot.SetAxisFont(axisFnt), plot.SetLegendFont(legendFnt))
```

eaburns:
@sbinet, the problem with bundling an arbitrary truetype font is that it won't work with the EPS and PDF backends. Neither EPS or PDF support font embedding; they only support the fonts named as keys in `vg.FontMap`. The fonts in `vg.FontMap` are required by the Postscript standard, so it's guaranteed that a Postscript interpreter can render these fonts even if they are not included in an EPS file. That's why Plotinum font support is so assinine.

A long, sad story:

Plotinum computes font metrics (basically the bounding box around the text) using freetype-go. The metrics that Plotinum computes must exactly match the metrics computed by the output format (Postscript and PDF) or else the text will be placed incorrectly in the resulting image. This means that we need TTF files (used by freetype-go) which are an *exact* match for the font metrics of the fonts named in `vg.FontMap` (used by Postscript). It is exceedingly rare for two fonts to have the exact same metrics. Luckily, there *are* fonts with the exact same metrics as those listed as keys in `vg.FontMap`: the Nimbus font series from Ghostscript.

Ghostscript is a GNU Postscript interpreter and it needs fonts matching the metrics required by the Postscript standard. Bingo! Plotinum can use these fonts and be assured that the metrics it computes from the TTF files match the metrics computed by Postscript. The only downside is that the Nimbus fonts, being from GNU, are released under the GPL (whatever that means for artwork). So,---cliche: I am not a lawyer----we can't include them in the code, but we can distribute them as long as we make it clear that the font files are covered under the GPL. That's what we do.

For your solution to work, we would have to find not only BSD-compatible fonts, but ones that match the exact font metrics of Courier, Times, and Helvetica. I've looked for such fonts for years without luck.

The only other solution is to implement font embedding for EPS and PDF. This is hard. Both formats use Type 1 fonts, not Truetype fonts. So we would need to convert, which would require a deep understanding of two very complex file formats. (Actually, there may be extensions to support Truetype within PS and PDF but I've never looked into it in detail.) We should probably open an issue for this, but I think that it would be a very big project requiring lots of research into Postscript, PDF, Truetype fonts and Type 1 fonts. That's why I never did it.

kortschak:
If this is done with functional options that that is part of the way, but what happens when a user doesn't call these and then depends on them?

On 30/03/2015, at 7:20 AM, "Sebastien Binet" <notifications@github.com<mailto:notifications@github.com>> wrote:


yes, but the error would have to be handled before passing vg.Font to plot.New(...).
e.g.:

fnt, err := vg.MakeFont("my-font", 12*vg.Point)
if err != nil { panic(err) }

p := plot.New(plot.SetFont(fnt))
p := plot.New(plot.SetTitleFont(fnt), plot.SetAxisFont(axisFnt), plot.SetLegendFont(legendFnt))

kortschak:
@eaburns Looking at vg FillText, the bahviour in these error cases is not consistent. If we go down the Plot-returns-error path then we can either make FillText return an error or explicitly state that it will panic (some do, some don't) - there are implications on plot.Plot if an error is returned wrt error handling there (I think it brings back a Plot []error field - a lá errors in LaTeX).

Either way, I think this discussion needs to happen at gonum-dev rather than here during review.

sbinet:
just so I don't forget (abusing github-pr as a poor's man bookmark server)... 
here is how `gxui` is dealing with fonts:
https://github.com/google/gxui/blob/master/gxfont/mkfont.go

eaburns:
Yes. Let's start a thread on gonum-dev.

eaburns:
Did we ever make a decision here?

kortschak:
No gonum-dev thread was started as far as I recall.

sbinet:
actually, you did:
 https://groups.google.com/forum/#!msg/gonum-dev/g7ZgAVItbbA/_qF-ouPQfWwJ
:)

kortschak:
Thank you. I should retire to a nursing home soon.

eaburns:

PR #192, opened 2015-04-03
vg: use plot WriterTo in tests

kortschak:
@eaburns

PR #193, opened 2015-04-03
vg: add format registration

kortschak:
This moves away from hard coded format support and allows much easier
extension for canvas types.

Waiting on #192 - not tested until that change is merged.

@eaburns

### DIFF
--- a/vg/vg.go
+++ b/vg/vg.go
@@ -8,7 +8,10 @@
 package vg
 
 import (
+	"fmt"
 	"image/color"
+	"io"
+	"sync"
 )
 
 // A Canvas is the main drawing interface for 2D vector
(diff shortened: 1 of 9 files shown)


kortschak:
A question here is whether vg.NewCanvasWriterTo means we should drop plot.Plot.WriterTo which is now really just a convenience method now.

vg/vg.go
```diff
@@ -85,6 +88,49 @@ type CanvasSizer interface {
...
+// CanvasWriterTo wraps behavior required for writing
+// a Canvas representation to a concrete image.
+type CanvasWriterTo interface {
+	CanvasSizer
+	io.WriterTo
+}
+
+var (
+	formatLock sync.RWMutex
+	formats    = make(map[string]func(w, h Length) CanvasWriterTo)
```
sbinet: make it:
```go
func(w,h Length, title string) ...
```
instead? so eps, x11 and possibly others (opengl, html5/d3) can be "properly" called?

vg/vg.go
```diff
@@ -85,6 +88,49 @@ type CanvasSizer interface {
...
+)
+
+// Register allows a format to be registered with the vg package.
+// Registered formats can be returned by NewCanvasWriterTo.
+func Register(format string, new func(w, h Length) CanvasWriterTo) {
+	formatLock.Lock()
+	formats[format] = new
```
sbinet: panic if duplicates?
behaviour should at least be documented.

sbinet:
wrt ``plot.Plot.WriterTo`` I suppose we could unexport it and reassess later on.

vg/vg.go
sbinet: fyi, `sql.Register` panics if dupes or if `nil`.
http://golang.org/pkg/database/sql/#Register

vg/vg.go
kortschak: I'm really not sure about this. Can you explain what is required for opengl and html5/d3?

I don't really want to add a parameter that is not required for the majority of cases. Functional options may be OK if we have options be func(vg.Canvas) and the package-specific options are defined within the relevant package and assert to the correct type to perform the operation.

vg/vg.go
kortschak: OK, I'll do that. I also need to add to plot Save that the extension switch is case insensitive.

vg/vg.go
kortschak: The thinking that I had while doing this was that we should support easy use of common formats. I'm not completely convinced that this includes OpenGL or X11. I'm troubled by how I'm dealing with EPS, but not hugely. Even if we add something like functional options, we still need to make some formats special - we need to assert on the returned Canvas or make decisions on the basis of the format string (and then provide information that we don't actually have under the current model - Plot.WriterTo would need to be extended to take the functional options to get this).

This needs some thought.

sbinet: depends on the use case, I suppose.
for articles plots, yeah, they are the "odd cousins".
for interactive plots, they're definitely the more useful and common formats.

speaking of which: it seems the naming has some cognitive dissonance. you speak of formats, I speak (as per my "vg/drivers" thread on gonum-dev) of drivers.
formats does hint at a disk or persistent representation.
drivers is more generic (which may or may not be a good thing)

sbinet: wrt the title: both openGL and X11 require a title for the new window in which the canvas will live.
as for EPS, an empty string may do, but it feels odd.
perhaps as odd as giving an empty string to PDF, PNG, ... factory functions, though.

if only one could connect the Title of the `plot.Plot` to the canvas creation...

kortschak: You are right wrt the semantics behind 'format'. I'm not yet convinced that we should be covering all cases with WriterTo, though I certainly could be. Currently I am focusing specifically towards storage formats.

>if only one could connect the Title of the plot.Plot to the canvas creation...

That's entirely feasible. Do you think a panic on "" is reasonable for OpenGL/X11, or is that not necessary?

sbinet: no, I don't think it's necessary.
it's customary (at least in the tools I have dealt with so far) to just generate a title (say, "gonum/plot-x11 [counter]") for these nameless windows.

eaburns:
Refer to the discussion about registration here: https://groups.google.com/d/topic/gonum-dev/DsL5sy2_QWE/discussion.

PR #196, opened 2015-04-18
all: use github.com/llgcode/draw2d

sbinet:
- `code.google.com/p/xyz` will be retired
- migrate to the new home of `draw2d`


sbinet:
@llgcode let me/us know when will the new home of [draw2d](https://github.com/llgcode/draw2d) be official.

PR #197, opened 2015-04-18
vg opengl

sbinet:
DO NOT SUBMIT

- first stab at an OpenGL-based backend.
- use `github.com/llgcode/draw2d/draw2dgl`

### DIFF
--- a/vg/vgl/canvas.go
+++ b/vg/vgl/canvas.go
@@ -1,0 +1,348 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package vgl
+
+import (
+	"fmt"
+	"image/color"
+	"runtime"
+	"time"
+
+	"github.com/go-gl/gl/v2.1/gl"
+	"github.com/go-gl/glfw/v3.1/glfw"
+	"github.com/gonum/plot/vg"
+	"github.com/llgcode/draw2d"
+	"github.com/llgcode/draw2d/draw2dgl"
+)
+
+// dpi is the number of dots per inch.
+const dpi = 96
+
+// Canvas implements the vg.Canvas interface, using OpenGL as a backend.
+type Canvas struct {
+	gc    draw2d.GraphicContext
+	win   *glfw.Window
+	w, h  vg.Length // width and height of the canvas
+	width vg.Length // width of the current line
+	color []color.Color
+}
+
+// make sure we implement vg.Canvas
+var _ vg.Canvas = (*Canvas)(nil)
+
+// channel where all OpenGL calls are made
+var glchan chan func()
+var done chan struct{}
+
+func init() {
+	// OpenGL calls need to be run on the main OS-thread.
+	// calling LockOSThread during init() ensures that is the case.
+	runtime.LockOSThread()
+}
+
+// Run runs OpenGL commands.
+// Run needs to be run on the main OS-thread.
+func Run(f func() error) error {
+	err := gl.Init()
+	if err != nil {
+		return fmt.Errorf("vgl: could not initialize OpenGL context (%v)", err)
+	}
+
+	quit := make(chan struct{})
+
+	done = make(chan struct{})
+	glchan = make(chan func())
+	go func() {
+		err = f()
+		quit <- struct{}{}
+	}()
+
+	for {
+		select {
+		case fct := <-glchan:
+			fct()
+			done <- struct{}{}
+		case <-quit:
+			return err
+		}
+	}
+
+	return err
+}
+
+// call runs an OpenGL command on the main OS-thread
+func call(fct func()) {
+	glchan <- fct
+	<-done
+}
+
+// New returns a new canvas with the size specified rounded up to the nearest
+// pixel.
+func New(w, h vg.Length, name string) (*Canvas, error) {
+	ww := int(w/vg.Inch*dpi + 0.5)
+	hh := int(h/vg.Inch*dpi + 0.5)
+	var window *glfw.Window
+	var err error
+	call(func() {
+		window, err = newWindow(ww, hh, name)
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	c := &Canvas{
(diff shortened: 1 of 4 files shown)


sbinet:
@eaburns @kortschak 
I have been banging my head against this one for too long.
if you see why nothing is printed out when one "go run"s the vgl_example.go (except for the rounded rectangle) feel free to drop your thoughts over here :)

eaburns:
I wanted to mention that I saw this, and it's on my todo list. But I'm going in vacation, so I won't be able to get to it for a few weeks. I hope that's OK.

sbinet:
of course. enjoy your vacations!

vg/vgl/canvas.go
```diff
@@ -0,0 +1,348 @@
...
+}
+
+// make sure we implement vg.Canvas
+var _ vg.Canvas = (*Canvas)(nil)
+
+// channel where all OpenGL calls are made
+var glchan chan func()
+var done chan struct{}
+
+func init() {
```
kortschak: You have these in all three files. Surely you just need one in main().

kortschak:
I can't get this to work. I get a window popping up, but it's black - it also goes away straight away because the program terminates (I added a 10s sleep and it still doesn't draw anything).

Is there any example code in draw2dgl?

sbinet:
there is this one:
 https://github.com/llgcode/draw2d/blob/master/cmd/roundedrectgl.go

kortschak:
That image transiently appears for the example you generate.

sbinet:
I am not sure what image you are referring too nor which example :)

the `vgl.(*Canvas).Paint()` method will exit as soon as you hit a key on the keyboard, or move the mouse (ie: any "window" activity.)

kortschak:
The image that the draw2gl example you pointed to creates - a rounded
rectangle.

The example that you wrote for the vgl package.

sbinet:
ok. yes, as I said, the rounded rectangle does appear.
but neither the histogram nor the scatter plot do.

sorry, I should have been clearer in my original description.

vg/vgl/canvas.go
eaburns: Does it require the main thread, or is the initializing thread sufficient?
I ask because sdl2, which uses OpenGL under the hood, only requires the calls be made by the thread creating the context, not the main thread. A comment in this bug agrees: https://bugzilla.libsdl.org/show_bug.cgi?id=1995. If that's the case, we can avoid this init requirement and instead spawn the Run go routine and call LockOSThread when the canvas is created.

vg/vgl/canvas.go
```diff
@@ -0,0 +1,348 @@
...
+	"fmt"
+	"image/color"
+	"runtime"
+	"time"
+
+	"github.com/go-gl/gl/v2.1/gl"
+	"github.com/go-gl/glfw/v3.1/glfw"
+	"github.com/gonum/plot/vg"
+	"github.com/llgcode/draw2d"
+	"github.com/llgcode/draw2d/draw2dgl"
```
eaburns: Does draw2d/draw2dgl actually use OpenGL to do rasterization? Looking at the code, it looks like it uses the freetype-go software rasterizer: https://github.com/llgcode/draw2d/blob/master/draw2dgl/gc.go#L9. Way back, I did an experiment using software rendering for interactive plotting with Plotinum. It was too slow. See: https://code.google.com/p/eaburns/source/browse/showplot/main.go.

If we provide an OpenGL implementation of vg, I think that it should use true hardware rasterization.

vg/vgl/canvas.go
```diff
@@ -0,0 +1,348 @@
...
+
+func init() {
+	// OpenGL calls need to be run on the main OS-thread.
+	// calling LockOSThread during init() ensures that is the case.
+	runtime.LockOSThread()
+}
+
+// Run runs OpenGL commands.
+// Run needs to be run on the main OS-thread.
+func Run(f func() error) error {
```
eaburns: What about an interface like this:
`NewWindow(w, h int, title string) Window`
`Window.Draw(f func(Canvas)error) error`
This way, you can draw multiple times to the same window.
Also, it's more clear that `Draw` expects `f` to draw to the given canvas.
With `Run`, it's not obvious from the API that you are expected to create a canvas to which you draw within `f`.

eaburns: By the way, I've this before to work around the annoying OpenGL thread limitation. I am pretty happy with it: https://github.com/eaburns/T/tree/master/ui.

vg/vgl/canvas.go
```diff
@@ -0,0 +1,348 @@
...
+
+func (c *Canvas) Paint() {
+	call(glfw.WaitEvents)
+	call(c.paint)
+	call(glfw.WaitEvents)
+	//c.run()
+}
+
+func (c *Canvas) DPI() float64 {
+	return float64(c.gc.GetDPI())
```
eaburns: GetDPI always returns -1. vg.Length.Dots uses Canvas.DPI, so all calls to Dots will be wrong. But I changed this to return 96.0 and it still only draws the rectangle.

eaburns:
I spent a bunch of time playing around, but I couldn't figure it out either. I'll look again later.

vg/vgl/canvas.go
sbinet: according to https://github.com/golang/go/wiki/LockOSThread OpenGL does need to service calls from the main thread.
w/o it I vaguely remember having random crashes (happening at opengl call sites)

vg/vgl/canvas.go
sbinet: wrt `Window.Draw` and `vgl.Run`: ack.

vg/vgl/canvas.go
sbinet: yep. even with that fix, the coordinates transformation was still amiss.

sbinet:
I am tempted to just put it on the back-burner and focus on a `html5/d3` backend to provide interactivity.

eaburns:
If you want to back-burner this, should we close the PR?

sbinet:
Abandonned (will be done better with `golang.org/x/exp/shiny`)

PR #198, opened 2015-04-30
plotter: fix Johnson graph.subgraph method

kortschak:
Previously we did not remove any nodes under s if they were a predecessor.


eaburns:
LGTM, given my very limited understanding of this. I don't know this code well, so ideally someone more familiar should review it. But, it's pretty simple, so if you are fairly confident, please merge it.

kortschak:
It reflects the code now merged into gonum/graph (the more generic and heaver version of this code). I am confident it is correct.

Issue #18, opened 2015-02-12
plot: axis ticks near 0 render with unreasonable precision


kortschak:
This should be a relatively simple fix, just ensure that precisions agree along the entire axis (within reason).

PR #199, opened 2015-05-11
Fix label rounding for small tick value.

btracey:
There are floating point errors in computing the tick value, which are eliminated by rounding to a float32. If the value is close to zero, this doesn't remove the rounding errors. If abs(value) < 1, add 1, convert to float32, and then subtract one. This fixes cases where there are a lot of digits for the axis label at 0

### DIFF
--- a/axis.go
+++ b/axis.go
@@ -361,7 +361,12 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
 	val := math.Floor(min/majorDelta) * majorDelta
 	for val <= max {
 		if val >= min && val <= max {
-			ticks = append(ticks, Tick{Value: val, Label: fmt.Sprintf("%g", float32(val))})
+			round := float32(val)
+			if val < 1 && val > -1 {
+				round = float32(val + 1)
+				round -= 1
+			}
+			ticks = append(ticks, Tick{Value: val, Label: fmt.Sprintf("%g", round)})
 		}
 		if math.Nextafter(val, val+majorDelta) == val {
 			break
--- a/axis_test.go
+++ b/axis_test.go
@@ -1,0 +1,31 @@
+package plot
+
+import "testing"
+
+func TestAxisSmallTick(t *testing.T) {
+	d := DefaultTicks{}
+	for _, test := range []struct {
+		Min, Max float64
+		Labels   []string
+	}{
+		{
+			Min:    -1.9846500878911073,
+			Max:    0.4370974820125605,
+			Labels: []string{"-1.6", "-0.8", "0"},
+		},
+	} {
+		ticks := d.Ticks(test.Min, test.Max)
+		var count int
+		for _, tick := range ticks {
+			if tick.Label != "" {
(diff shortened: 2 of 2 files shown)


btracey:
This would fix #18

eaburns:
Can you please add "Fixed #18" to the commit message?

btracey:
Done.

axis.go
```diff
@@ -361,7 +361,12 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
 	val := math.Floor(min/majorDelta) * majorDelta
 	for val <= max {
 		if val >= min && val <= max {
-			ticks = append(ticks, Tick{Value: val, Label: fmt.Sprintf("%g", float32(val))})
+			round := float32(val)
+			if val < 1 && val > -1 {
+				round = float32(val + 1)
+				round -= 1
+			}
+			ticks = append(ticks, Tick{Value: val, Label: fmt.Sprintf("%g", round)})
```
eaburns: Why not set `Value: round` too?

btracey: Well, first, this is trying to be a minimal change to the existing code. What values the ticks are actually at is a separate matter.

Second, probably not. if you float64(round) you get values that are less close to the tick than the original value.

kortschak:
TBH I would prefer that a fix to #18 relies less on float magic. Updates... would be better.

btracey:
TBH?

btracey:
I'm not sure what the general fix would be, but really nothing should rely on floating point magic. We could merge this to fix the symptom, and open up a second issue to fix the problem.

eaburns:
TBD == To Be Honest, I believe.

kortschak:
Mmmm, #18 is the problem, this updates that.

btracey:
Done.

kortschak:
Needs a `#` before the 18.

axis.go
eaburns: I don't understand your remark about float64(round). I was suggesting using the more-accurate, float32 for the actual tick location in addition to the string value. If the float32 is more accurate why not put the tick mark there? (And I agree that the error is so small that it won't change the display in any meaningful way, but I think that it's confusing to use different values for both if one's more accurate.)

Don't worry about the churn, this would be a very minor addition.

btracey: Well, changing to float32 would be a breaking change (for anyone using custom tick marks for example).

eaburns: OK. Now I understand your comment about float64(round). Thanks.

eaburns:
LGTM. If someone can come up with a less magic fix then that sounds great too.

kortschak:
I think in the long run we should support something like the precision and width fmt verb modifiers, but at the moment, it should be possible without any breaking change to find the mode precision for the class of tick (major or minor) and cap to that precision with rounding.

kortschak:
LGTM

btracey:
Travis seems to be down, so merging.

kortschak:
This broke plot/vg tests.

PR #201, opened 2015-05-14
Revert "Fix label rounding for small tick value."

kortschak:
Reverts gonum/plot#199

This change causes the line thickness tests in plot/vg to break. The breakage is to round the 0.3 tick label to 0.29999995.

I think the bad zero is less bad than that.

@btracey 

Updates #18.


kortschak:
Fix posted.

PR #202, opened 2015-05-14
plot: fix near-zero rounding errors

### LINKED ISSUE #18: plot: axis ticks near 0 render with unreasonable precision

kortschak:
Fixes #18.

### DIFF
--- a/axis.go
+++ b/axis.go
@@ -5,13 +5,17 @@
 package plot
 
 import (
-	"fmt"
 	"image/color"
 	"math"
+	"strconv"
 
+	"github.com/gonum/floats"
 	"github.com/gonum/plot/vg"
 	"github.com/gonum/plot/vg/draw"
 )
+
+// displayPrecision is a sane level of float precision for a plot.
+const displayPrecision = 4
 
 // Ticker creates Ticks in a specified range
 type Ticker interface {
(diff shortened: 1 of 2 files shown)


axis.go
```diff
@@ -5,14 +5,17 @@
...
 	"image/color"
 	"math"
+	"strconv"
 
 	"github.com/gonum/plot/vg"
 	"github.com/gonum/plot/vg/draw"
 )
 
+// sane is a sane level of float precision for a plot.
+const sane = 4
```
sbinet: why not just call it `fprec` ? or `fltprec` ? `floatPrec`? 
I'd say `sane` is a bit insane :}

sbinet:
LGTM

eaburns:
Should round be exported so folks can use it for custom tickers?

kortschak:
I'd like to see it in the standard library - this has been asked for, but I don't think an issue was ever filed.

BTW The fix is incorrect at the moment. I'll push a commit to fix that in the next few minutes.

eaburns:
I also agree with Sebastian; can we rename sane to TickPrecision. Can we also make it a variable so that the default can be changed?

kortschak:
I've renamed sane, but I don't think it should be exported or a var.

PTAL

axis.go
kortschak: The word or the value? I'd say a precision of anything more than 1 part in 10,000 on a plot is probably more than necessary, but it is an upper bound on the precision.

kortschak:
Round might be a candidate for floats, @btracey?

sbinet:
still LGTM

btracey:
That seems like a reasonable addition

kortschak:
I overthought the precisionOf function. That is fixed now.

PTAL

@btracey, I'll send a PR for that, then remove it from here after it's merged.

kortschak:
Ugh, didn't overthink; under-thought. Floats are horrible.

kortschak:
OK. I think I'm done here. PTAL @sbinet @eaburns.

eaburns:
LGTM

sbinet:
LGTM

PR #203, opened 2015-05-14
{vgimg,vgx11}: update draw2d imports

kortschak:
Currently failing due to rounding error.


kortschak:
@sbinet has already posted this change

PR #210, opened 2015-06-15
Fix tics when the maximum tic is negative.

btracey:
Fixes Issue #210.


kortschak:
LGTM though tests on the tick generators (without plotting) would be nice.

kortschak:
That's a very meta comment.

btracey:
Submitting now to help fix the behavior.

How should the tests be done? I could test explicitly for the given values, but it seems like that would not be robust to changes in the tick selection mechanism.

kortschak:
Our current tests for this are not robust to this, so I don't see this as worse. Fine to submit though.

PR #211, opened 2015-07-18
added method for canvas tiles; changed plotter examples to tiles

ctessum:
I've created a method for tiling the canvas to make subfigures, and I have also changed the plotter examples to be put into subfigures instead of a bunch of individual files. PTAL.

### DIFF
--- a/plotter/main.go
+++ b/plotter/main.go
@@ -13,6 +13,7 @@ import (
 	"log"
 	"math"
 	"math/rand"
+	"os"
 
 	"github.com/gonum/matrix/mat64"
 
@@ -49,21 +50,53 @@ var examples = []struct {
 }
 
 var formats = []string{
-	".eps",
-	".pdf",
-	".svg",
-	".png",
-	".tiff",
-	".jpg",
+	"eps",
+	"pdf",
+	"svg",
+	"png",
+	"tiff",
+	"jpg",
 }
 
 func main() {
-	for _, ex := range examples {
-		for _, f := range formats {
-			err := ex.p.Save(4*vg.Inch, 4*vg.Inch, ex.name+f)
-			if err != nil {
-				log.Fatalf("failed to save %s%s: %v", ex.name, f, err)
+	const (
+		p     = 1 * vg.Centimeter
+		nrows = 4
+		ncols = 5
+	)
+	for _, f := range formats {
+		c, err := draw.NewFormattedCanvas(10*ncols*vg.Centimeter,
+			10*nrows*vg.Centimeter, f)
+		if err != nil {
+			log.Fatal(err)
+		}
+		dc := draw.New(c)
+		tiles := draw.Tiles{
+			Rows: nrows, Cols: ncols,
+			PadTop: p, PadBottom: p,
+			PadRight: p, PadLeft: p,
+			PadX: p, PadY: p,
+		}
+		ii := 0
+		for j := 0; j < nrows; j++ {
+			for i := 0; i < ncols; i++ {
+				if ii < len(examples) {
+					ex := examples[ii]
+					ex.p.Draw(tiles.At(dc, i, j))
+					if err != nil {
+						log.Fatalf("failed to draw tile for  %s.%s: %v", ex.name, f, err)
+					}
+					ii++
+				}
 			}
+		}
+		w, err := os.Create("examples." + f)
+		if err != nil {
+			log.Fatal(err)
+		}
+		_, err = c.WriteTo(w)
+		if err != nil {
+			log.Fatal(err)
 		}
 	}
 }
--- a/vg/draw/canvas.go
+++ b/vg/draw/canvas.go
@@ -5,11 +5,16 @@
 package draw
 
 import (
+	"fmt"
 	"image/color"
 	"math"
 	"strings"
 
 	"github.com/gonum/plot/vg"
+	"github.com/gonum/plot/vg/vgeps"
+	"github.com/gonum/plot/vg/vgimg"
+	"github.com/gonum/plot/vg/vgpdf"
+	"github.com/gonum/plot/vg/vgsvg"
 )
 
 // A Canvas is a vector graphics canvas along with
@@ -222,6 +227,39 @@ func New(c vg.CanvasSizer) Canvas {
 	return NewCanvas(c, w, h)
 }
 
+// NewFormattedCanvas creates a new vg.CanvasWriterTo with the specified
+// image format.
+//
+// Supported formats are:
+//
+//  eps, jpg|jpeg, pdf, png, svg, and tif|tiff.
+func NewFormattedCanvas(w, h vg.Length, format string) (vg.CanvasWriterTo, error) {
+	var c vg.CanvasWriterTo
+	switch format {
+	case "eps":
+		c = vgeps.New(w, h)
+
+	case "jpg", "jpeg":
+		c = vgimg.JpegCanvas{Canvas: vgimg.New(w, h)}
+
+	case "pdf":
+		c = vgpdf.New(w, h)
+
+	case "png":
+		c = vgimg.PngCanvas{Canvas: vgimg.New(w, h)}
+
+	case "svg":
+		c = vgsvg.New(w, h)
+
+	case "tif", "tiff":
+		c = vgimg.TiffCanvas{Canvas: vgimg.New(w, h)}
+
+	default:
+		return nil, fmt.Errorf("unsupported format: %q", format)
+	}
+	return c, nil
+}
+
 // NewCanvas returns a new (bounded) draw.Canvas of the given size.
 func NewCanvas(c vg.Canvas, w, h vg.Length) Canvas {
 	return Canvas{
@@ -246,7 +284,7 @@ func (c *Canvas) Contains(p Point) bool {
 	return c.ContainsX(p.X) && c.ContainsY(p.Y)
 }
 
-// Contains returns true if the Canvas contains the
+// ContainsX returns true if the Canvas contains the
 // x coordinate.
 func (c *Canvas) ContainsX(x vg.Length) bool {
 	return x <= c.Max.X+slop && x >= c.Min.X-slop
@@ -276,21 +314,57 @@ func (c *Canvas) Y(y float64) vg.Length {
 	return vg.Length(y)*(c.Max.Y-c.Min.Y) + c.Min.Y
 }
 
-// Crop returns a new Canvas corresponding to the receiver
-// area with the given number of inches added to the minimum
+// Crop returns a new Canvas corresponding to the Canvas
+// c with the given lengths added to the minimum
 // and maximum x and y values of the Canvas's Rectangle.
-func (c Canvas) Crop(minx, miny, maxx, maxy vg.Length) Canvas {
+// Note that cropping the right and top sides of the canvas
+// requires specifying negative values of right and top.
+func Crop(c Canvas, left, right, bottom, top vg.Length) Canvas {
 	minpt := Point{
-		X: c.Min.X + minx,
-		Y: c.Min.Y + miny,
+		X: c.Min.X + left,
+		Y: c.Min.Y + bottom,
 	}
 	maxpt := Point{
-		X: c.Max.X + maxx,
-		Y: c.Max.Y + maxy,
+		X: c.Max.X + right,
+		Y: c.Max.Y + top,
 	}
 	return Canvas{
-		Canvas:    vg.Canvas(c),
+		Canvas:    c,
 		Rectangle: Rectangle{Min: minpt, Max: maxpt},
+	}
+}
+
+// Tiles creates regular subcanvases from a Canvas.
+type Tiles struct {
+	// Cols and Rows specify the number of rows and columns of tiles.
+	Cols, Rows int
+	// PadTop, PadBottom, PadRight, and PadLeft specify the padding
+	// on the corresponding side of each tile.
+	PadTop, PadBottom, PadRight, PadLeft vg.Length
+	// PadX and PadY specify the padding between columns and rows
+	// of tiles respectively..
+	PadX, PadY vg.Length
+}
+
+// At returns the subcanvas within c that corresponds to the
+// tile at column x, row y.
+func (ts Tiles) At(c Canvas, x, y int) Canvas {
+	tileH := (c.Max.Y - c.Min.Y - ts.PadTop - ts.PadBottom -
+		vg.Length(ts.Rows-1)*ts.PadY) / vg.Length(ts.Rows)
+	tileW := (c.Max.X - c.Min.X - ts.PadLeft - ts.PadRight -
+		vg.Length(ts.Cols-1)*ts.PadX) / vg.Length(ts.Cols)
+
+	ymax := c.Max.Y - ts.PadTop - vg.Length(y)*(ts.PadY+tileH)
+	ymin := ymax - tileH
+	xmin := c.Min.X + ts.PadLeft + vg.Length(x)*(ts.PadX+tileW)
+	xmax := xmin + tileW
+
+	return Canvas{
+		Canvas: vg.Canvas(c),
+		Rectangle: Rectangle{
+			Min: Point{X: xmin, Y: ymin},
+			Max: Point{X: xmax, Y: ymax},
+		},
 	}
 }
 
@@ -333,14 +407,14 @@ func (c *Canvas) StrokeLine2(sty LineStyle, x0, y0, x1, y1 vg.Length) {
 	c.StrokeLines(sty, []Point{{x0, y0}, {x1, y1}})
 }
 
-// ClipLineXY returns a slice of lines that
+// ClipLinesXY returns a slice of lines that
 // represent the given line clipped in both
 // X and Y directions.
 func (c *Canvas) ClipLinesXY(lines ...[]Point) [][]Point {
 	return c.ClipLinesY(c.ClipLinesX(lines...)...)
 }
 
-// ClipLineX returns a slice of lines that
+// ClipLinesX returns a slice of lines that
 // represent the given line clipped in the
 // X direction.
 func (c *Canvas) ClipLinesX(lines ...[]Point) (clipped [][]Point) {
@@ -356,7 +430,7 @@ func (c *Canvas) ClipLinesX(lines ...[]Point) (clipped [][]Point) {
 	return
 }
 
-// ClipLineY returns a slice of lines that
+// ClipLinesY returns a slice of lines that
 // represent the given line clipped in the
 // Y direction.
 func (c *Canvas) ClipLinesY(lines ...[]Point) (clipped [][]Point) {
--- a/plot.go
+++ b/plot.go
@@ -16,7 +16,6 @@
 package plot
 
 import (
-	"fmt"
 	"image/color"
 	"io"
 	"math"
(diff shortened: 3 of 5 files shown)


sbinet:
nice!

plotter/main.go
```diff
@@ -49,22 +51,61 @@ var examples = []struct {
...
-	".jpg",
+	"eps",
+	"pdf",
+	"svg",
+	"png",
+	"tiff",
+	"jpg",
+}
+
+type canvasWriter interface {
```
sbinet: is that needed now with the CanvasWriterTo interface you expose in `vg`?

plotter/main.go
```diff
@@ -49,22 +51,61 @@ var examples = []struct {
...
+	//for _, ex := range examples {
+	//	for _, f := range formats {
+	//		err := ex.p.Save(4*vg.Inch, 4*vg.Inch, ex.name+"."+f)
+	//		if err != nil {
+	//			log.Fatalf("failed to save %s.%s: %v", ex.name, f, err)
+	//		}
+	//	}
+	//}
+	for _, f := range formats {
+		c, err := draw.NewFormattedCanvas(4*5*vg.Inch, 4*4*vg.Inch, f)
```
sbinet: perhaps lift and use the `nrows` and `ncols` constants here instead of these magic `4` and `5` litterals?

vg/draw/canvas.go
```diff
@@ -294,6 +332,34 @@ func (c Canvas) Crop(minx, miny, maxx, maxy vg.Length) Canvas {
 	}
 }
 
+// Tile returns new Canvases corresponding to the reciever canvas tiled
+// into the given number of rows and columns (row major, starting from the
+// top left corner) where t, b, l, r, h,
+// and w are the padding above, below, to the left, to the right, between rows,
+// and between columns of the tiles, respectively.
+func (c Canvas) Tile(rows, cols int, t, b, l, r, h, w vg.Length) [][]Canvas {
```
sbinet: perhaps rename it into `Tiles`?

sbinet: perhaps define:
```go
type Tiles struct {
 Values []Canvas
 Rows, Cols int
}

func (t Tiles) At(i,j int) Canvas { ... }
```
for more easily ranging through all tiles? (and still retain convenient direct access to each sub-canvas)

vg/draw/canvas.go
```diff
@@ -294,6 +332,34 @@ func (c Canvas) Crop(minx, miny, maxx, maxy vg.Length) Canvas {
...
+	o := make([][]Canvas, rows)
+	tileH := (c.Max.Y - c.Min.Y - t - b - vg.Length(rows-1)*h) / vg.Length(rows)
+	tileW := (c.Max.X - c.Min.X - l - r - vg.Length(cols-1)*w) / vg.Length(cols)
+	for j := 0; j < rows; j++ {
+		maxy := c.Max.Y - t - vg.Length(j)*(h+tileH)
+		miny := maxy - tileH
```
sbinet: I would rename them `xmax` and `ymin`

vg/draw/canvas.go
```diff
@@ -294,6 +332,34 @@ func (c Canvas) Crop(minx, miny, maxx, maxy vg.Length) Canvas {
...
+		o[j] = make([]Canvas, cols)
+		for i := 0; i < cols; i++ {
+			minx := c.Min.X + l + vg.Length(i)*(w+tileW)
+			maxx := minx + tileW
```
sbinet: ditto: `xmin` and `xmax`

sbinet:
3 comments:

- I am a bit concerned with the new dependencies of `vg/draw` against the `vg/eps` and co. packages. I, of course, offer no solution except for #193.

- what about plots with different sizes tiles? like: https://atlas.web.cern.ch/Atlas/GROUPS/PHYSICS/PAPERS/HIGG-2014-13/fig_01.png
should we try to come up with a unified way/API to create regular tiles and "non-regular" tiles?

- also: what about inserted plots? like this: http://matplotlib.org/examples/pylab_examples/axes_demo.html

plotter/main.go
ctessum: It doesn't work with the dots in there. It doesn't seem like it would have worked in the old version either but I didn't check.

ctessum:
There would be two ways to make subplots of different sizes. If you still wanted them to tile together, you just make multiple sets of tiles on the same canvas (e.g. tiles1 := c.Tile(2,2,...); tiles2 := c.Tile(4,4,...)), or you could just put a plot in the main canvas and then make the tiles over it. Otherwise, you can just use draw.Crop(), which is kind of conterintuitive because it requires you to specify the size of the margin rather than the bounds of the subplot, but it works. 

I don't have any great ideas about the vg dependencies, but the way it is in the master it is pretty cumbersome to change image formats.

plotter/main.go
sbinet: are we talking about the same thing? (I am talking about `canvasWriter` which happens to be exactly like `vg/draw.CanvasWriterTo`)

ctessum: Oh I see. Yes, canvasWriter is now redundant.

kortschak:
@vdobler has been doing work on something like this. would you please take a look here https://github.com/vdobler/plot-1/tree/redesign - also note his post to gonum-dev. He'll be back online soon I guess.

ctessum:
I hadn't seen that before but it looks interesting. I think the two solutions would have different use cases. His solution facilitates making a certain type of plot. The solution here is lower-level and could be one of the building blocks for making faceted plots but could also be used for other things. For instance, my current use-case that inspired writing it is making matrices of geographic maps using [this package](http://godoc.org/github.com/ctessum/geom/carto) which is based on vg.

kortschak:
It would be good to make sure we build something that integrates well, so probably you and Volker should discuss the an approach when he is back.

plotter/main.go
kortschak: >It doesn't work with the dots in there. It doesn't seem like it would have worked in the old version either but I didn't check.

I'm confused; Save and and WriteTo take difference strings semantically. Save takes a file name destination while WriteTo takes a format.

plotter/main.go
```diff
@@ -49,22 +51,61 @@ var examples = []struct {
...
+		if err != nil {
+			log.Fatal(err)
+		}
+		dc := draw.New(c)
+		const (
+			p     = 0.5 * vg.Inch
```
kortschak: Can this either be points or SI please.

vg/draw/canvas.go
kortschak: I'd prefer something like this too.

Also, I think changing the parameter names to: top, bottom, left, right, padx and pady would improve the docs (remember that people are likely to interact more with code completion suggestions than the godoc day to day.

plotter/main.go
ctessum: I see. That explains it.

ctessum:
PTAL. I left the function name as Tile() instead of Tiles() because tile is a verb and tiles is not, but I can change it if desired. I also changed the documentation to Crop() to be clearer (at least to me). @vdobler any thoughts?

plotter/main.go
```diff
@@ -49,22 +50,61 @@ var examples = []struct {
...
+	"jpg",
 }
 
 func main() {
-	for _, ex := range examples {
-		for _, f := range formats {
-			err := ex.p.Save(4*vg.Inch, 4*vg.Inch, ex.name+f)
-			if err != nil {
-				log.Fatalf("failed to save %s%s: %v", ex.name, f, err)
+	//for _, ex := range examples {
```
kortschak: Is this kept here for a reason?

vg/draw/canvas.go
```diff
@@ -279,14 +317,16 @@ func (c *Canvas) Y(y float64) vg.Length {
 // Crop returns a new Canvas corresponding to the receiver
 // area with the given number of inches added to the minimum
 // and maximum x and y values of the Canvas's Rectangle.
-func (c Canvas) Crop(minx, miny, maxx, maxy vg.Length) Canvas {
+// Note that cropping the right and top sides of the canvas
+// requires specifying negative values of xmaxcrop and ymaxcrop.
+func (c Canvas) Crop(xmincrop, ymincrop, xmaxcrop, ymaxcrop vg.Length) Canvas {
```
kortschak: I don't think this stutter adds to the clarity.

vg/draw/canvas.go
```diff
@@ -294,6 +334,50 @@ func (c Canvas) Crop(minx, miny, maxx, maxy vg.Length) Canvas {
...
+}
+
+// At returns the canvas at the given row and column.
+func (t Tiles) At(row, col int) Canvas {
+	return t.Values[row*t.Cols+col]
+}
+
+// Tile returns new Canvases corresponding to the reciever canvas tiled
+// into the given number of rows and columns (row major, starting from the
+// top left corner) where t, b, l, r, h,
```
kortschak: Fix parameter names here.

vg/draw/canvas.go
```diff
@@ -279,14 +317,16 @@ func (c *Canvas) Y(y float64) vg.Length {
...
+		Y: c.Min.Y + ymincrop,
 	}
 	maxpt := Point{
-		X: c.Max.X + maxx,
-		Y: c.Max.Y + maxy,
+		X: c.Max.X + xmaxcrop,
+		Y: c.Max.Y + ymaxcrop,
 	}
 	return Canvas{
 		Canvas:    vg.Canvas(c),
```
kortschak: While you are here - delete the redundant vg.Canvas conversion.

kortschak:
Please change to Tiles, the verb in any use of the method is the assignment operator.

As an aside is there a considered reason for these functions (Crop and Tile[s]) to be methods on draw.Canvas? It seems to me that they could be package-level funcs. Compare:

`cc := draw.Crop(c, minx, miny, maxx, maxy)`

`cc := c.Crop(minx, miny, maxx, maxy)`

and

`t := draw.TilesIn(c, rows, cols, top, bottom, left, right, padx, pady)`

`t := c.Tiles(rows, cols, top, bottom, left, right, padx, pady)`

The main advantage this has for Crop is that there is no confusion about whether the call alters the receiver (the docs say this, but it's better to have the code speak for itself where possible). The advantage in Tile[sIn] is less strong, but I think it reads better.

vdobler:
Hello everybody

I skimmed the discussion and have some remarks regarding Tiles:
This tiling is basically what gnuplot does if the `layout` option is used
in the multiplot command. Regular tiling of plots on a canvas is dead
common so the multiplot-mode of gnuplot deserved an option to do
this automatically. So plot should support this too.

BUT: The main difference between gnuplot's solution and the current
pull request with Tiles is that gnuplot's `layout` option is a convenience
function to a _general_ multiplot mode which does allow irregular
tilings and subplots. (Generating regular tiling with the general multiplot
mode is possible but requires manual and error prone work.)

What gnuplot's layout and the proposed Tiles have in common is that
both solution draw _completely_ _unrelated_ plots on a single page/canvas:
Neither axis ranges nor plot bordes (needed to draw the axis) are 
shared. Common axis can be done in gnuplot and plot by manually
setting them; common plot borders are not possible in plot (always determined
automatically to actual label size) and have to be specified manually in
gnuplot.

The faceted plot I will continue to work on is the other extreme: Often
all facets will share the same x and y range, all borders must be the
same for proper alignment of the actual graphs, facet row/column
labels have to be drawn and inner axis are typically omitted.

I think package plot should get some form of real multiplot support (like
gnuplot or R has) and convenience tiling via Tiles should be a helper
function/method/tool based on this generic "multiple plots on one canvas"
solution. Maybe this generic solution could handle shared axis in an
automatic (and sensible) way. Then plots like the Cern example Sebastien
showed would be possible without hassle.

ctessum:
The draw package doesn't know anything about axis ranges, so I think it might make sense to have function(s) in the draw package that just make sub-canvases, and then to have function(s) that build on top of that in the plot and/or plotter packages that handle setting common axis properties and labels. In terms of  general multiplot support, I think Crop already does that. Crop allows setting arbitrarily sized sub-canvases, and Tile handles the specific (and common) case where you want the sub-canvases to be regularly sized and spaced. In line with the Go "[do less, enable more](http://blog.golang.org/open-source#TOC_3.)" philosophy, I think these two functions enable a lot, and make the common use case easy, within a small footprint.

vg/draw/canvas.go
ctessum: In this context, the word crop means to "make smaller" or "remove material". In this function, if I want to remove 2 cm from the right of the canvas, which at least I would think of as cropping +2 cm, I have to enter -2 cm as the argument for maxx, which is counterintuitive, at least to me. Although the documentation as is is technically correct, I had to originally read it multiple times before figuring out why I wasn't getting the expected results.

kortschak: Sorry, I was referring to the change to parameter names; I don't think suffixing them all with crop adds anything.

kortschak: How about left, right, top and bottom?

ctessum:
PTAL. I didn't changes the Crop and Tiles methods to package functions because then there would be both a type and a function called "Tiles", which is not allowed. I could change it to TilesIn if that's desired, but it is more keystrokes than Tiles.

kortschak:
Why not change Crop? There is no namespace conflict there.

TilesIn is good for me, but wait for @eaburns.

plotter/main.go
```diff
@@ -49,22 +50,53 @@ var examples = []struct {
...
+	for _, f := range formats {
+		c, err := draw.NewFormattedCanvas(10*ncols*vg.Centimeter,
+			10*nrows*vg.Centimeter, f)
+		if err != nil {
+			log.Fatal(err)
+		}
+		dc := draw.New(c)
+		const (
+			p     = 1 * vg.Centimeter
+			nrows = 4
```
eaburns: Why are these defined again here. They are already defined above.

vg/draw/canvas.go
```diff
@@ -279,21 +317,65 @@ func (c *Canvas) Y(y float64) vg.Length {
 // Crop returns a new Canvas corresponding to the receiver
 // area with the given number of inches added to the minimum
```
eaburns: Can you re-word this? It's no longer using inches.

vg/draw/canvas.go
```diff
@@ -279,21 +317,65 @@ func (c *Canvas) Y(y float64) vg.Length {
...
+func (t Tiles) At(row, col int) Canvas {
+	return t.Values[row*t.Cols+col]
+}
+
+// Tiles returns new Canvases corresponding to the reciever canvas tiled
+// into the given number of rows and columns (row major, starting from the
+// top left corner) where top, bottom, left, right, padx,
+// and pady are the padding above, below, to the left, to the right, between columns,
+// and between rows of the tiles, respectively.
+func (c Canvas) Tiles(rows, cols int, top, bottom, left, right, padx, pady vg.Length) *Tiles {
```
eaburns: There are a lot of parameters here. I'd like to avoid that. What if we restructured it as follows:
```
// Tiles specifies a tiling that can be applied to a Canvas.
// ... whatever else ...
type Tiles struct {
    // Cols and Raws specify the number of rows and columns of tiles.
    Cols, Rows int
    // PadTop, PadBottom, PadRight, and PadLeft specify the padding on the corresponding side of each tile.
    PadTop, PadBottom, PadRight, PadLeft vg.Length
    // PadX and PadY specify the padding between columns and rows of tiles respectively..
    PadX, PadY vg.Length
}

// At returns the subcanvas within c that corresponds to the tile at column x, row y.
type (ts Tiles) At(c Canvas, x, y int) Canvas { ... }
```
So, instead of:
```c.Tiles(5, 6, 0, 0, 0, 0, 1*vg.Millimeter, 1*vg.Milimeter)```
One would do:
```t := Tiles{ Rows: 5, Cols: 6, PadX: 1*vg.Millimeter, PadY: 1*vg.Millimeter }```
This is both more clear because the parameters are named, and it lets you take advantage of the zero-initialization.

Finally, it solves the issue of wanting both a type and a function named `Tiles`. Instead, you just have the type.

**Comment by eaburns** [comment]
I like this. However, as @vdobler points out, it won't align plot axes. We should probably open an issue to add support for that. I think that it could work as functionality in the plot package which makes use of this canvas tiling to actually split the canvas.

One simple way to do it would be to apply the maximum X-axis padding in a column to all plots in that column, and to apply the maximum Y-axis padding in a row to all plots in that row. Do it for each column and each row. This would align the X axes of plots in the same column and the Y axes of plots in the same row. (Separately from this, we can think about how to extend the axes into the padding. I think that would cover all of @vdobler's concerns.)

**Comment by kortschak** [comment]
It was my understanding that @vdobler's experiments would, at some point, be merged. Won't that cover the shared axis functionality - this set of changes would form a basis for his eventual changes here.

**Comment by eaburns** [comment]
OK

**Comment by kortschak** [comment]
*ping* @ctessum

**Comment by vdobler** [comment]
The suggestion of Ethan is basically how gnuplot works. I'm fine with that.

**Comment by ctessum** [comment]
I liked the suggestions. Sorry for the delay. PTAL.

File: vg/draw/canvas.go [hunk 14]
```diff
@@ -276,24 +314,59 @@ func (c *Canvas) Y(y float64) vg.Length {
...
 
-// Crop returns a new Canvas corresponding to the receiver
-// area with the given number of inches added to the minimum
+// Crop returns a new Canvas corresponding to the Canvas
+// c with the given lengths added to the minimum
 // and maximum x and y values of the Canvas's Rectangle.
-func (c Canvas) Crop(minx, miny, maxx, maxy vg.Length) Canvas {
+// Note that cropping the right and top sides of the canvas
+// requires specifying negative values of right and top.
+func Crop(c Canvas, right, bottom, left, top vg.Length) Canvas {
```
sbinet: I like the new `right`, `bottom`, `left` and `top` names.
that's really, *really* nitpicking, but: is it the right order though?
I don't have a preference and, in that department, consistency is king.

I suppose you chose a clockwise order.
wouldn't `left, right, bottom, top` make more sense to left-right reading people? (sorry `<right-to-left languages>`)

(feel free to ignore me, though)

vg/draw/canvas.go
```diff
@@ -276,24 +314,59 @@ func (c *Canvas) Y(y float64) vg.Length {
...
-		Y: c.Max.Y + maxy,
+		X: c.Max.X + left,
+		Y: c.Max.Y + top,
 	}
 	return Canvas{
-		Canvas:    vg.Canvas(c),
+		Canvas:    c,
 		Rectangle: Rectangle{Min: minpt, Max: maxpt},
 	}
 }
```
eaburns: Add a comment on the type.

vg/draw/canvas.go
```diff
@@ -276,24 +314,59 @@ func (c *Canvas) Y(y float64) vg.Length {
...
 
+type Tiles struct {
+	// Cols and Raws specify the number of rows and columns of tiles.
```
eaburns: Rows

ctessum:
Anything else?

vg/draw/canvas.go
ctessum: Do you mean the "Tiles" type? If so, I added one, but suggestions on the wording are welcome.

kortschak:
Rebase onto master. There should be one failing test.

vg/draw/canvas.go
eaburns: Yeah, that's what I meant. Your comment seems fine.

ctessum:
Rebased. Should I merge it?

eaburns:
I just skimmed very quickly, but it seemed good. I'm not sure what the story is with the expected failing test, so maybe wait for Dan.

kortschak:
LGTM

The failing test is just the vgx11 package because upstream has not moved to the new freetype package.

eaburns:
Ah. Right. Thanks. LGTM.

PR #216, opened 2015-08-27
Update freetype imports

jle:
Looks like code.google.com/p/freetype-go redirects to github.com/golang/freetype now. This fixes `go get` for this project, but some dependent projects like draw2d may still refer to the old path.

PR #219, opened 2015-09-10
fixd #216, Update freetype imports

l2x:


kortschak:
There is a pending PR, #216. We are waiting on upstream fixes.

PR #222, opened 2015-09-29
vg: drop DPI from Canvas

kortschak:
The DPI value is only ever used internally by types to determine the number of dots for a given Length, the alternative is to give the Dots method the DPI value rather than the Canvas.

The DPI method is retained on types that need to know their resolution, although this could be made unexported.

@eaburns This is exploratory, so I'm not fussed if you reject it, but it does drop one method from the vg.Canvas set and it doesn't really look like it impacts in a negative way to my mind. Please take a look.

### DIFF
--- a/plot_test.go
+++ b/plot_test.go
@@ -35,8 +35,8 @@ func TestLegendAlignment(t *testing.T) {
 		l.Add(n, b)
 	}
 
-	r := recorder.New(100)
-	c := draw.NewCanvas(r, 100, 100)
+	var r recorder.Canvas
+	c := draw.NewCanvas(&r, 100, 100)
 	l.Draw(draw.Canvas{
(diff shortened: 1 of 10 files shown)


eaburns:
Sorry, in very busy this week, but I hope to look at this this weekend. Reading the description, I like the sound of it.

kortschak:
There is room for removing more of the DPI methods - those canvases that merely return a package constant don't need it.

eaburns:
LGTM. I like it. Thanks.

kortschak:
Are you happy for me to drop the DPI method from all types except the img-based types? I can leave a package const DPI for information.

kortschak:
PTAL

eaburns:
Yes. LGTM.

PR #226, opened 2015-10-02
add XYLabels

ctessum:


plotter/labels.go
```diff
@@ -115,3 +112,20 @@ func (l *Labels) GlyphBoxes(p *plot.Plot) []plot.GlyphBox {
 	}
 	return bs
 }
+
+// XYLabeller combines the XYer and Labeller types
```
eaburns: Add a period at the end.

plotter/labels.go
```diff
@@ -115,3 +112,20 @@ func (l *Labels) GlyphBoxes(p *plot.Plot) []plot.GlyphBox {
...
+type XYLabeller interface {
+	XYer
+	Labeller
+}
+
+// XYLabels holds XY data with labels.
```
eaburns: Can you add a note on how the labels match up with the XYs?
Maybe something like, "The ith label corresponds to the ith XY."?

eaburns:
This LGTM. I don't remember why I didn't do it in the first place, but I remember consciously avoiding it. I think that I was just trying to cut down on the number of interfaces. But this seems fine to me.

ctessum:
Looks like I mis-merged again. Sorry about that.

Issue #78, opened 2015-03-15
Add Rotation to the text style.


eaburns:
Original [issue 51](https://code.google.com/p/plotinum/issues/detail?id=51) created by eaburns on 2012-08-08T18:18:01.000Z:

Both plotter.Labels and tick marks should be able to draw with different rotations, so it might as well go into the text style.

aboinameddave:
Any news on being able to rotate the axis label orientation? This would be useful in combination for time/date labels on the x-axis because these labels are long.

eaburns:
Was this fixed by #227?

ctessum:
To rotate tick marks, it is for example plot.Plot.X.Tick.TextStyle.Rotation

To rotate the axis label, it is for example plot.Plot.X.Label.TextStyle.Rotation

There is an example [here](https://godoc.org/github.com/gonum/plot/plotter#example-package--Rotation); the output given by the example is [here](https://github.com/gonum/plot/blob/master/plotter/testdata/rotation_golden.png).

ctessum:
This issue can probably be closed now, right?

eaburns:
Yes, we can close it.

PR #227, opened 2015-10-02
Add rotation to draw.TextStyle

ctessum:


vg/draw/canvas.go
```diff
@@ -32,6 +32,9 @@ type TextStyle struct {
 
 	// Font is the font description.
 	Font vg.Font
+
+	// Rotation is the text rotation in radians.
```
kortschak: State which direction 0 is and which orientation the rotation is - both vary depending on context.

axis.go
```diff
@@ -280,12 +280,10 @@ func (a *verticalAxis) size() (w vg.Length) {
 // draw draws the axis along the left side of a draw.Canvas.
 func (a *verticalAxis) draw(c draw.Canvas) {
 	x := c.Min.X
+	a.Label.TextStyle.Rotation = math.Pi / 2.
```
kortschak: +=?

vg/draw/canvas.go
```diff
@@ -587,14 +590,24 @@ func (c *Canvas) FillText(sty TextStyle, x, y vg.Length, xalign, yalign float64,
...
 	c.SetColor(sty.Color)
 
-	ht := sty.Height(txt)
-	y += ht*vg.Length(yalign) - sty.Font.Extents().Ascent
+	c.Push()
+	c.Rotate(sty.Rotation)
+
+	cos := vg.Length(math.Cos(sty.Rotation))
+	sin := vg.Length(math.Sin(sty.Rotation))
+	yprime := y*cos - x*sin
```
kortschak: x, y = y*sin + x*cos,  y*cos - x*sin

axis.go
```diff
@@ -280,12 +280,10 @@ func (a *verticalAxis) size() (w vg.Length) {
 // draw draws the axis along the left side of a draw.Canvas.
 func (a *verticalAxis) draw(c draw.Canvas) {
 	x := c.Min.X
+	a.Label.TextStyle.Rotation += math.Pi / 2.
```
eaburns: It bothers me that calling `draw` has the side affect of Incrementing `Rotation`. Won't this cause a second call to draw to have an incorrect rotation? Can we do this without changing the fields of the axis? At very least, we should have a deferred call that restores the original rotation.

eaburns:
I really want something like this to make it in. I tried it *at least* once before (see #78, which, by the way, should be marked as fixed if this goes in), and failed, so I'm nervous about this without some more testing. Specifically, what I'd like to see is a plot with a `NominalX` axis (see `Example_boxPlots`) where the axis labels are rotated at π/4. Rotating the labels is very useful when trying to jam lots of box lots or bars into one plot (consider the case where each label is a date). See an example here: http://matplotlib.org/examples/api/date_index_formatter.html.

axis.go
kortschak: Yeah, sorry, that was my fault. What about just not modifying the angle at all?

kortschak:
I've had a lot of success with text rotation using essentially this technique in the biogo/graphics/rings package.

axis.go
eaburns: Agreed. We can change line 286 to:
```
sty := a.Label.TextStyle
sty.Rotation += math.Pi/2
c.FillText(sty, x, c.Center().Y, -0.5, 0, a.Label.Text)
```
Otherwise, everyone would need to manually set the rotation of the label text each time, which would be annoying.

kortschak:
@ctessum Would you make the push/rotate/adjust/pop code conditional on rotation being non-zero.

kortschak:
Further suggestion after looking at how I did this in rings; when you rotate, what you rotate around is important (there is a linear mapping, so it's not crucial, but it is very helpful to have the ability to say rotate around and end or the middle etc.). Anyway, have a look at https://github.com/biogo/graphics/blob/master/rings/axis.go#L167. The situation there is a little more complicated because TextPlacement must be called since angles change throughout the plot, so the x and y adjustment may be dynamic.

axis.go
kortschak: That requires that there be a note on the axis documentation that the origin angle is at right angles to everything else, but OK.

kortschak:
The bounding box needs to be adjusted for rotation. If a non-zero rotation is used, the text goes off the edge of the canvas.

eaburns:
WRT what you rotate around: that is an interesting point. When rotating a tick label, you'd like an end of the label to line up with its tick, so you may want to rotate around an end, not the center.

As for glyphboxes: it's not a perfect fit. The glyphbox padding code will push the axes around in order to make room in the draw area, but the labels are on the opposite side of the axes from the draw area. Instead, see `tickLabelWidth`: https://github.com/gonum/plot/blob/master/axis.go#L487. That code should account for the rotation.

kortschak:
>When rotating a tick label, you'd like an end of the label to line up with its tick, so you may want to rotate around an end, not the center.

The operative word being may. I can see cases where the centre is what is wanted. Hence having a rotation centre point that can be adjusted.

>As for glyphboxes: it's not a perfect fit.

It would still be nice to be able to draw them - certainly to be able to draw glyph boxes around labels generally. What is the reason the GlyphBox method exists on that type?

eaburns:
The Labels type in plotter is for drawing labels on data points *within* the draw area. They are a plotter concept, not inherent to plots. They are just another Plotter---another way to draw the data. For better or worse, this is different from tick labels. Tick labels are integrated directly into the plot package. They are a fundamental piece of axes. I.e., they don't use the Plotter generalization, but instead their drawing is hard-coded into axis.go.

eaburns:
A movable rotation point sounds good to me. As long as the zero-value is whatever we think the most common rotation will be, so that it rarely needs to change.

ctessum:
I may be missing something, but I think you can already specify the center point of rotation by using the x and y offset arguments to the FillText function. (0,0) would cause it to rotate around the upper left corner of the label, (-0.5,-0.5) would cause it to rotate around the center of the label, etc. Is what you're suggesting something in addition to that?

kortschak:
You can if you have direct access to the FillString method, but not if it is called for you, as is the case for axes. I guess the question is how to thread that information through. Do you add fields in TextStyle to pass to FillString for adjustment, or do you add fields to Axis?

ctessum:
That sounds like an issue that's independent of adding rotation abilities, but we can definitely fix it here. Another aspect of what I think is the same or similar issue is that in plotter.Labels, all the labels currently have to have the same alignment, but there are cases where you would want to give different labels different alignments. 

I think I would vote for adding it to TextStyle and keeping the default as (0,0).

eaburns:
Adding the offset to TextStyle sounds good to me. The zero value works for most cases, so it can just be ignored by all but power-users.

sbinet:
ping ?

eaburns:
I didn't read over this in detail, but from my last comment I assumed that there was more to do here. Any update on the status?

ctessum:
Sorry for the delay. I think I've made all the requested changes, but I seem to have made a mess with git. I will try to fix it.

ctessum:
So I don't know how I messed up this badly, but I can't figure out how to fix it so I've started over in PR#249. Sorry about this.

kortschak:
For future reference if you know the last good commit ref, you can `git reset --hard $REF && push -f origin $BRANCH` and then try again.

ctessum:
I am at least aware of that. I think the problem happened when I tried to rebase it onto the master branch. Now there are somehow 24 (mostly unrelated) commits instead of 3.

kortschak:
I think you could have hard reset to https://github.com/gonum/plot/commit/2b889eae5aff241ba944244e009dcfc339fe287b and you would have saved the situation.

ctessum:
Thanks. I think that did it. Let me know how it looks. I'll delete the other one.

kortschak:
Don't delete the other PR yet, I'll use it to compare.

kortschak:
They are identical according to git, so you can close #249 and delete its branch.

I'll find some time soon to go over this, but also wait for @eaburns.

axis.go
```diff
@@ -104,12 +105,25 @@ func makeAxis() (Axis, error) {
...
-	}
-	a.Tick.Label = draw.TextStyle{
-		Color: color.Black,
-		Font:  tickFont,
+		Color:  color.Black,
+		Font:   labelFont,
+		XAlign: -0.5,
+		YAlign: 0,
+	}
+	if vertical {
```
eaburns: I'd probably do it like this, so that only the pieces which are changing conditionally are changed by the conditional.
```
var xalign, yalign = -0.5, 0
if vertical {
  xalign, yalign = -1, -0.5
}
a.Tick.Label = draw.TextStyle{…, XAlign: xalign, YAlign: yalign }
```

legend.go
```diff
@@ -83,12 +83,12 @@ func makeLegend() (Legend, error) {
...
 	iconx := c.Min.X
-	textx := iconx + l.ThumbnailWidth + l.TextStyle.Width(" ")
-	xalign := 0.0
+	sty := l.TextStyle
+	textx := iconx + l.ThumbnailWidth + sty.Width(" ")
 	if !l.Left {
 		iconx = c.Max.X - l.ThumbnailWidth
 		textx = iconx - l.TextStyle.Width(" ")
-		xalign = -1
+		sty.XAlign -= 1.
```
eaburns: Why did this change from `= -` to `-=`?

Also, why the stray dot in `1.`? I'd prefer `1.0` to the stay dot, but I'm perfectly happy with just plane old `1`.

plot.go
```diff
@@ -87,11 +87,11 @@ func New() (*Plot, error) {
 	if err != nil {
 		return nil, err
 	}
-	x, err := makeAxis()
+	x, err := makeAxis(false)
```
eaburns: Personally, I don't like passing a bool here. At the callsite, it's impossible to tell what the bool means. I'd prefer a const:
```
type axisLayout bool
const (
    verticalAxis axisLayout = true
    horizontalAxis axisLayout = false
}
```
or something.

Another option is to leave the XAlign and YAlign as their zero value and set them here at the callsite, since that is the first place where we know how to align the text.

plot.go
```diff
@@ -106,8 +106,10 @@ func New() (*Plot, error) {
 		Legend:          legend,
 	}
 	p.Title.TextStyle = draw.TextStyle{
-		Color: color.Black,
-		Font:  titleFont,
+		Color:  color.Black,
+		Font:   titleFont,
+		XAlign: -0.5,
```
eaburns: I was always pretty happy with this XAlign and YAlign design for aligning text, but reading it years later, it's pretty opaque. Maybe we could make some helpful, and more-readable functions to set it:

```
func CenteredAbove(sty *TextStyle) {
    sty.XAlign = -0.5
    sty.YAlign = -1
}
// etc.
```

eaburns: It also occurs to me that if we make such a function, we can pass _that_ as the argument to makeAxis:

`func makeAxis(tickAlign func(*TextStyle)) {…}`

plot_test.go
```diff
@@ -211,3 +212,23 @@ func formatActions(actions []recorder.Action) string {
 	}
 	return buf.String()
 }
+
+// printActionDiff prints the first line that is different between two actions.
+func printActionDiff(got, want []recorder.Action) string {
```
eaburns: printFirstDiff

vg/draw/canvas.go
```diff
@@ -32,6 +32,23 @@ type TextStyle struct {
...
 	Font vg.Font
+
+	// Rotation is the text rotation in radians, counterclockwise from
+	// the default drawing direction.
+	Rotation float64
+
+	// XAlign and YAlign specify the alignment of the text.
+	// They are multiplied by the width
+	// and height of each label respectively and the
+	// added to the final location.  E.g., XAlign=-0.5
```
eaburns: I prefer your `(XAlign, YAlign) = (-0.5, -0.5)`. But I also think that this can be improved by changing it to a table of useful examples:
```
// For example:
// 	XAlign	YAlign	text location
// 	-0.5		-0.5		vertically and horizontally centered
// 	-0.5		0		below and horizontally centered
// 	-1		-0.5		vertically centered and right aligned
```
(Please double check that these are right).

Also, if we add functions to set these, maybe mention those functions as a convenient way to set some useful alignments.

eaburns:
I added some small comments. Generally, it looks very good. One big thing that is missing in my opinion is that there is nothing testing rotated text other than the axis labels. I'd _love_ to see rotated tick labels like in the plot that I link above. It would also be nice to see a plot with plotter.Labels rotated to various interesting angles around various interesting center points (the label text can describe the rotation and the center point, making the test image easy to understand).

legend.go
ctessum: If it's `=-`, then any alignment that was already there gets overwritten. I changed it to `-=` so that any existing alignment is adjusted rather than ignored.

The linter suggested `sty.XAlign--`, so I changed it to that.

plot.go
ctessum: I'm changing it to 

``` go
type orientation bool
const (
    vertical orientation = true
    horizontal orientation = false
}
```
because some of the suggested names were already in use

plot.go
ctessum: I like the idea of making it more obvious what these things do, but the functions also seemed a little difficult to wrap one's head around so I'm trying out a constant-based approach. Let me know what you think.

vg/draw/canvas.go
ctessum: As mentioned above, I changed this to use constants. Let me know what you think.

ctessum:
PTAL.

ctessum:
I'm not sure why two of the golden images are different and the Y tick labels on the image I added aren't aligned quite correctly. Suggestions welcome.

ctessum:
Ping?

sbinet:
according to github, this branch has conflicts.
could you resolve them and update?
(thanks)

ctessum:
PTAL. I also fixed some of the go-vet errors that have been bothering me for a while. Ideally, one would create a separate pull request for fixing the linter and go-vet errors but it seems unlikely that anyone will actually do that.

You may notice that many of the golden images are changed. Most of them do not have any visible changes, and I think at least in some cases the differences are just floating point errors having to do with a change in the method of calculating text rectangles. A couple of them have different random data because I changed rand.Seed from 1 to 0. There are also a couple of images that are different because the way that the text and data interacts with the very right edge of the canvas has slightly changed, but I don't understand exactly how or why that behavior has changed.

kortschak:
Why was the rand seed changed to 0? The default seed for the the rand package is 1.

axis.go
```diff
@@ -43,7 +43,8 @@ type Axis struct {
 		// Text is the axis label string.
 		Text string
 
-		// TextStyle is the style of the axis label text.
+		// TextStyle is the style of the axis label text. For the vertical axis,
```
kortschak: Can you shorten the line lengths here. Doc comments in structs and interfaces don't get flowed by godoc, so we need to be aware of that. Generally, shorter line lengths are easier to read in this context.

Also, I think changing the text to "For the vertical axis, one quarter turn counter clockwise will be added to the label text before drawing."

internal/cmpimg/cmpimg.go
```diff
@@ -2,7 +2,7 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-// package cmpimg compares the raw representation of images taking into account idiosyncracies related to their underlying format (SVG, PDF, PNG, ...)
+// Package cmpimg compares the raw representation of images taking into account idiosyncracies related to their underlying format (SVG, PDF, PNG, ...)
```
kortschak: Break this line while you are here.

plot.go
```diff
@@ -80,18 +80,25 @@ type DataRanger interface {
 	DataRange() (xmin, xmax, ymin, ymax float64)
 }
 
+type orientation bool
```
kortschak: This is unexported, undocumented and has no methods. Is it needed? The consts can be untyped boolean.

plotter/general_test.go
```diff
@@ -30,6 +30,7 @@ func goldenPath(path string) string {
 // checkPlot checks a generated plot against a previously created reference.
 // If generateTestData = true, it regereates the reference.
 func checkPlot(ExampleFunc func(), t *testing.T, filenames ...string) {
+	rand.Seed(0)
```
kortschak: To make this concurrency-safe, use a rand.Rand instead of the package-level functions.

plotter/rotation_test.go
```diff
@@ -0,0 +1,106 @@
...
+	sin := func(n int, xmax float64) XYs {
+		xy := make(XYs, n)
+		for i := 0; i < n; i++ {
+			xy[i].X = xmax / float64(n) * float64(i)
+			xy[i].Y = math.Sin(xy[i].X) * 100
+		}
+		return xy
+	}
+
+	// These points will make up our sin curve.
```
kortschak: s/sin/sine/ in comments.

vg/draw/canvas.go
```diff
@@ -32,8 +32,47 @@ type TextStyle struct {
 
 	// Font is the font description.
 	Font vg.Font
+
+	// Rotation is the text rotation in radians, counterclockwise from
```
kortschak: Shorten the lines on these doc comments.

vg/draw/canvas.go
```diff
@@ -32,8 +32,47 @@ type TextStyle struct {
...
+	// the default drawing direction.
+	Rotation float64
+
+	// XAlign and YAlign specify the alignment of the text.
+	// Alignment and rotation are interrelated:
```
kortschak: This seems to complicated for an in-type comment.

vg/draw/canvas.go
```diff
@@ -32,8 +32,47 @@ type TextStyle struct {
...
+	// when (XAlign, YAlign) = (XCenter,YCenter), the text is rotated about its center,
+	// (XLeft,YBottom) causes the text to rotate around its bottom left corner, etc.
+	XAlign XAlignment
+	YAlign YAlignment
 }
 
+// XAlignment specifies text alignment in the X direction. Three preset
+// options (XLeft, XCenter, and XRight) are available, but an arbitrary alignment
```
kortschak: This parenthetic statement is not necessary, they will see the default values in the godoc.

vg/draw/canvas.go
```diff
@@ -32,8 +32,47 @@ type TextStyle struct {
...
+	// XLeft aligns the left edge of the text with the specified location.
+	XLeft XAlignment = 0
+	// XCenter aligns the horizontal center of the text with the specified location.
+	XCenter XAlignment = -0.5
+	// XRight aligns the right edge of the text with the specified location.
+	XRight XAlignment = -1
+)
+
+// YAlignment specifies text alignment in the Y direction. Three preset
+// options (YLeft, YCenter, and YRight) are available, but an arbitrary alignment
```
kortschak: Same here.

vg/draw/canvas.go
```diff
@@ -576,29 +615,54 @@ func isect(p0, p1, clip, norm vg.Point) vg.Point {
 }
 
 // FillText fills lines of text in the draw area.
-// The text is offset by its width times xalign and
-// its height times yalign.  x and y give the bottom
-// left corner of the text befor e it is offset.
-func (c *Canvas) FillText(sty TextStyle, pt vg.Point, xalign, yalign float64, txt string) {
+// x and y specify the location where the text is to be drawn.
```
kortschak: x and y? `pt specifies the location where the text is to be drawn.`

vg/draw/canvas.go
```diff
@@ -576,29 +615,54 @@ func isect(p0, p1, clip, norm vg.Point) vg.Point {
...
 		c.FillString(sty.Font, pt.Add(vg.Point{X: xoffs, Y: n * sty.Font.Size}), line)
 	}
+
+	if sty.Rotation != 0 {
+		c.Pop()
+	}
+}
+
+// rotatePoint applies rotation Θ (in radians) about the origin to point p.
+func rotatePoint(Θ float64, p vg.Point) vg.Point {
```
kortschak: Use the name of the letter rather than the greek. This is a bug waiting to happen.

vg/draw/canvas.go
```diff
@@ -621,9 +685,51 @@ func (sty TextStyle) Height(txt string) vg.Length {
 }
 
 // Rectangle returns a rectangle giving the bounds of
-// this text assuming that it is drawn at 0, 0
+// this text assuming that it is drawn at 0, 0.
```
kortschak: `at (0, 0)`

vg/draw/canvas.go
ctessum: It depends on the language being written in, right?

vg/draw/canvas.go
ctessum: I think the interrelated part was requested by @eaburns . Suggestions?

vg/draw/canvas.go
ctessum: You mean an έντομο waiting to happen?

vg/draw/canvas.go
kortschak: Does it need to be said? The directions handled are left or right and a quarter turn rotation ccw results in the same relative transform for both.

vg/draw/canvas.go
kortschak: It seems to me that it fits better in a package level comment.

ctessum:
PTAL. Now it looks like the only golden image that is visibly changed is plotter/testdata/verticalBoxPlot_golden.png

vg/draw/canvas.go
kortschak: Indeed.

gob/gob_test.go
```diff
@@ -25,12 +25,14 @@ func init() {
 }
 
 func TestPersistency(t *testing.T) {
+
```
kortschak: Personal preference, but I dislike leading blank lines in functions. Up to you.

gob/gob_test.go
```diff
@@ -25,12 +25,14 @@ func init() {
...
+	random := rand.New(rand.NewSource(1))
```
kortschak: Use `rnd` instead of `random`.

vg/draw/canvas.go
```diff
@@ -33,20 +33,24 @@ type TextStyle struct {
 	// Font is the font description.
 	Font vg.Font
 
-	// Rotation is the text rotation in radians, counterclockwise from
-	// the default drawing direction.
+	// Rotation is the text rotation in radians,
+	// counterclockwise from the default drawing
+	// direction. (In English, the default drawing
```
kortschak: I don't like this here.

Discussion of this also feels like a package-level comment.

kortschak:
LGTM with some reservations about documentation. Obviously please wait for @eaburns.

vg/draw/canvas.go
ctessum: Does package-level mean as a comment on the struct rather than the field, or does it mean adding it as part of the overall package summary comment?

ctessum:
PTAL

gob/gob_test.go
```diff
@@ -25,12 +25,13 @@ func init() {
 }
 
 func TestPersistency(t *testing.T) {
+	rnd := rand.New(rand.NewSource(1))
```
eaburns: It seems like this change could go into a separate PR.

plot_test.go
```diff
@@ -208,3 +209,23 @@ func formatActions(actions []recorder.Action) string {
 	}
 	return buf.String()
 }
+
+// printFirstDiff prints the first line that is different between two actions.
```
eaburns: I think this could also go in a separate PR.

eaburns:
Thanks for this. The new sine plot looks very nice!
I skimmed the code, and besides @kortschak's one open remark about commenting the text rotation LGTM.

By the way, a lot of code changed here. For next time, I would recommend breaking this change up into 3 different PRs:
1. test changes with respect to random seeds
2. adding X: Y: labels to all of the points
3. adding text rotation.
That would have made it much easier to review.

ctessum:
Thanks. Regarding to the open comment, I think I responded to it a while ago, and then a contradictory comment was made more recently, and when I responded to the more recent one it caused the original one to open back up. @kortschak let me know if you are satisfied with the current documentation.

plotter/general_test.go
```diff
@@ -81,6 +79,7 @@ func checkPlot(ExampleFunc func(), t *testing.T, filenames ...string) {
 
 // Draw the plot logo.
 func Example() {
+
```
kortschak: Why do we have these?

vg/draw/canvas.go
```diff
@@ -26,14 +26,55 @@ type Canvas struct {
 }
 
 // TextStyle describes what text will look like.
+// Alignment and rotation are interrelated: when
```
kortschak: I don't think having this text here really improves the readability of this. The use of examples in type doc comments is not something that we see in most packages.

vg/draw/canvas.go
```diff
@@ -26,14 +26,55 @@ type Canvas struct {
...
 type TextStyle struct {
 	// Color is the text color.
 	Color color.Color
 
 	// Font is the font description.
 	Font vg.Font
+
+	// Rotation is the text rotation in radians.
+	// Positive rotation is counterclockwise.
+	Rotation float64
```
kortschak: Apologies for the reshift, but the text was troublesome. Here it could be (not perfect but clearer):

```
// Text rotation is performed around the axis
// defined by XAlign and YAlign.
```

axis.go
```diff
@@ -44,6 +44,9 @@ type Axis struct {
 		Text string
 
 		// TextStyle is the style of the axis label text.
+		// For the vertical axis, one quarter turn
+		// counter clockwise will be added to the label
```
kortschak: AFAIUI ASE has "counterclockwise" as one word.

eaburns:
The comment that I was referring to was about stating what a rotation of 0 means. It doesn't appear to be addressed. Presumably it means drawing the text horizontally left-to-right, but I agree that we might as well say so.

After that and Dan's few other remarks, LGTM.

plotter/general_test.go
ctessum: I'm assuming "these" refers to empty lines at the beginning of functions? I will delete them.

ctessum:
PTAL. My understanding is that @kortschak has undergone an evolution in his position regarding the documentation for the text rotation since that first comment. His most recent suggestion is 

``` go
// Text rotation is performed around the axis
// defined by XAlign and YAlign.
```

In the last commit I have amended that so that it follows the convention of starting with the variable name:

``` go
	// Rotation is the text rotation in radians, performed around the axis
	// defined by XAlign and YAlign.
```

It may not be completely necessary to state that it's in radians, but at least in my case I tend to think about things in degrees moreso so to the extent that is also true for other users it may be useful to have that as a reminder.

Let me know if you see any further changes.

PR #229, opened 2015-10-08
Added horizontal BarChart

ctessum:
...and corresponding tests. I've formatted the tests as examples.


kortschak:
Please make the tests and the examples distinct. It's good to have examples, and the examples should be correct - hence running them, but the examples should not really be the tests (I'm guilty of having done this in the past and I feel shame).

axis.go
```diff
@@ -327,8 +331,9 @@ func (a *verticalAxis) GlyphBoxes(*Plot) (boxes []GlyphBox) {
 		}
 		h := a.Tick.Label.Height(t.Label)
 		box := GlyphBox{
-			Y:         a.Norm(t.Value),
-			Rectangle: draw.Rectangle{draw.Point{Y: -h / 2}, draw.Point{Y: h / 2}},
+			Y: a.Norm(t.Value),
+			Rectangle: draw.Rectangle{Min: draw.Point{Y: -h / 2},
```
kortschak: New line after `Rectangle{` and after the last draw.Point close brace.

axis.go
```diff
@@ -244,8 +247,9 @@ func (a *horizontalAxis) GlyphBoxes(*Plot) (boxes []GlyphBox) {
 		}
 		w := a.Tick.Label.Width(t.Label)
 		box := GlyphBox{
-			X:         a.Norm(t.Value),
-			Rectangle: draw.Rectangle{draw.Point{X: -w / 2}, draw.Point{X: w / 2}},
+			X: a.Norm(t.Value),
+			Rectangle: draw.Rectangle{Min: draw.Point{X: -w / 2},
```
kortschak: Same as below.

plotter/barchart.go
```diff
@@ -89,65 +97,111 @@ func (b *BarChart) StackOn(on *BarChart) {
...
-			{xmin, ymin},
-			{xmin, ymax},
-			{xmax, ymax},
-			{xmax, ymin},
+		valMin := trVal(bottom)
+		valMax := trVal(bottom + ht)
+
+		var pts []draw.Point
+		var poly []draw.Point
+		switch b.Horizontal {
```
kortschak: if instead of switch.

plotter/barchart.go
```diff
@@ -89,65 +97,111 @@ func (b *BarChart) StackOn(on *BarChart) {
...
+			}
+			poly = c.ClipPolygonX(pts)
 		}
-		poly := c.ClipPolygonY(pts)
 		c.FillPolygon(b.Color, poly)
 
-		pts = append(pts, draw.Point{xmin, ymin})
-		outline := c.ClipLinesY(pts)
+		var outline [][]draw.Point
+		switch b.Horizontal {
```
kortschak: ditto

plotter/barchart.go
```diff
@@ -89,65 +97,111 @@ func (b *BarChart) StackOn(on *BarChart) {
...
+	}
 
 	for i, ht := range b.Values {
-		x := b.XMin + float64(i)
-		xmin := trX(float64(x))
-		if !c.ContainsX(xmin) {
-			continue
+		catVal := b.XMin + float64(i)
+		catMin := trCat(float64(catVal))
+		switch b.Horizontal {
```
kortschak: if instead of switch.

plotter/barchart.go
```diff
@@ -89,65 +97,111 @@ func (b *BarChart) StackOn(on *BarChart) {
 
 // Plot implements the plot.Plotter interface.
 func (b *BarChart) Plot(c draw.Canvas, plt *plot.Plot) {
-	trX, trY := plt.Transforms(&c)
+	var trCat, trVal func(float64) vg.Length
+	switch b.Horizontal {
```
kortschak:
```
trCat, trVal = plt.Transforms(&c)
if b.Horizontal {
    trCat, trVal = trVal, trCat
}
```

plotter/barchart.go
```diff
@@ -89,65 +97,111 @@ func (b *BarChart) StackOn(on *BarChart) {
...
 func (b *BarChart) GlyphBoxes(plt *plot.Plot) []plot.GlyphBox {
 	boxes := make([]plot.GlyphBox, len(b.Values))
 	for i := range b.Values {
-		x := b.XMin + float64(i)
-		boxes[i].X = plt.X.Norm(x)
-		boxes[i].Rectangle = draw.Rectangle{
-			Min: draw.Point{X: b.Offset - b.Width/2},
-			Max: draw.Point{X: b.Offset + b.Width/2},
+		cat := b.XMin + float64(i)
+		switch b.Horizontal {
```
kortschak: ditto

ctessum:
How should they be distinct? I combined them because the best thing I could think of for a test was also the best thing I could think of for an example. Should the test use draw.Recorder instead?

kortschak:
They don't have to be semantically distinct, just don't try to use an example as a test. We already keep files of graphic output in testdata, so that's fine.

I don't have a problem with duplication here. There may well be divergence as other cases come up, but wait for @eaburns.

eaburns:
For box plots, there are different types for vertical and horizontal. We should be consistent and either use different types for both or use a Boolean flag for both.

ctessum:
I would vote for the boolean flag, as it makes the API more straightforward in my opinion, but I will defer to the judgement of others.

ctessum:
PTAL

plotter/boxplot.go
```diff
@@ -287,19 +312,11 @@ func (o boxPlotOutsideLabels) Label(i int) string {
...
-type HorizBoxPlot struct{ *BoxPlot }
-
-// MakeHorizBoxPlot returns a HorizBoxPlot,
-// plotting the values in a horizontal box plot
-// centered along a fixed location of the y axis.
-func MakeHorizBoxPlot(w vg.Length, loc float64, vs Valuer) (HorizBoxPlot, error) {
-	b, err := NewBoxPlot(w, loc, vs)
-	return HorizBoxPlot{b}, err
-}
+type horizBoxPlot struct{ *BoxPlot }
```
eaburns: Could you please add a TODO note that we should just merge this code into the methods above, like you've done for bar charts?

plotter/boxplot.go
```diff
@@ -397,3 +414,25 @@ type horizBoxPlotOutsideLabels struct {
 func (o horizBoxPlotOutsideLabels) XY(i int) (float64, float64) {
 	return o.box.Value(o.box.Outside[i]), o.box.Location
 }
+
+// ValueLabels implements both the Valuer
+// and Labellel interfaces.
```
eaburns: Labeller

plotter/main.go
```diff
@@ -262,10 +262,13 @@ func Example_groupedHorizontalBoxPlots() *plot.Plot {
 
 	w := vg.Points(20)
 	for y := 0.0; y < 3.0; y++ {
-		b0 := must(plotter.MakeHorizBoxPlot(w, y, uniform)).(plotter.HorizBoxPlot)
+		b0 := must(plotter.NewBoxPlot(w, y, uniform)).(*plotter.BoxPlot)
+		b0.Horizontal = true
```
eaburns: Seeing this, I agree that it's better than the previous approach.

eaburns:
Thanks a lot for doing this.
I'm not sure why all the test files changed. I assume that this was related to @kortschak's comment, but I wasn't following that closely. Assuming so, the rest LGTM.

ctessum:
I wanted to make a test for the horizontal boxplot and quartplot, and once I did that it seemed to make sense to make the rest of them consistent. Since all the examples in main.go are now duplicated, should I delete main.go?

eaburns:
I think so. main.go was a hack because I didn't know how to test this. You've all figured it out, and this CL obviates the need for the hack.

ctessum:
PTAL

plotter/boxplot.go
```diff
@@ -287,19 +312,13 @@ func (o boxPlotOutsideLabels) Label(i int) string {
...
-type HorizBoxPlot struct{ *BoxPlot }
-
-// MakeHorizBoxPlot returns a HorizBoxPlot,
-// plotting the values in a horizontal box plot
-// centered along a fixed location of the y axis.
-func MakeHorizBoxPlot(w vg.Length, loc float64, vs Valuer) (HorizBoxPlot, error) {
-	b, err := NewBoxPlot(w, loc, vs)
-	return HorizBoxPlot{b}, err
-}
+// TODO: Merge code for horizontal and vertical box plats as has been done for
```
sbinet: s/plats/plots/

plotter/errbars_test.go
```diff
@@ -0,0 +1,126 @@
...
+	handleEx(err)
+	yerrs, err := NewYErrorBars(data)
+	handleEx(err)
+	p.Add(scatter, xerrs, yerrs)
+
+	checkPlot("examplePlots", "errorBars", "png", p, 200, 200,
+		handleEx, exampleLog)
+
+	// Output:
+	// Image can be seen at https://github.com/gonum/plot/plotter/examplePlots/errorBars.png.
```
sbinet: might as well give the correct path:
`https://github.com/gonum/plot/tree/master/plotter/examplePlots/errorBars.png`
:)

ctessum:
PTAL. @kortschak ?

plotter/barchart_test.go
```diff
@@ -0,0 +1,178 @@
...
+	p2.NominalY(horizontalLabels...)
+	checkPlot("examplePlots", "horizontalBarChart", "png", p2, 100, 100,
+		handleEx, exampleLog)
+
+	// Output:
+	// Image can be seen at https://github.com/gonum/plot/tree/master/plotter/examplePlots/verticalBarChart.png.
+	// Normally, you would use plot.Save().
+	// Image can be seen at https://github.com/gonum/plot/tree/master/plotter/examplePlots/horizontalBarChart.png.
+	// Normally, you would use plot.Save().
+
```
kortschak: Delete blank line,

plotter/general_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+type logFunc func(format string, args ...interface{})
+
+var testLog = func(t *testing.T) logFunc {
+	return func(s string, arg ...interface{}) {
+		t.Logf(s, arg...)
+	}
+}
+
+var exampleLog = func(s string, arg ...interface{}) {
```
kortschak: Replace instances of `exampleLog` with `fmt.Printf`.

plotter/barchart_test.go
```diff
@@ -0,0 +1,178 @@
...
+func ExampleBarChart() {
+	// Create the plot values and labels.
+	values := Values{0.5, 10, 20, 30}
+	verticalLabels := []string{"A", "B", "C", "D"}
+	horizontalLabels := []string{"Label A", "Label B", "Label C", "Label D"}
+
+	// Create a vertical BarChart
+	p1, err := plot.New()
+	handleEx(err)
```
kortschak: I'm really not a fan of using error handling functions for character count reduction at the best of times, but especially not in example code when the function is trivial and definition is not available to the reader.

plotter/general_test.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	lf("Image can be seen at https://github.com/gonum/plot/tree/master/plotter/%s/%s.%s.\n"+
+		"Normally, you would use plot.Save().\n", dir, name, ext)
+}
+
+// errFunc is an interface for functions that deal with errors.
+type errFunc func(err error)
+
+// handleEx prints err if err!=nil
+var handleEx = func(err error) {
```
kortschak: This is not what I would have expected this to do.

plotter/general_test.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		fmt.Println(err)
+	}
+}
+
+// handleTest returns a function that fails the test if err!=nil
+var handleTest = func(t *testing.T) errFunc {
```
kortschak: Is there a reason this needs to be a var?

plotter/general_test.go
```diff
@@ -0,0 +1,128 @@
...
+	return func(err error) {
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+}
+
+type logFunc func(format string, args ...interface{})
+
+var testLog = func(t *testing.T) logFunc {
```
kortschak: Is there a reason this needs to be a var?

plotter/bubbles_test.go
```diff
@@ -40,3 +43,82 @@ func TestBubblesRadius(t *testing.T) {
 		}
 	}
 }
+
+func ExampleBubbles() {
+
```
kortschak: Delete blank line.

plotter/bubbles_test.go
```diff
@@ -40,3 +43,82 @@ func TestBubblesRadius(t *testing.T) {
...
+	checkPlot("examplePlots", "bubbles", "png", p, 200, 200,
+		handleEx, exampleLog)
+
+	// Output:
+	// Image can be seen at https://github.com/gonum/plot/tree/master/plotter/examplePlots/bubbles.png.
+	// Normally, you would use plot.Save().
+}
+
+func TestBubbles(t *testing.T) {
+
```
kortschak: Delete blank line.

plotter/errbars_test.go
```diff
@@ -0,0 +1,126 @@
...
+	"math/rand"
+	"testing"
+
+	"github.com/gonum/plot"
+	"github.com/gonum/plot/vg/draw"
+)
+
+// ExampleErrors draws points and error bars.
+func ExampleErrors() {
+
```
kortschak: Delete blank line.

plotter/functions_test.go
```diff
@@ -0,0 +1,95 @@
...
+	"math"
+	"testing"
+
+	"github.com/gonum/plot"
+	"github.com/gonum/plot/vg"
+)
+
+// ExampleFunction draws some functions.
+func ExampleFunction() {
+
```
kortschak: Delete blank line.

plotter/functions_test.go
```diff
@@ -0,0 +1,95 @@
...
+	checkPlot("examplePlots", "functions", "png", p, 200, 200,
+		handleEx, exampleLog)
+
+	// Output:
+	// Image can be seen at https://github.com/gonum/plot/tree/master/plotter/examplePlots/functions.png.
+	// Normally, you would use plot.Save().
+}
+
+func TestFunction(t *testing.T) {
+
```
kortschak: Delete blank line.

plotter/general_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+	"github.com/gonum/plot"
+	"github.com/gonum/plot/vg"
+)
+
+var generateTestData = flag.Bool("regen", false, "Uses the current state to regenerate the test data.")
+
+// checkPlot checks a generated plot against a previously created reference.
+// If generateTestData = true, it regenerates the reference first.
+func checkPlot(dir, name, ext string, p *plot.Plot, width, height vg.Length,
```
kortschak: Don't line wrap this.

plotter/general_test.go
```diff
@@ -0,0 +1,128 @@
...
+	errf errFunc, lf logFunc) {
+
```
kortschak: Delete blank line.

plotter/general_test.go
```diff
@@ -0,0 +1,128 @@
...
+	filename := filepath.Join(dir, fmt.Sprintf("%s.%s", name, ext))
+
+	c, err := p.WriterTo(width, height, ext)
+	errf(err)
```
kortschak: s/errf/handleError/g

plotter/general_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+	want, err := ioutil.ReadAll(f)
+	errf(err)
+	f.Close()
+	if !bytes.Equal(buf.Bytes(), want) {
+		errf(fmt.Errorf("image mismatch for %s\n", filename))
+		return
+	}
+	lf("Image can be seen at https://github.com/gonum/plot/tree/master/plotter/%s/%s.%s.\n"+
```
kortschak: s/lf/log/g

I don't like the second sentence it doesn't fit in a test log. Just add a comment in the examples.

plotter/heat_test.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"testing"
+
+	"github.com/gonum/matrix/mat64"
+	"github.com/gonum/plot"
+	"github.com/gonum/plot/palette"
+)
+
+func ExampleHeatMap() {
+
```
kortschak: Delete blank line.

plotter/heat_test.go
```diff
@@ -0,0 +1,66 @@
...
+	checkPlot("examplePlots", "heatMap", "png", p, 100, 100,
+		handleEx, exampleLog)
+
+	// Output:
+	// Image can be seen at https://github.com/gonum/plot/tree/master/plotter/examplePlots/heatMap.png.
+	// Normally, you would use plot.Save().
+}
+
+func TestHeatMap(t *testing.T) {
+
```
kortschak: Delete blank line.

plotter/boxplot_test.go
```diff
@@ -0,0 +1,213 @@
...
+	"math/rand"
+	"testing"
+
+	"github.com/gonum/plot"
+	"github.com/gonum/plot/vg"
+)
+
+func ExampleBoxPlot() {
+	// Create the sample data.
+	rand.Seed(int64(0))
```
kortschak: rand.Seed(0)

Note that the standard rand seed is 1 (initially package rand seeds with 1).

plotter/scatter_test.go
```diff
@@ -0,0 +1,136 @@
...
+	p.Add(s, l, lpLine, lpPoints)
+	p.Legend.Add("scatter", s)
+	p.Legend.Add("line", l)
+	p.Legend.Add("line points", lpLine, lpPoints)
+	// Add a GlyphBox plotter for debugging.
+	p.Add(NewGlyphBoxes())
+
+	checkPlot("examplePlots", "scatterTest", "png", p, 200, 200,
+		handleTest(t), testLog(t))
+
```
kortschak: Delete blank line.

plotter/general_test.go
ctessum: exampleLog is used in the checkPlot function, which is a single function that is used to check plots for both tests and examples. Because errors and log messages are handled differently in examples and tests, checkPlot takes as arguments functions for handling errors and log messages. 

In order to make this change, there would also need to be two versions of checkPlot, one for examples and one for tests. I can see good and bad sides of doing this so I will defer to the judgement of others.

plotter/barchart_test.go
ctessum: I did this because how we handle the error is not relevant to the example, so I didn't see any benefit from making the definition available to the reader. I see the benefit of doing it this way as allowing the reader to focus on the code that is relevant, whereas the potential disbenefit is that it could possibly cause confusion for the reason you point out. I will defer to your judgement regarding whether the benefits outweigh the drawbacks. A third option could be just to make the function name more descriptive or add a comment saying that the user can choose how to handle the error or something like that.

plotter/general_test.go
ctessum: Printing an error message will cause the example to fail in a way that is instructive. I figured that panicking or exiting the program would be more heavy-handed than necessary. This is another argument for not explicitly handling the errors in the examples because we don't want people to just copy the example error handling code, as it would not be appropriate in almost any other case.

It needs to be a var so it can be passed to checkPlot() but we create separate checkPlots for tests and examples then it wouldn't need to be. Same goes for below.

plotter/general_test.go
ctessum: Another option would be to create wrappers for checkPlot:

``` go
func checkPlotExample(dir, name, ext string, p *plot.Plot, width, height vg.Length) {
	checkPlot(dir, name, ext, p, width, height,
		func(err error) {
			if err != nil {
				fmt.Println(err)
			}
		},
		func(s string, arg ...interface{}) {
			fmt.Printf(s, arg...)
		},
	)
}

func checkPlotTest(dir, name, ext string, p *plot.Plot, width, height vg.Length, t *testing.T) {
	checkPlot(dir, name, ext, p, width, height,
		func(t *testing.T) errFunc {
			return func(err error) {
				if err != nil {
					t.Fatal(err)
				}
			}
		}(t),
		func(t *testing.T) logFunc {
			return func(s string, arg ...interface{}) {
				t.Logf(s, arg...)
			}
		}(t),
	)
}

```

I'm not sure the extra complexity is worth it to avoid the copypasting though.

plotter/general_test.go
kortschak: A log.Fatalf would do the same thing and be more explanatory to the reader - which is the point of an example.

>It needs to be a var so it can be passed to checkPlot() but we create separate checkPlots for tests and examples then it wouldn't need to be. Same goes for below.

This is not true. A func can be used as a parameter to a function without being a declared variable. The issue here is that fmt.Printf returns (int, error), however that is a separate issue from the mode of declaration of the function.

http://play.golang.org/p/kTO9j7QEoR

plotter/barchart_test.go
kortschak: I can understand the motivation, but I don't agree that the reduction in line count is worth the potential cognitive dissonance cause by calling something that has not obvious definition.

plotter/general_test.go
kortschak: TBH I'd prefer two functions, but wait for others.

plotter/barchart_test.go
eaburns: I think that the motivation is to hide test-irrelevant logic, not to reduce the line count. But I agree with @kortschak that
```
if err := nil { log.Fatalf(err) }
```
isn't a lot of logic to be hidden.

I don't think that we need to cater to folks that will copy/paste this, and I'm not sure that `handleErr(err)` will fix the copy/paste problem anyway. Something like `handleErr` is not sufficient to drop in proper error handling; true error handling will need to do something within this function itself. For example, it will need to return the error or fix whatever caused it.

(BTW, do we still plan to fix plot.New so that it doesn't return an error, since there is no possible recovery for it?)

plotter/general_test.go
eaburns: It looks like there is a lot of duplication between the Examples and the Tests. Is it true that the only difference is which functions are passed to `checkPlot`?

Why not have the Test function call the Example functions? If each Example saves the plot with a fixed name (say, "plot.png") then the tests can call `Example_XYZ` and pick up the saved plot from the know path to do their testing.

This solves three problems:
1. No more code duplication.
2. The examples can actually call `p.Save`, not `checkPlot`.
3. No more worry about funcs/vars.

eaburns:
I hate to ask it, but it seems like the majority of this PR has become about refactoring tests. Can we split it into two PRs, one which refactors tests and one which adds horizontal bar charts?

plotter/general_test.go
ctessum: I think the idea is that the tests and examples may diverge in the future. I'm fine with either way though.

ctessum:
That would separate the PR from this code review. Would it work for the final squashed version to be 2 commits within one pull request?

eaburns:
You're right. Having the squashed version be two commits sounds perfect. Thanks!

plotter/general_test.go
ctessum: @kortschak do you have an opinion about this? Ethan's suggestion seems like an elegant solution but I'm not sure if it's consistent with your original rationale for separating the tests and examples.

kortschak: I dislike the idea of writing to a fs when it's not strictly necessary, but I guess this is the least bad approach, so long as the examples do not have an `// Output:` line (they should not be run again) and the calling test cleans up after itself.

eaburns: I also dislike writing to the fs in tests, but they already seem to be doing that.

ctessum:
PTAL

vg/vg_test.go
```diff
@@ -42,7 +42,7 @@ func TestLineWidth(t *testing.T) {
 		for _, typ := range formats {
 			p, err := lines(w)
 			if err != nil {
-				log.Fatalf("failed to create plot for %v:%s: %v", w, typ, err)
+				log.Fatal("failed to create plot for %v:%s: %v", w, typ, err)
```
kortschak: This should be Fatalf (and below).

kortschak:
Can you put the golden files in a directory called "testdata".

plotter/general_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+	"github.com/gonum/plot"
+	"github.com/gonum/plot/vg"
+)
+
+var generateTestData = flag.Bool("regen", false, "Uses the current state to regenerate the test data.")
+
+// checkPlot checks a generated plot against a previously created reference.
+// If generateTestData = true, it regereates the reference.
+func checkPlot(ExampleFunc func(), t *testing.T, paths ...string) {
```
kortschak: I think it would be nicer to take a list of basenames rather than paths - the path base can be Chdir'd to in lieu.

As an aside, does this behave correctly when invoked from somewhere other than `$GOPAPTH/.../gonum/plot`?

plotter/boxplot_test.go
```diff
@@ -0,0 +1,158 @@
...
+	// the given names for y=0, y=1 and y=2.
+	p2.NominalY("Uniform\nDistribution", "Normal\nDistribution",
+		"Exponential\nDistribution")
+
+	err = p2.Save(200, 200, "examplePlots/horizontalBoxPlot.png")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Now, make a grouped box plot. /////////////////////////////////////
```
kortschak: Delete following blank line and trailing slashes.

plotter/barchart_test.go
```diff
@@ -0,0 +1,175 @@
...
+		log.Fatal(err)
+	}
+	p2.Add(horizontalBarChart)
+	p2.NominalY(horizontalLabels...)
+	err = p2.Save(100, 100, "examplePlots/horizontalBarChart.png")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Now, make a different type of BarChart. //////////////////////////
```
kortschak: Delete following blank line and trailing slashes.

plotter/barchart_test.go
```diff
@@ -0,0 +1,175 @@
...
+	p.NominalX("Zero", "One", "Two", "Three", "Four", "",
+		"Six", "Seven", "Eight", "Nine", "Ten")
+
+	p.Add(NewGlyphBoxes())
+	err = p.Save(300, 250, "examplePlots/barChart2.png")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Now, make a stacked BarChart. /////////////////////////////////////
```
kortschak: Delete following blank line and trailing slashes.

plotter/boxplot_test.go
```diff
@@ -0,0 +1,158 @@
...
+	"math/rand"
+	"testing"
+
+	"github.com/gonum/plot"
+	"github.com/gonum/plot/vg"
+)
+
+func ExampleBoxPlot() {
+	// Create the sample data.
+	rand.Seed(0) // The default random seed is 1.
```
kortschak: Since the example plots are not run by the testing harness without help from us, this can be factored out to checkPlot just prior to the call to the example function (and at the same time, may as well make it seed with 1).

plotter/general_test.go
```diff
@@ -0,0 +1,155 @@
...
+	ExampleFunc()
+
+	// Read the images we've just generated and check them against the
+	// Golden Images.
+	for i, path := range paths {
+		errored := false
+		f, err := os.Open(path)
+		if err != nil {
+			t.Error(err)
+			errored = true
```
kortschak: Use t.Failed().

kortschak:
On the use of log.Fatal[f], I think we need to use log.Panic[f]. This is horrible, but is the only way we can guarantee that checkPlot will retain control if any calls error.

plotter/boxplot_test.go
```diff
@@ -0,0 +1,158 @@
...
+	// the given names for x=0, x=1 and x=2.
+	p1.NominalX("Uniform\nDistribution", "Normal\nDistribution",
+		"Exponential\nDistribution")
+
+	err = p1.Save(200, 200, "examplePlots/verticalBoxPlot.png")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Now, make the same plot but horizontal.
```
eaburns: Since this is testing a different plot, can it please go into a different example function?

eaburns:
Thanks for sticking with this change for so long and through so much review. I think that it's close to done. I haven't looked over as thoroughly as @kortschak, but it looks pretty good to me. Just minor comments and it obviously needs a rebase. From a quick glance, I see 3 main changes that should each have a commit (maybe you can find more---all the better):
1. minor cleanups (re-indenting, spelling fixes, etc.)
2. testing refactors
3. horizontal bar charts

I also think that we should open a bug to re-implement box and quartile plots so that they don't use two types, but so that they look more like the bar chart implementation. It'll be more consistent that way, and maybe it would be a useful exercise for someone who wants to better learn the codebase.

plotter/boxplot_test.go
ctessum: My understanding is that godoc will only match one example to each type, so if my understanding is correct if we split it into multiple examples we would have to choose which one goes in the godoc.

plotter/general_test.go
ctessum: Yes, it seems that the test package moves to the correct working directory before running each test.

ctessum:
Any other changes? If not, I can squash it into the three commits listed above, at least as close as possible.

eaburns:
If you got all of the comments above, squash away!

kortschak:
Not quite what I meant with the save paths, but I can fix that in another PR.

PR #232, opened 2015-12-01
Allows function plotting to work with discontinuous functions.

jcnnghm:
Consider the function sqrt(x^2 - 1), graphed as follows:

```go
package main

import (
        "image/color"
        "math"

        "github.com/gonum/plot"
        "github.com/gonum/plot/plotter"
        "github.com/gonum/plot/vg"
)

func main() {
        p, err := plot.New()
        if err != nil {
                panic(err)
        }
        p.Title.Text = "Functions"
        p.X.Label.Text = "X"
        p.Y.Label.Text = "Y"

        // A quadratic function x^2
        discont := plotter.NewFunction(func(x float64) float64 { return math.Sqrt(x*x - 1) })
        discont.Color = color.RGBA{B: 255, A: 255}
        discont.Samples = 200

        // Add the functions and their legend entries.
        p.Add(discont)
        p.Legend.Add("sqrt(x^2 - 1)", discont)
        p.Legend.ThumbnailWidth = 0.5 * vg.Inch

        // Set the axis ranges.  Unlike other data sets,
        // functions don't set the axis ranges automatically
        // since functions don't necessarily have a
        // finite range of x and y values.
        p.X.Min = -3
        p.X.Max = 3
        p.Y.Min = -1
        p.Y.Max = 3

        // Save the plot to a PNG file.
        if err := p.Save(4*vg.Inch, 4*vg.Inch, "functions.png"); err != nil {
                panic(err)
        }
}
```

In `master`, this code will infinite loop, because the function returns `NaN` for x values between -1 and 1, where the solution isn't a real number.  

Instead of plotting all of the sample points at once, my change plots each continuous section of real-number solutions independently, allowing for discontinuous functions.  Running the inline code with my change creates this graph:

functions


sbinet:
great! could you also add a test for this?

kortschak:
I like this, but it does need the above function to be included in a test (probably wait until #229 is merged since that touches many test files and probably makes testing easier).

On a side note, having contiguous lines allows an xkcd style vg.Canvas to be implemented, which is something I want for reasons.

jcnnghm:
Cool, I'll wait for the testing changes in #229.

plotter/functions.go
```diff
@@ -39,7 +41,20 @@ func (f *Function) Plot(c draw.Canvas, p *plot.Plot) {
 		line[i].X = trX(x)
 		line[i].Y = trY(f.F(x))
 	}
-	c.StrokeLines(f.LineStyle, c.ClipLinesXY(line)...)
+
+	// For every continuous block of non-NaN Y values, stroke lines
```
eaburns: Is it possible to roll this into the previous loop? I'm thinking about something like this:
```
var line []draw.Point
for i := 0; i < f.Samples; i++ {
	x := p.X.Min + float64(i)*d
	y := f.F(x)
	if !math.IsNan(y) {
		line = append(line, draw.Point{X: trX(x), Y: trY(Y) })
	}
	// If we are at the end of a contiguous segment, stroke.
	if (math.IsNan(y) || i == f.Samples-1) {
		if len(line) > 1 {
			c.StrokeLines(f.LineStyle, c.ClipLinesXY(line)...)
		}
		line = line[:0]
	}
}
```

eaburns:
Any news on this? I think that we all liked it, and #229 is now in.

sbinet:
ping?

eaburns:
Any news on this? We are trying to clean up our PRs. It seems like this change has some outstanding comments. Do you plan to address them? If not, we plan to close this soon.

eaburns:

PR #234, opened 2015-12-10
Move functions requiring wide dependencies to their own files.

yohcop:
This makes it easier to use gonum/plot without having to install
all of the dependencies if one does not care about them. For example,
if tiff, pdf, eps and x11 output are not required, it is easy to delete
the vg/vgimg/tiff.go file, vg/draw/formattedcanvas.go (depends on all
the formats known to plot), and the directories vg/vgpdf, vg/vgeps and
vg/vgx11.

I left pdf and jpg in vg/vgimg/vgimg.go, since they do not require a
third party library.

### DIFF
--- a/vg/draw/canvas.go
+++ b/vg/draw/canvas.go
@@ -5,16 +5,11 @@
 package draw
 
 import (
-	"fmt"
 	"image/color"
 	"math"
 	"strings"
 
 	"github.com/gonum/plot/vg"
-	"github.com/gonum/plot/vg/vgeps"
-	"github.com/gonum/plot/vg/vgimg"
-	"github.com/gonum/plot/vg/vgpdf"
-	"github.com/gonum/plot/vg/vgsvg"
 )
 
 // A Canvas is a vector graphics canvas along with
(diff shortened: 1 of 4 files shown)


yohcop:
FWIW, I just noticed that tiff is in golang.org/x/image which is needed anyway in vg/draw. So I can revert that change if needed.

kortschak:
I don't really understand the motivation behind this.

yohcop:
Yeah, it's a little weird :). The general idea is to make vendoring easy by requiring minimal modifications after the project is cloned.

When a project uses vendoring, it may be undesired to have to pull a large number of unneeded dependencies within the project' source tree. And a fair number of go projects use vendoring - enough that go 1.5 has a vendoring experiment (https://docs.google.com/document/d/1Bz5-UB7g2uPBdOx-rw5t9MxJwkfpx90cqG9AFL0JAYo/edit)

Ideally, there would be the main plot library requiring the very minimum amount of dependencies (things that use the standard library such as png can stay), and then, for other outputs such as svg, x11, eps, pdf, a separate package would be available. So if a project does not use those packages, other dependencies such as bitbucket.org/zombiezen/gopdf are not necessary and there's no need to clone them at all. But of course for this to work the core packages should not depend on plot/vg/vgpdf either.

Currently plot/ and plot/vg depend on plot/vg/vgpdf only because of the NewFormattedCanvas function and Save and WriterTo in plot.go (I forgot those in this pull request, but I can add them), which creates dependencies to all of those packages. They are mostly convenience functions though, and deleting them does not break the rest of the functionality.

What I did is move those functions to a separate file. Now the project using vendoring just has to delete those files. It's not as pretty as "just not use a package", but it's easier than having to find and delete functions inside source files.

I picked this solution because it does not break anything. But if you would be ok with moving NewFormattedCanvas, WriterTo and Save to their own package, it's even better.

I hope that makes sense?

eaburns:
Previously, there was discussion about changing backends to register with the plot package, but I don't remember the result of that discussion. I think that would be a better approach to get what you want.

kortschak:
Yes, I actually originally wrote it that way when the current approach was implemented, but the registration approach was rejected. The discussion will be in the PR presumably. From memory the issue is that Save is complicated by that since the format string is required.

kortschak:
I misremembered, but #193 is the PR that has that.

yohcop:
If there are plans to make this better, forget this merge request. It is sort of a hack of course, my main goal was to maintain full api compatibility -- but I'm not sure if you care much about that right now. If that pull request #193 is still being considered, it is 100% better than my hack :)

kortschak:
#193 requires work to get it to merge, I'm happy to do that if there is a push for it. There is presumably other discussion backing why it has been left, but I don't remember where it is. @eaburns, do you recall?

eaburns:
I'm not sure that I ever saw #193, but I found some relevant discussion where, of course, I was arguing against registration. And upon re-reading my argument, I'm convinced by what I said, and you seemed happy with it too, @kortschak.

https://groups.google.com/d/topic/gonum-dev/DsL5sy2_QWE/discussion

yohcop:
Ok, cool. It appear we all want the same thing: fix/get rid of the dependency of the main packages on individual drivers.

Re. the discussion, I'd like to ask not to use build tags. It may be fine when the software is built manually or by your own scripts, but in some cases you may not be able to control this (e.g. appengine). That being said, using build tags would still require cleaning up those dependencies, so I'd be OK anyway.

I'm happy to send a pull requrest just deleting those 3 problematic functions btw :)

kortschak:
@eaburns Thanks for finding that. Yes, that was what I was vaguely remembering. Do you have a plan to move this forward? I'm happy to make the changes if you do.

eaburns:
I don't have any plans other than what was said in that discussion. If you are interested, feel free to make the change. I'll take a look at it. However, I will be away next week, so there may be a bit of delay.

eaburns:

Issue #174, opened 2015-03-17
Add support for a second Y axis


eaburns:
Original [issue 146](https://code.google.com/p/plotinum/issues/detail?id=146) created by eaburns on 2014-11-28T17:44:15.000Z:

I have implemented support for a second Y axis here: https://code.google.com/r/cashraistlin-plotinum/source/detail?r=f71f5b39b33d58a397250165730a0d2ecf44ecac

Would you mind to review and merge the changes?

PR #235, opened 2015-12-13
Adding multiple Y axes.

yohcop:
The same thing could be made with the X axis, but it's far less common,
so I didn't do it.

The interface stays mostly the same. Instead of accessign `plt.Y`,
`plt.Ys[0]` (or any other index) must be used.

`plt.AddYAxis()` adds a new Y axis, and `plt.AddWithAxis` adds a plotter
with the given axis.

Tests have been updated. but only the svg and pdf golden files changed.
The differences are only in the order things are painted. The Y axis is
now painted before the x axis.

plotter/main has been updated with an example using 2 Y axes. More than
two axes are also possible of course.

--
There is one possibly awkward thing: I had to change the Plotter and GlyphBoxer interfaces to take the x and y axis to paint with. This makes sense for plotters, but it is slightly weird for the Axis struct since they are not painted within another axis. Maybe Axis and Plot don't really need to implement GlyphBoxer and Plotter?

Suggestions and comments welcome!


kortschak:
Please no.

https://www.lri.fr/~isenberg/publications/papers/Isenberg_2011_ASO.pdf
http://www.perceptualedge.com/articles/visual_business_intelligence/dual-scaled_axes.pdf

A useful alternative would be to be able to specify where the axis is placed. This I believe is part of another issue.

In addition to this, the main focus at the moment is to get plot to a point where we can declare API stability - there is an informal deadline of the end of the year (which we will miss) because plotinum becomes unavailable after then. It would be much better to focus development effort on that rather than new features (other than those that are already accepted and on the roadmap).

yohcop:
OK fine, I can keep this in my repo. That being said... :)

Regarding the API. I can understand that. Although it's not much of an API change I think compared to what's currently in the repo. But ok.

But regarding the two papers linked, I don't quite understand your point. Regardless of what functionality you provide, people can always make "meaningless" or deceptive charts (just randomly use a log scale and people can read your charts differently...). Is gonum/plot targeted at a single use-case? like financial charts?

My use-case is fitness-related charts. Say graph run or ride data. X axis would be time or distance. If I want to overlay say heart rate, elevation, pace, cadence, temperature, power, etc... Of course, not necessarily 20 things at a time, but often elevation is drawn in the background, and one or two time series painted as lines over it, like HR (heart rate). Because if HR increases, it's interesting to see that the cause could be elevation change, or pace change, etc. You just can't plot this unrelated data with the same Y axis. And plotting them separately sometimes (not always) make things hard to read or correlate events. Or, you can stack things up. But it's not better as it fills up the page, requiring scrolling, and now I can't even look at HR and elevation at the same time... I think this is a reasonable chart: http://leppoc.net/tmp/strava.png, and it's not possible to draw this without multiple axes.

My point is that I don't think this library takes a stance on what you should and should not be doing (and it would be hard - but if it does, let me know ;) ). I can already draw a barchart with multiple series of unrelated "stuff", or stack things that shouldn't be stacked, and use a log scale when not appropriate.
 
So I still believe there are use cases for multi Y axes. Your call if you want this in your library or not, you are free to make an executive decision of course. I need it, and I can still keep that in my repo. I just thought it could be useful to others as well.

yohcop:
For reference, I just found #174 , which someone implemented early this year (it has only one secondary axis, so in a way this is more general), but seemed to have been rejected based on the API (or maybe it was also a "I don't like the way it looks" case ;) )

This change also doesn't draw the axis on the right, but it could be done later.

ctessum:
Yohann, while I can see your point, the strava chart has 3 data series and only two practical places to put axes (right and left). I think this illustrates a fundamental limitation of the approach: it only works for plots with exactly two different data series. You could argue that three axes would be too visually confusing, but I assume the paper linked above argues that even 2 axes are too visually confusing. Other ways to make this type of plot would be with stacked axes: http://www.mdpi.com/sensors/sensors-15-20873/article_deploy/html/images/sensors-15-20873-g008-1024.png, or with one axis where all data series are normalized to their initial, final, or central value: http://www.edwardtufte.com/bboard/images/000311-7405.gif

I think one case where it does make sense to have two sets of axes is when you want to display the same data in two different sets of units (e.g., Celsius and Fahrenheit). This could probably be implemented with fewer changes because it wouldn't require specifying which axis each dataset is plotted to.

kortschak:
> My point is that I don't think this library takes a stance on what you should and should not be doing (and it would be hard - but if it does, let me know ;) ).

There is opinion in the library, yes. For example, there is no pie chart plotter.

> I can already draw a barchart with multiple series of unrelated "stuff", or stack things that shouldn't be stacked, and use a log scale when not appropriate.

It's true that we allow plotting of things that may be misleading, but this is because without that ability it is not possible to plot informative things. The benefit outweighs the cost. The ability to put two series on different bases on the same plot is almost never informative and often deceptive. With reference to the strava chart, I would put them in three panels - note that there is no axis for pace or hr in that chart anyway, so all that is being compared is contemporaneity.

Currently it is not possible, but I think it should be reasonably to overlay plots on the same canvas (the reason it's not currently possible is that a white opaque box is placed behind the plot at the moment - there are other reasons to allow this as well). With the capacity to place an axis on the right, you would then get the behaviour you are asking for.

yohcop:
Fine, the Strava chart is missing the vertical axis. Just an example. But they could be painted the same color as the data to make it clear what you are reading, like this other example from Training Peaks: http://leppoc.net/tmp/tp.png (yes, the pink line is also missing a scale...). This chart even supports more data: http://leppoc.net/tmp/tp2.png. I can quickly answer things like "how many times IF (blue dots) was above 1.0. I can say right away that since September it only happened twice. (note that if I was designing this chart, I would rescale the right-side axes to align tics on the grid to make it even easier to read, like http://leppoc.net/tmp/og-vs-rr.png which I made with gnuplot) And these are just examples, not necessarily perfect ones. The first TP chart is what I usually look at, the second just gets too messy.

Of course they can be stacked (and strava actually does that even if it's not obvious), but they are on top of the elevation in the background, or just independent charts/different panels, but especially on mobile it can take a lot of vertical screen space, or require each plot to be super skinny, flattening all the data, and definitely not making things more readable.

Like in the sensor example Chris sent with multiple panels: It's fine, only if I don't care about actual values. It's not possible to say what was the max of gyro_y.

So even if you think they are bad charts printed on paper, multiple Y axes could still be a good answer in different scenario like mobile.

FWIW, I'm not pushing for my particular implementation. re-drawing on top of the same canvas is also an option. It would be weird for this use-case though, because the first layer would still need to know that a second layer is coming, to leave space on the right for another axis, so they can't be completely independent Plot() calls sharing a canvas.

Finally, I think the API change needed to support this is pretty minimum. We can probably argue and go back and forth many times about deciding if the plots I send are good or bad, tasteful or not, or could be made better, but that's just bikeshedding. I'm not the first asking for this, other libs support this, and it doesn't change much of the API or internal design of the library. So I think it would just be nice to have I suppose. Just consider this a feature request if you like. I needed this, I implemented it, period :)

eaburns:
For the record, I agree with @kortschak.

yohcop:
Yeah, I know you don't like the way it looks. But it sounds a little like preventing your users from picking colors and forcing them to use a predefined palette because you dislike other colors. There are a lot of papers on color scales as well.

Last 2 examples comparing the 2 strategies. The Garmin app uses individual charts, the Training Peaks app uses multiple axes:
http://leppoc.net/tmp/garmin.png vs http://leppoc.net/tmp/tp-mobile.png
The Garmin charts are mostly useless to me when trying to compare power to heart rate. I can't even see both on the same screen. The tp example tells me all I want to know. For example it clearly shows the "heart rate lag" behind power changes, which would be hard to see if they were just individual graphs.

Feel free to close if I'm not going to convince you.

ctessum:
The tp-mobile graph has a forced zero, which is also not a best practice. So it shows that the heart rate lags behind the power changes but it it misrepresents the magnitude of the heart rate changes. It might benefit from plotting each series as a percent change from the average, and then only one axis would be needed.

yohcop:
It would be hard to read for most users: "130% from average HR" doesn't tell much. It may make sense from a pure scientific point of view, but for this application and the usual things users of this app are trying to understand from those graph, percentages don't make sense. I know what 160bpm and 210W mean for me and I can compare them, because I work with those numbers every day. percentage from averages would be meaningless. Seeing that HR went from 120 to 160 is what I want to know, not 0.8 to 1.1.

Also, 10% change in power doesn't necessarily equates to 10% change in heart rate. the relationship is much more complicated than that.

eaburns:

zhongjingjogy:
I have been using Python's Matplotlib for a long time and when I turn to golang, I found it is hard to plot a figure with twin axes. Finally, I got it through to create a plot with twin axes, thanks a lot to @yohcop 's commit !I replace the gonum's plot with yohcop's implementation, and I got something like this.

couple1

PR #237, opened 2016-01-15
vg: provide ParseLength to create a vg.Length from a string

sbinet:
Making vg.Length implement flag.Value allows for nicer CLI, such
as:
 $ my-cmd -plot-x-size=20cm -plot-y-size=15cm

### DIFF
--- a/vg/len.go
+++ b/vg/len.go
@@ -4,6 +4,11 @@
 
 package vg
 
+import (
+	"strconv"
+	"strings"
+)
+
 // A Length is a unit-independent representation of length.
 // Internally, the length is stored in postscript points.
 type Length float64
@@ -28,4 +33,36 @@ func (l Length) Dots(dpi float64) float64 {
 // Points returns the length in postscript points.
 func (l Length) Points() float64 {
 	return float64(l)
+}
+
+// ParseLength parses a Length string.
+// A Length string is a possible signed floating number with a unit.
+// e.g. "42cm" "2.4in" "66pt"
+// If no unit was given, ParseLength assumes it was (postscript) points.
+// Currently valid units are:
+//   mm (millimeter)
+//   cm (centimeter)
+//   in (inch)
+//   pt (point)
+func ParseLength(value string) (Length, error) {
+	var unit Length = 1
+	switch {
+	case strings.HasSuffix(value, "in"):
+		value = value[:len(value)-len("in")]
+		unit = Inch
+	case strings.HasSuffix(value, "cm"):
+		value = value[:len(value)-len("cm")]
+		unit = Centimeter
+	case strings.HasSuffix(value, "mm"):
+		value = value[:len(value)-len("mm")]
+		unit = Millimeter
+	case strings.HasSuffix(value, "pt"):
+		value = value[:len(value)-len("pt")]
+		unit = 1
+	}
+	v, err := strconv.ParseFloat(value, 64)
+	if err != nil {
+		return 0, err
+	}
+	return Length(v) * unit, nil
 }
--- a/vg/vg_test.go
+++ b/vg/vg_test.go
@@ -97,3 +97,68 @@ func lines(w vg.Length) (*plot.Plot, error) {
 
 	return p, nil
 }
+
+func TestParseLength(t *testing.T) {
+	for _, table := range []struct {
+		str  string
+		want vg.Length
+		err  error
+	}{
+		{
+			str:  "42.2cm",
+			want: 42.2 * vg.Centimeter,
+		},
+		{
+			str:  "42.2mm",
+			want: 42.2 * vg.Millimeter,
+		},
+		{
+			str:  "42.2in",
+			want: 42.2 * vg.Inch,
+		},
+		{
+			str:  "42.2pt",
+			want: 42.2,
+		},
+		{
+			str:  "42.2",
+			want: 42.2,
+		},
+		{
+			str: "999bottles",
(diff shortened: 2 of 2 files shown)


sbinet:
alternatively, if adding `*Length.{Set,String}` is felt as too cumbersome, one could put those into a `plotutil.LengthAsFlagValue` function+type:

```go
func LengthAsFlagValue(l *Length) flag.Value {
 return lengthType{l}
}

type lengthType struct {
 ptr *vg.Length
}

func (*lengthType) String() string { ... }
func (*lengthType) Set(v string) error { ... }
```

which could be used like so:

```go
package main

func main() {
  var xsize = 20*vg.Centimeter
  flag.Var(plotutil.LengthAsFlag(&xsize), "xsize", "set plot x-size")
  flag.Parse()
  // ...
}
```

sbinet:
ping?

vg/len.go
```diff
@@ -29,3 +35,36 @@ func (l Length) Dots(dpi float64) float64 {
...
+
+// Set is the method to set a flag value, part of the flag.Value interface.
+// e.g. "42cm" "2.4inch" "66pt"
+// If no unit was given, assume (postscript) points.
+func (l *Length) Set(value string) error {
+	var (
+		unit Length = 1
+	)
+	switch {
+	case strings.HasSuffix(value, "inch"):
```
eaburns: 'in' is a standard abbreviation for inch.

vg/len.go
```diff
@@ -29,3 +35,36 @@ func (l Length) Dots(dpi float64) float64 {
 func (l Length) Points() float64 {
 	return float64(l)
 }
+
+func (l *Length) String() string {
+	return fmt.Sprintf("%vpt", float64(*l))
+}
+
+// Set is the method to set a flag value, part of the flag.Value interface.
+// e.g. "42cm" "2.4inch" "66pt"
```
eaburns: The comment should enumerate all available abbreviations, so that no one has to hunt through the code for them.

eaburns:
I don't think adding these two functions is cumbersome.

vg/len.go
sbinet: done.

vg/len.go
sbinet: done.

sbinet:
BTW: I just added a test for invalid `Set` arguments.
Should I test for the `err.Error()` string or rather for the type of the error returned by `strconv.ParseFloat`?
one is more sensitive to possible string modifications inside the stdlib (AFAIK, there is no binding contract wrt the stability of the content of these strings).
the latter is less precise in its failure modes...

eaburns:
Test the `err.Error()` string. Testing the strings is a bit fragile, but sometimes it really is the simplest. However, when I do this, I usually compare it to a regexp so that I can test only the most important part of the string, in this case "invalid syntax".

kortschak:
I think this goes a step too far. The majority of the work, for a far more generalisable tool, would be to provide a ParseLength function. (I'm not completely convinced that even that is necessary, but it is far less specialised than support for flag).

sbinet:
I am all for generality but `flag.Value` is quite general already. See, one can already implement `ParseLength` thanks to `flag.Value` support:

```go
func ParseLength(s string) vg.Length {
   var v vg.Length
   err := v.Set(s)
   if err != nil {
     panic(err)
   }
   return v
}
```

also, even `time.Duration` implements `flag.Value` :)

kortschak:
If `ParseLength(s string) (Length, error)` is implemented, then flag.Value is trivial to implement in approximately the same number of lines, however it is more in line with the orthogonal approach of Go; vg is an internal library, not something that faces the end user.

Note that `time.Duration` does not implement `flag.Value`, pretty much for the same reasons - the flag package has a function `Duration` that returns a `*time.Duration` and `DurationVar` that takes a `*time.Duration`. They use a `time.durationValue` type internally to hide this messiness.

sbinet:
ok, I'll send a revised version with only `ParseLength(s string) (vg.Length, error)`.
where should this live?
- `github.com/gonum/plot/vg` ?
- `github.com/gonum/plot/plotutils` ?
- elsewhere entirely?

kortschak:
vg seems reasonable. @eaburns?

sbinet:
done. PTAL.

vg/len.go
```diff
@@ -29,3 +35,41 @@ func (l Length) Dots(dpi float64) float64 {
 func (l Length) Points() float64 {
 	return float64(l)
 }
+
+func (l *Length) String() string {
```
kortschak: Do we need Length to be a fmt.Stringer? I know it was necessary for flag.Value implementation, but we're not doing that now.

sbinet: no, it's not needed.
removed.

PTAL.

vg/len.go
```diff
@@ -29,3 +34,37 @@ func (l Length) Dots(dpi float64) float64 {
...
+// A Length string is a possible signed floating number with a unit.
+// e.g. "42cm" "2.4in" "66pt"
+// If no unit was given, ParseLength assumes it was (postscript) points.
+// Currently valid units are:
+//   mm (millimeter)
+//   cm (centimeter)
+//   in (inch)
+//   pt (point)
+func ParseLength(value string) (Length, error) {
+	var (
```
kortschak: This should be a single line var declaration.

kortschak:
LGTM after minor comment addressed.

Please update the PR name before you merge.

PR #240, opened 2016-01-21
plotutil: add GoldenRatio

sbinet:
This CL adds a simple functionality to create plots following the
Golden proportions.

Typical use would look like:

```go
  w, h := plotutil.GoldenRatio(10*vg.Centimeter)
  err := p.Save(w, h, "foo.png")
```

### DIFF
--- a/plotutil/goldenratio.go
+++ b/plotutil/goldenratio.go
@@ -1,0 +1,25 @@
+// Copyright ©2016 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package plotutil
+
+import (
+	"math"
+
+	"github.com/gonum/plot/vg"
+)
+
+// phi is the golden ratio, as defined by:
+//  https://en.wikipedia.org/wiki/Golden_ratio
+var phi = vg.Length(1+math.Sqrt(5)) / 2
+
+// GoldenRatio returns the width and the height such as:
+//  width = size
+//  height = size/phi
+// so that width and height follow the golden proportions.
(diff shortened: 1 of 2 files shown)


eaburns:
Seems good to me.

plotutil/goldenratio_test.go
```diff
@@ -0,0 +1,24 @@
...
+package plotutil
+
+import (
+	"testing"
+
+	"github.com/gonum/plot/vg"
+)
+
+func TestGoldenRatio(t *testing.T) {
+
```
kortschak: Delete empty line here.

sbinet: done.

sbinet:
ping?
(I have "only" 1 SGTM and 0 LGTM :))

plotutil/goldenratio.go
```diff
@@ -0,0 +1,25 @@
...
+	"math"
+
+	"github.com/gonum/plot/vg"
+)
+
+// phi is the golden ratio, as defined by:
+//  https://en.wikipedia.org/wiki/Golden_ratio
+var phi = vg.Length(1+math.Sqrt(5)) / 2
+
+// GoldenRatio returns the width and the height such as:
```
kortschak: s/such as/such that/

kortschak:
Can we not just provide a Phi const and let clients do the division or multiplication? It seems more general to me.

sbinet:
well... I suppose that if we are just forwarding `math.Phi`, it's probably not worth it.
perhaps as an example in the wiki.
or just somewhere else.

abandonning as it doesn't have enough bang for the bucks.

PR #245, opened 2016-02-02
flexible styles (e.g. a custom plot style)

sbinet:
This CL exports a few internals of `plot`:
- `plot.HorizontalAxis`
- `plot.VerticalAxis`
- `plot.Axis`
- `plot.Legend.Draw`

such that user-defined plot styles can be created.

an example of such a user-defined plot style has been added as `plotter.ExampleHistogramGnuplot`.


sbinet:
one of my users (@ebusato) asked whether it was possible to have a more "high energy physics"-y style for his plots.
I revived my old branch.
PTAL.

kortschak:
I'm OK with this.

sbinet:
@eaburns what do you think?

plotter/histogram_gnuplotstyle_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+type GnuplotStyle struct{}
+
+func (s GnuplotStyle) DrawPlot(p *plot.Plot, c draw.Canvas) {
+	if p.BackgroundColor != nil {
+		c.SetColor(p.BackgroundColor)
+		c.Fill(c.Rectangle.Path())
+	}
+
+	{
```
eaburns: Why the extra nesting here?

plotter/histogram_gnuplotstyle_test.go
```diff
@@ -0,0 +1,123 @@
...
+		xpad := p.X.Padding
+		ypad := p.Y.Padding
+		defer func() {
+			p.X.Padding = xpad
+			p.Y.Padding = ypad
+		}()
+	}
+
+	p.X.Padding = 0
```
eaburns: I don't like this. It seems wrong to have a style parameter as part of the plot which we sometimes silently ignore. Can we find a different solution?

plotter/histogram_gnuplotstyle_test.go
sbinet: just to limit the scope of xpad and ypad.
I can remove.

eaburns:
Sorry I'm slow. My wife and I just had a baby, so it's been a very busy and sleepless week.

I don't like GNU plot's style. That is one of the reasons I wrote my own plotting package to begin with. (Actually, Plotinum was my second plotting package, see https://code.google.com/archive/p/caml-spt/) That said, I don't have a problem with this type of change. While I can complain about an ugly style, I don't have the artistic ability actually create a beautiful one myself. That is why Plotters exist, so people with better taste than myself can make their own styles. This seems inline with that idea.

Two things:
 1. I don't like that Plot.Axis.Padding is silently ignored by the GNU plot style. We should find a different way to communicate style parameters. I think that they should be part of the style structs themselves, not the Plot or Axis.
 2. More of a note or complaint: this doesn't entirely mimic GNU plot. GNU plot has the terrible behavior of drawing the tick marks INTO the plot, obscuring the data. If someone truly wants to mimic GNU plot, they would need the ability to do that too. I'm OK leaving it out.

sbinet:
no worries (and congratulations!)

wrt having the ticks inside/outside the data-canvas: we actually might need to support it, if/when we support multiple Y-axes (ie: if you put Y-axis-1 on the left and Y-axis-2 on the right)

I am not hellbent on having a gnuplot-style, it's just the (general looking) style my community is used to.

plotter/histogram_gnuplotstyle_test.go
sbinet: I am open to suggestions :)
I actually tried to come up with a new interface to encompass this padding issue... to no avail.

as you implied, one way to fix this is to shift things around:

- remove Axis.Padding
- add XAxisPadding and YAxisPadding fields inside each of the various `XyzPlotStyle` structs

possibly do the same for all the other style-dependent fields of `plot.Plot`, `plot.Axis` and `plot.Legend` ?
_ie:_ completely separate "data" from "style" for `plot.Plot`, `plot.Axis` and `plot.Legend`.
then, `plot.Plot` would look like:

```go
type Plot struct {
    Title string
     X, Y Axis
     Legend Legend
     Style draw.PlotStyle
    plotters []Plotter
}

type Axis struct {
   Min, Max float64
   Label string
   Scale Normalizer
}
```

and define, say, in `vg/draw`:

```go
type PlotStyle {
   Title {
     Padding vg.Length
     draw.TextStyle
   }
   BackgroundColor color.Color
   X, Y AxisStyle
   Legend LegendStyle
}

type AxisStyle struct {
   Label TextStyle
   LineStyle
   Padding vg.Length
}

type LegendStyle struct {
   draw.TextStyle
   Padding vg.Length
   Top, Left bool
}
```

having said that...

to flip it around: imagine the default style of `plot.Plot` was the gnuplot-like one, _ie_ with 4 borders delimiting the canvas, and thus with 4 axes to draw.
then, implementing the current `plot.Plot` style would consist in ignoring 2 of the axes delimiting the canvas. :)

another example: nominal axes. the current code for NominalX actually overrides whatever was inside `Axis.Padding`: https://github.com/gonum/plot/blob/master/plot.go#L396

actually: another possibility is to modify `plot.New() (*Plot, error)` to take an optional `plot.Style` which would have a default `MakeAxis` field function (or method) to be used in place of `plot.makeAxis`. (so the possible inconsistency of `plot.Plot.{X,Y}.Padding` would be solved by default)

so... which option would you be more in favor of?

sbinet: I am open to suggestions :)
I actually tried to come up with a new interface to encompass this padding issue... to no avail.

as you implied, one way to fix this is to shift things around:

- remove Axis.Padding
- add XAxisPadding and YAxisPadding fields inside each of the various `XyzPlotStyle` structs

possibly do the same for all the other style-dependent fields of `plot.Plot`, `plot.Axis` and `plot.Legend` ?
_ie:_ completely separate "data" from "style" for `plot.Plot`, `plot.Axis` and `plot.Legend`.
then, `plot.Plot` would look like:

```go
type Plot struct {
    Title string
     X, Y Axis
     Legend Legend
     Style draw.PlotStyle
    plotters []Plotter
}

type Axis struct {
   Min, Max float64
   Label string
   Scale Normalizer
}
```

and define, say, in `vg/draw`:

```go
type PlotStyle {
   Title {
     Padding vg.Length
     draw.TextStyle
   }
   BackgroundColor color.Color
   X, Y AxisStyle
   Legend LegendStyle
}

type AxisStyle struct {
   Label TextStyle
   LineStyle
   Padding vg.Length
}

type LegendStyle struct {
   draw.TextStyle
   Padding vg.Length
   Top, Left bool
}
```

having said that...

to flip it around: imagine the default style of `plot.Plot` was the gnuplot-like one, _ie_ with 4 borders delimiting the canvas, and thus with 4 axes to draw.
then, implementing the current `plot.Plot` style would consist in ignoring 2 of the axes delimiting the canvas. :)

another example: nominal axes. the current code for NominalX actually overrides whatever was inside `Axis.Padding`: https://github.com/gonum/plot/blob/master/plot.go#L396

actually: another possibility is to modify `plot.New() (*Plot, error)` to take an optional `plot.Style` which would have a default `MakeAxis` field function (or method) to be used in place of `plot.makeAxis`. (so the possible inconsistency of `plot.Plot.{X,Y}.Padding` would be solved by default)

so... which option would you be more in favor of?

eaburns: Hm. It's a hard problem. I'll have to think more about it. At least it shouldn't go in the vg package. Vg is totally oblivious to what it's drawing. Yeah, the motivation for writing it was for making plots, but it is its own package that is also intended to be used outside of drawing plots. It shouldn't have details about plot styles.

sbinet: FWIW, for my quick pie-in-the-sky proposal, I put them in the `vg/draw` package because all the other `XyzStyle` (`draw.TextStyle`, `draw.LineStyle`,...) were there :)

sbinet:
ping?

eaburns:
Did we ever come up with a solution for my point 1, above?

plotter/histogram_gnuplotstyle_test.go
sbinet: done (extra nesting removed)

sbinet:
rebased against master (f1e699e)

wrt your point 1 (silently ignoring plot parameters), pragmatically, it is inevitable, I believe, that some plotting style ignore some plot parameters (a style that doesn't display axes at all won't need axes' styles).
I have also laid out a few possible alternatives I dabbled with in comment https://github.com/gonum/plot/pull/245#discussion_r51909697

what do you think about them ?

sbinet:
Today, I tried a bit harder to have a general `theme.Theme` structure (which I called `style.Plot`, together with a `plot.Style` interface)
The beginning of such a work is over there:
 
 https://github.com/gonum/plot/compare/master...sbinet:theme-plot

I had to actually couple each of the "data" structs with the "style" structs (see `plot.Axis.Style` and `plot/style.Axis` for example, or `plot.Plot.Style`).
At least I didn't manage to come up with a nice decoupled way to apply styles (or themes) to plots, axes, legends and ticks...

so... while there was a nice side-effect of grouping the style related informations into "style" structs (potentially: no more `error` needed in the defaultstyle for `plot.New`), I think that #245 is a good enough compromise so people can re-use a good chunk of types and funcs from `gonum/plot` while still being able to customize the `gonum/plot.Plot` final look.

@kortschak @eaburns PTAL.

Can we please decide whether this is deemed useful or not? (if not then I can concentrate on providing that feature inside "my" `plot` package)

sbinet:
ok. I'll integrate this into "my" `plot` package.
thanks for your time.

Issue #252, opened 2016-04-03
plotter: testing is not idempotent with respect to failure


kortschak:
Invoking `go test` on `github.com/gonum/plotter` is not idempotent if there is a failure since the test overwrites the golden values. This should never happen unless the regen flag is passed in.

The ideal solution is to not write anything to a file system however this is required because of the use of example tests for the general testing. I'm even more convinced now that that requirement is too hard.

ctessum:
I think the issue is that because the is a visual library, at least the examples need to be visual, and therefore need to write to the filesystem. It's desirable to have ensure that the examples produce the expected output, so at least some of the testing has to be visual-based too.

In #253 I believe I've made the test idempotent to the likely sources of failures, which may (or may not) be close enough.

PR #253, opened 2016-04-04
make plotter tests idempotent in cases where they run to completion

ctessum:
This is in response to #252. Ideally, the golden images would only be edited in cases where the user requests it. However, in order for the example functions to use p.Save() (which is probably desirable to help people understand the example functions), they need to write directly to the file system every time. If we completely separate the test functions from the example functions, then we no longer have any assurance going forward that the example functions are current and produce the files in the 'testdata' directory.

As a compromise solution, I've changed checkPlot() so that it runs the function as before, which overwrites the golden image, but if the new image is different than the previous golden image, it writes out the previous golden image in it original location and writes out the new image with the "_errored" suffix.

A drawback of this is that if the test's don't run to completion---for example, if the hard drive runs out of space in the middle of the test---the golden images may not be reverted to their original versions. However, this problem would be apparent in a pull request comparison on github, so it is not too subtle.

Other options:

1. Instead of using p.Save() in the example functions, have them write to a buffer, which could be compared to the golden image without overwriting it. I think this would make the examples substantially more confusing. We could also use a "hidden" save function and explain that a user would normally use p.Save(), but I put this in a previous version and it was not popular.
2. Delete the example functions so that there are only test functions in the source code itself, and move the examples to a separate gallery. I'm skeptical of this because then the gallery would require separate maintenance to keep it current with the library, and i've seen cases in other libaries where the example functions don't run or they don't produce the advertised output, which is frustrating.

### DIFF
--- a/plotter/general_test.go
+++ b/plotter/general_test.go
@@ -75,11 +75,11 @@ func checkPlot(ExampleFunc func(), t *testing.T, paths ...string) {
 			errored = true
 		}
 		if errored {
-			// If there has been an error, write out the golden image for inspection.
-			ext := filepath.Ext(path)
-			noext := strings.TrimSuffix(path, ext)
-			goldenPath := noext + "_golden" + ext
-			f, err := os.Create(filepath.Join("testdata", goldenPath))
+			// If there has been an error, write the golden image back to where it
+			// was originally and write the new image to a different file for comparison.
+
+			// Write golden image back to original location.
+			f, err = os.Create(filepath.Join("testdata", path))
 			if err != nil {
 				t.Fatal(err)
 			}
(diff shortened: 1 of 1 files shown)


kortschak:
My personal opinion is that the approach attempts to do too much. This was my concern when the PR that made the change was being reviewed and has not really changed. I don't have a solution. Sorry.

I'll think more about this.

ctessum:
It's true that this solution isn't ideal, and there is probably a better solution, but I think it does at least solve the immediate problem of the mutating golden images, which appears in general to be causing havoc.

kortschak:
OK. LGTM as an interim fix.

Updates #252.

eaburns:
I didn't realize that ExamplePlot always overwrites the golden files even in the case where we aren't generating a new golden set. I think that it would be easier to have ExamplePlot save to a different location. In other words, never overwrite the golden files unless requested.

sbinet:
wouldn't it be less surprising and less magical to have the example create `foo.png` and the test compare it to `foo_golden.png` ? (so rename all the `foo.xyz` files under `testdata` as `foo_golden.xyz`)

regenerating the golden file can then either still be done via `-regen` or just a simple `cp foo.xyz foo_golden.xyz`.

eaburns:
@sbinet, yeah, that is what I was suggesting.

ctessum:
Sure. Is that the consensus?

eaburns:
I think so.

sbinet:
superseded by #282

PR #250, opened 2016-04-01
Liberation fonts

sbinet:
First stab at embedding Liberation fonts.

The embedded asset is generated by running `go generate ./vg/fonts`
`vg/fonts/mk-fonts.go` downloads a version of the liberation fonts from Fedora, untars it somewhere and runs `go-bindata -o liberation_fonts_generated.go`

The increase in size for a given binary (the histogram test) is as follows:

```
run-hist-old  8458416
run-hist-new 10809200
```

perhaps we could recoup some space by having people `import _ "gonum.org/plot/vg/fonts"` using the same kind of mechanism than `database/sql`'s `Driver` ?
In that case, perhaps it would be better to have the Liberation fonts live under `"gonum.org/plot/vg/fonts/liberation"` and put the `fonts` registration mechanism under `"vg/fonts"`
But this would require client's code modification... (a relatively simple one, but still...)

Also, there is a fair amount of test-files updates...
could someone make sure I haven't made any mistake? (the `Liberation` fonts are somewhat different in height.)

Fixes #107

PR #254, opened 2016-04-05
vg: make TestLineWidth more reliable

sbinet:
golang/go@53efe1e121e68fe changed and improved `compress/flate`'s
matching algorithm.
This breaks tests that rely on stable on-disk byte content.

Instead, decode the image.Image from the raw []byte content and compare
the resulting decoding images.

For PDF images, decode the raw []byte using rsc.io/pdf.Reader and
compare each page's content.

### DIFF
--- a/vg/vg_test.go
+++ b/vg/vg_test.go
@@ -8,15 +8,18 @@ import (
 	"bytes"
 	"flag"
 	"fmt"
+	"image"
 	"io/ioutil"
 	"log"
 	"os"
 	"path/filepath"
+	"reflect"
 	"testing"
 
 	"github.com/gonum/plot"
 	"github.com/gonum/plot/plotter"
 	"github.com/gonum/plot/vg"
+	"rsc.io/pdf"
 )
 
 var generateTestData = flag.Bool("regen", false, "Uses the current state to regenerate the test data.")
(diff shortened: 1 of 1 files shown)


sbinet:
tested on go-tip (`go version devel +1f5b1b2 Mon Apr 4 10:05:05 2016 +0000 linux/amd64`)

eaburns:
Sounds fine. But, why not do the simpler byte-equality on the PDFs too?

eaburns:
Sorry, I read this before the comment on #250. Does the PDF change here fix all of the breakage in the PDF tests that you mentioned in #250?

sbinet:
yes. at golang/go@1f5b1b2, `go get ./... && go test ./...` succeeds flawlessly and completely.

eaburns:
LGTM.

vg/vg_test.go
```diff
@@ -65,22 +68,90 @@ func TestLineWidth(t *testing.T) {
...
+				if err != nil {
+					t.Fatalf("failed to retrieve test image infos: %v", err)
+				}
+				want, err := pdf.NewReader(f, fi.Size())
+				if err != nil {
+					t.Fatalf("failed to decode test image (typ=%s): %v", typ, err)
+				}
+
+				r := bytes.NewReader(buf.Bytes())
+				// FIXME: bytes.Reader.Size was introduced only after go-1.4
```
kortschak: Make this a TODO(sbinet) rather than a FIXME.

kortschak:
Nice. LGTM after minor change.

vg/vg_test.go
sbinet: done.

PR #255, opened 2016-04-07
travis: workaround go-get-insecure for rsc.io/pdf

sbinet:
while the `go get rsc.io/pdf` failure is fixed upstream
(rsc/pdf#4), check out manually the repository from github and put it
in the correct place under $GOPATH.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ env:
 
 # Get coverage tools, and start the virtual framebuffer.
 before_install:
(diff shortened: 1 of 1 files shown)


eaburns:
LGTM.

eaburns:
Actually, can you pass -insecure to go get instead?

sbinet:
you could, but `go get -insecure` was introduced some time after `1.4.3`.
I am not sure if `#define` are possible in `.travis.yml` files... :)

eaburns:
OK then this is fine.

PR #256, opened 2016-04-08
vg,plotter: add vg.Canvas.DrawImage and plotter.Image

sbinet:
ImagePlotter embeds an image.Image and plots it via the plot.Plotter
interface.


kortschak:
I'm not a fan of this approach for two reasons.

1. It will create huge svg files for svg users (probably also pdf).

2. It prevents extraction of images from generated pdf files.

The problem is that in many cases the vector graphic formats do allow inclusion of raster images, but this requires additional methods in vg.Canvas. I'm OK with conditional method use, but I don't think @eaburns likes this.

sbinet:
so... add something like this to `vg.Canvas`:

```go
type Canvas interface {
 DrawImage(x1,y1,x2,y2 Length, img image.Image)
}
```

(it would have been great to have the same kind of signature than `image/draw.Drawer` but we can't, b/c of circular imports: `Rectangle` and `Point` live in `vg/draw`)

`svg` has support for such a thing (I would need to check `svgo` does too):
```svg
<image id="image1JPEG" x="240" y="0" width="240" height="150"
 xlink:href="data:image/jpg;base64,/9j/4..."/>
```
`vgpdf` and `vgeps` would be the next "trouble makers"

sbinet:
`vgpdf` should be fine: https://godoc.org/bitbucket.org/zombiezen/gopdf/pdf#Canvas.DrawImage
`vgeps` would need to grow the functionality (based on the c++ code from @eaburns https://github.com/eaburns/search/blob/master/graphics/image.cc#L102)
`vgsvg` is fine as well: `svgo` already provides the functionality: https://godoc.org/github.com/ajstarks/svgo#SVG.Image

alternatively, instead of adding a new method to the `vg.Canvas` interface, we could define `vg/draw.RasterCanvas` with that new extra method.
(and have `ImagePlotter` try to sniff it out and use it or go the pedestrian way if it is not implemented)

eaburns:
Just echoing what @sbinet already said:
I don't think that the plan was ever to fake a raster image with vector graphics; it was always to add a DrawImage method to the Canvas. I think that I checked once, and all of the backends support this except EPS. It shouldn't be too hard to add it to EPS.

As for the signature of the method, I would prefer that we use Points or Rectangles to lots of xs and ys. Did we ever decided whether we are still allowing API breaking changes? In my opinion, it's sad if we can never improve the API. This isn't the standard library of a popular programming language. I think we should allow some kind of path for improving the API — whether we just let things break and hope that folks have vendored, or come up with some kind of versioning scheme.

sbinet:
if we introduce `vg/draw.RasterCanvas` then there's no API breakage and we can use `draw.Rectangle` and `draw.Point`

eaburns:
I guess that's OK. It's just slightly annoying that some canvases may not implement it and give runtime errors.

sbinet:
on the other hand, adding a new method to the `vg.Canvas` isn't *that* big of a deal: according to godoc.org, there is only 33 importers of `gonum/plot/vg` and out of these, only one implements its own `vg.Canvas`:
 https://godoc.org/github.com/go-hep/hplot/vgshiny
(and I have control on that guy :P)

eaburns:
Oh. I was only considering the API breakage from moving Rectangle and Point from vg/draw into vg. I hadn't considered breakage from adding the new method. Now, I'm leaning toward your vg/draw.RasterCanvas idea.

sbinet:
oops, didn't see your comment before now.
I'll introduce `vg/draw.RasterCanvas` later on.

kortschak:
We said we were allowed to break the API during the move, which has unfortunately taken longer than expected, but is happening. I'd vote for a change here.

kortschak:
>I don't think that the plan was ever to fake a raster image with vector graphics; it was always to add a DrawImage method to the Canvas.

That was my understanding, but Plot does this.

sbinet:
so IIUC, putting `DrawImage` into `vg.Canvas` is ok.
the amount of breakage by adding a new method to `vg.Canvas` should be limited.

now, if we were to define `DrawImage` as:

```go
type Canvas interface {
   // draw image img at point p with size r
   DrawImage(pt Point, r Rectangle, img image.Image)
}
```

that would mean pulling `vg/draw.Point` and `vg/draw.Rectangle` out of `vg/draw` and back into `vg`.
the amount of breakage would be more sizeable.
(but the migration would be fairly mechanical.)

I am also a bit unsure about the semantics of `DrawImage`:
 - should it just be given a `(x,y)` point and an `image.Image` and then draw ? (so clients (ie: plotters) of `DrawImage` would clip/re-scale/etc... as needed?)
 - or should it have a signature more like `{,golang.org/x/}image/draw.Draw` instead?

eaburns:
Adding to Canvas seems OK to me. It's unlikely that there are many Canvas implementations in the wild, and we can change all of the ones that we control in one commit.

As for moving Rectangle and Point, I'm not sure what to do. Plotters use these types, and by design there should be many Plotter implementations in the wild. If we don't think folks will want to position images at sub-pixel values then the new method could take image.Point for the destination location. The Rectangle is slicing up a raster image, so that should use image.Rectangle either way.

To answer the question about what interface to provide (either draw.Draw-like or something else), we should do some research and see what is available across all backends. For example, if all backends support scaling, we should probably not require the user to do it themselves.

kortschak:
Why does the method need to take.a Point parameter? Surely only the Rectangle is required since the pt part is implicit in the rect.Min field. This doesn't address the problem of what to do with the types, but it simplifies the signature.

Personally, i thinking moving them to vg, while painful, is probably the right thing to do.

sbinet:
ok, I'll prepare another separate PR with the move of `Point` and `Rectangle` to `vg`.

eaburns:
The point is the location in the dest at which to put the image. The rectangle is the sub-image of the image to use. The point is necessary, arguable the rectangle is not. Again it depends on what the backends support.

eaburns:
I guess draw.Draw does it the reverse. The point is in the source and the rectangle is the dest. Anyway, I don't care much, so whatever you ask think is best.

kortschak:
`image.Image` already defines a rectangle (with int coordinate rather than the float64 coordinate here), so there is no need to add that to this method. What I see is

```
type Canvas interface {
   ...
   // Draw source image scaled to fit the destination rectangle.
   DrawImage(dst Rectangle, src image.Image)
}
```

If you need to crop the image, that can be done by wrapping the src so that that returned bounds are tightened.

eaburns:
Yes.

kortschak:
For the record, the names of parameters here are for explanation, I'd use rect and img in the final product.

sbinet:
rebased on top of #266 

functional `DrawImage` backends:
```
vg/recorder: implement vg.Canvas.DrawImage
vg/vgtex: implement vg.Canvas.DrawImage
vg/vgimg: implement vg.Canvas.DrawImage
vg/vgsvg: implement vg.Canvas.DrawImage
vg/vgpdf: implement vg.Canvas.DrawImage
```

non-functional (yet):
```
vg/vgeps: first stab at implementing vg.Canvas.DrawImage
```

(I would probably need some help on this one. finding specs about the EPS format is... not for the faint of heart.)

also: as `LaTeX` doesn't provide a way to embed image assets (say, like a `base64`-encoded image string) I had to resort to saving the `image.Image` as a `PNG` file and `\pgfimage`-insert it.
that's ugly.

I will probably switch to a somewhat simpler and lighter weight test image for `plotter/image_test.go`...

PTAL.

sbinet:
Needs #261 
Needs #266

vg/recorder/recorder.go
```diff
@@ -381,29 +384,69 @@ type FillString struct {
...
 }
 
 func (a *FillString) callerLocation() *callerLocation {
 	return &a.l
 }
 
+// DrawImage corresponds to the vg.Canvas.DrawImage method
+type DrawImage struct {
+	Rectangle vg.Rectangle
+	Image     []byte
```
kortschak: Is there a particular reason why we are encoding the image rather than just storing it as a image.Image?

vg/recorder/recorder.go
```diff
@@ -381,29 +384,69 @@ type FillString struct {
...
+	img, err := png.Decode(buf)
+	if err != nil {
+		panic(fmt.Errorf("recorder: error decoding image from PNG: %v\n", err))
+	}
+	c.DrawImage(a.Rectangle, img)
+}
+
+// Call returns the pseudo method call that generated the action.
+func (a *DrawImage) Call() string {
+	return fmt.Sprintf("%sDrawImage(%v, %v)", a.l, a.Rectangle, a.Image)
```
kortschak: Here, if we retain the image as an image.Image, either render the call as just the bounds of the image or as that plus the tour/pic format ("IMAGE:"+bas64EncodingOfPNG).

kortschak: Use the %#v verb of the rectangles.

vg/vg.go
```diff
@@ -73,7 +74,10 @@ type Canvas interface {
 
 	// FillString fills in text at the specified
 	// location using the given font.
-	FillString(f Font, x, y Length, text string)
+	FillString(f Font, pt Point, text string)
+
+	// DrawImage draws the image, scaled to fit the destination rectangle.
```
kortschak: Wrap this line a little earlier to be consistent with the other methods - remember that godoc doesn't flow text in a decl block.

kortschak:
#261 and #266 have the go ahead.

kortschak:
Squashing this to meaningful commits is going to be a nightmare if you retain all three PRs.

I would suggest that #261 and #266 be abandoned, do the squashing here into three commits that correspond - in order - to #261, #266 and the DrawImage changes).

vg/recorder/recorder.go
sbinet: nope.
stored as `image.Image`.

vg/recorder/recorder.go
sbinet: done

vg/vg.go
sbinet: done

sbinet:
I have squashed (but, obviously, not pushed yet) #261 and #266 into 1 commit per PR, and rebased this one on top of those 2 commits.

shall I push #261 and #266 to master?
or wait a bit (say till Monday) for people to comment on the gonum-dev thread ?

plotter/image.go
```diff
@@ -0,0 +1,135 @@
...
+type Image struct {
+	img            image.Image
+	cols           int
+	rows           int
+	xmin, xmax, dx float64
+	ymin, ymax, dy float64
+}
+
+// NewImage creates a new image plotter.
+func NewImage(img image.Image, xmin, ymin, xmax, ymax float64) *Image {
```
kortschak: Add commentary for the parameters here.

plotter/image.go
```diff
@@ -0,0 +1,135 @@
...
+	if r >= img.rows || r < 0 {
+		panic("plotter/image: illegal range")
+	}
+	return img.ymin + float64(r)*img.dy
+}
+
+func (img *Image) z(c, r int) color.Color {
+	// image.Image.At(0,0) returns the upper-left pixel
+	// gonum/plot expects 0,0 to be at the bottom-left.
+	v := img.img.At(c, img.rows-1-r)
```
kortschak: Just return this directly.

vg/recorder/recorder.go
```diff
@@ -381,29 +385,66 @@ type FillString struct {
...
+// Call returns the pseudo method call that generated the action.
+func (a *DrawImage) Call() string {
+	var buf bytes.Buffer
+	err := png.Encode(&buf, a.Image)
+	if err != nil {
+		panic(fmt.Errorf("recorder: error encoding image to PNG: %v", err))
+	}
+	b64 := base64.StdEncoding.EncodeToString(buf.Bytes())
+	str := fmt.Sprintf("{%#v, IMAGE:%s}", a.Image.Bounds(), b64)
+	return fmt.Sprintf("%sDrawImage(%#v, %v)", a.l, a.Rectangle, str)
```
kortschak: `return fmt.Sprintf("%sDrawImage(%#v, {%#v, IMAGE:%s})", a.l, a.Rectangle, a.Image.Bounds(), b64)` and delete L440.

kortschak:
LGTM after minor comments addressed and input from @eaburns.

Thank you for doing this. I think waiting until this gets Ethan's OK and then doing them all at once makes sense.

plotter/image.go
sbinet: done

plotter/image.go
sbinet: actually, `plotter.Image.z` is not needed anymore as we don't fallback on the poor man's raster-in-vg approach.
removed.

vg/recorder/recorder.go
sbinet: done

plotter/image.go
```diff
@@ -0,0 +1,130 @@
...
+import (
+	"image"
+	"math"
+
+	"github.com/gonum/plot"
+	"github.com/gonum/plot/vg"
+	"github.com/gonum/plot/vg/draw"
+)
+
+// Image implements the Plotter interface.
```
eaburns:
```
// Image is a plotter that draws a scaled, raster image.
```

plotter/image.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// DataRange implements the DataRange method
+// of the plot.DataRanger interface.
+func (img *Image) DataRange() (xmin, xmax, ymin, ymax float64) {
+	xmax = img.xmax
+	xmin = img.xmin
+	ymax = img.ymax
+	ymin = img.ymin
+	return xmin, xmax, ymin, ymax
```
eaburns: just `return img.xmax, img.xmin, img.ymax, img.ymin`.

plotter/image.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// GlyphBoxes implements the GlyphBoxes method
+// of the plot.GlyphBoxer interface.
+func (img *Image) GlyphBoxes(plt *plot.Plot) []plot.GlyphBox {
```
eaburns: Can't this method just return nil?

GlyphBoxes are used when a glyph extends, in plot space, beyond the data point in data space. For example, a scatter plot point is infinitly small in data space, but the glyph has non-zero size in plot space. It has some radius specified as a vg.Length. So, the point gets a glyphbox in order to make room for the glyph in the case that the infinitely small data point is too close to the edge of the axis bounds.

In this case, the image never extends beyond the bounds of the data-space rectangle in NewImage, so it shouldn't need any padding.

plotter/image_test.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	defer f.Close()
+
+	img, err := png.Decode(f)
+	if err != nil {
+		log.Fatalf("error decoding image file: %v\n", err)
+	}
+
+	p.Add(NewImage(img, 100, 100, 200, 200))
+	p.Add(NewGlyphBoxes())
```
eaburns: No need for this. Drawing the glyphboxes is just for debugging them. If someone wants to display an image, you don't want them copying this line.

vg/vgeps/vgeps.go
```diff
@@ -176,17 +178,106 @@ func (e *Canvas) trace(path vg.Path) {
...
+	c.Translate(rect.Min)
+	fmt.Fprintf(c.buf, "%.*g %.*g scale     %% scale pixels to points\n", pr, width, pr, height)
+	fmt.Fprintf(c.buf, "%d %d 8 [%d 0 0 %d 0 0]     ", iwidth, iheight, iwidth, iheight)
+	fmt.Fprintf(c.buf, "%%%% width height colordepth transform\n")
+	fmt.Fprintf(c.buf, "/datasource currentfile /ASCII85Decode filter /RunLengthDecode filter def\n")
+	fmt.Fprintf(c.buf, "/datastring %d string def   ", iwidth*3)
+	fmt.Fprintf(c.buf, "%% %d = width * color components\n", iwidth*3)
+	fmt.Fprintf(c.buf, "{datasource datastring readstring pop}\n")
+	fmt.Fprintf(c.buf, "false       %%%% false == single data source (rgb)\n")
+	fmt.Fprintf(c.buf, "3   %%%% number of color components\n")
```
eaburns: It's been forever since I've looked at any of this. Do you know if we can do 4 components and add an alpha?

eaburns:
Thanks a lot for doing this. All of these recent changes have been great. Next time, though, please try to hold off sending PRs until their dependent changes are in, because this was very difficult to review.

Assuming that you verified that each backend does indeed plot an image (I didn't try), LGTM.  As for how to submit, if you have a version squashed up into three commits, then we might as well put them all in at once.

sbinet:
@eaburns all backends work *except* `vg/vgeps`.
the `/ASCII85Decode` part was easy (assuming the stuff in the standard library does what the `EPS` driver expects)
but I am unsure the `/RunLengthDecode` part is doing the right thing.

as we are using `EPS-3` do you know whether we could use another compressor (other than RLE)? one that is provided by the stdlib would be best...

eaburns:
Honestly, I don't remember anything about it other than where to find that old code. If the current thing is not working, I think that we either need to fix it before submitting or make it panic("unimplemented") and submit the fix in a separate PR.

sbinet:
I would go with `panic("unimplemented")`.

eaburns:
SGTM, please open a corresponding issue to implement it. If possible, also paste the current broken implementation in the issue.

vg/vgeps/vgeps.go
sbinet: I don't have the `EPS` knowledge to answer this.

plotter/image_test.go
sbinet: right.

plotter/image.go
sbinet: ok. (I used it for debugging and figured it could be useful for users too. but you're right.)

plotter/image.go
sbinet: done

plotter/image.go
sbinet: done

sbinet:
done: #271

sbinet:
also: apologies for the difficult review.

eaburns:
LGTM

kortschak:
It's worth noting what the problem with the eps code is on that issue.

PR #260, opened 2016-04-11
vg: fix Canvas.SetColor godoc string

### LINKED ISSUE #259: vg: fix typos in Canvas.SetColor godoc strings
this was spotted during #258 : 

fix doc on `vg.Canvas.SetColor` to read:
```
// SetColor sets the current drawing color.
// Note that fill color and stroke color are
// the same, so if you want different fill
// and stroke colors then you must set a color,
// draw fills, set a new color and then draw lines.
//
// The initial color is black.  If SetColor is
// called with a nil color then black is used.
SetColor(c color.Color)
```

sbinet:
Fixes #259.

### DIFF
--- a/vg/vg.go
+++ b/vg/vg.go
@@ -33,9 +33,9 @@ type Canvas interface {
 
 	// SetColor sets the current drawing color.
 	// Note that fill color and stroke color are
-	// the same so if you want different fill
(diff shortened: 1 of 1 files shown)


eaburns:
LGTM.

PR #261, opened 2016-04-12
vg: migrate draw.Point+draw.Rectangle back to vg

### LINKED ISSUE #262: vg: migrate draw.Point and draw.Rectangle back to vg
this is needed by #256 

also: we need to adapt some of the wiki examples:
 https://github.com/gonum/plot/wiki/Creating-Custom-Plotters:-A-tutorial-on-creating-custom-Plotters

sbinet:
Fixes #262

### DIFF
--- a/vg/draw/canvas.go
+++ b/vg/draw/canvas.go
@@ -22,7 +22,7 @@ import (
 // to which drawing should take place.
 type Canvas struct {
 	vg.Canvas
-	Rectangle
+	vg.Rectangle
 }
 
 // TextStyle describes what text will look like.
@@ -63,13 +63,13 @@ type GlyphStyle struct {
 type GlyphDrawer interface {
 	// DrawGlyph draws the glyph at the given
 	// point, with the given color and radius.
-	DrawGlyph(*Canvas, GlyphStyle, Point)
+	DrawGlyph(*Canvas, GlyphStyle, vg.Point)
 }
 
 // DrawGlyph draws the given glyph to the draw
 // area.  If the point is not within the Canvas
 // or the sty.Shape is nil then nothing is drawn.
-func (c *Canvas) DrawGlyph(sty GlyphStyle, pt Point) {
+func (c *Canvas) DrawGlyph(sty GlyphStyle, pt vg.Point) {
 	if sty.Shape == nil || !c.Contains(pt) {
 		return
 	}
@@ -79,7 +79,7 @@ func (c *Canvas) DrawGlyph(sty GlyphStyle, pt Point) {
 
 // DrawGlyphNoClip draws the given glyph to the draw
 // area.  If the sty.Shape is nil then nothing is drawn.
-func (c *Canvas) DrawGlyphNoClip(sty GlyphStyle, pt Point) {
+func (c *Canvas) DrawGlyphNoClip(sty GlyphStyle, pt vg.Point) {
 	if sty.Shape == nil {
 		return
 	}
@@ -89,10 +89,10 @@ func (c *Canvas) DrawGlyphNoClip(sty GlyphStyle, pt Point) {
 
 // Rectangle returns the rectangle surrounding this glyph,
 // assuming that it is drawn centered at 0,0
-func (g GlyphStyle) Rectangle() Rectangle {
-	return Rectangle{
-		Min: Point{-g.Radius, -g.Radius},
-		Max: Point{+g.Radius, +g.Radius},
+func (g GlyphStyle) Rectangle() vg.Rectangle {
+	return vg.Rectangle{
+		Min: vg.Point{-g.Radius, -g.Radius},
+		Max: vg.Point{+g.Radius, +g.Radius},
 	}
 }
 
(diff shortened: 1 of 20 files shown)


sbinet:
it's pure code movement.
*except* for some methods of `Point` that I had to export (`minus` -> `Minus`, `dot`, `scale` and `plus`) because they were used by `draw.Canvas` internally (to build `vg.Path`s)

sbinet:
BTW, while we're at it.
perhaps it would be interesting to migrate `vg.Length`, `vg.Point` and `vg.Rectangle` to `x/image/math/fixed.Int52_12` and friends ?
 https://godoc.org/golang.org/x/image/math/fixed

this would probably ensure better compatibility and interoperability with other libraries.

what do you think?

kortschak:
*If* there is a desire to change the dimension type, it should happen in another PR.

sbinet:
agreed.

vg/geom.go
```diff
@@ -0,0 +1,58 @@
...
+	X, Y Length
+}
+
+// Dot returns the dot product of two points.
+func (p Point) Dot(q Point) Length {
+	return p.X*q.X + p.Y*q.Y
+}
+
+// Plus returns the component-wise sum of two points.
+func (p Point) Plus(q Point) Point {
```
kortschak: Since we're exposing these for the first time, how about we choose more commonly used names.

`Add`

vg/geom.go
```diff
@@ -0,0 +1,58 @@
...
+	return Point{p.X + q.X, p.Y + q.Y}
+}
+
+// Minus returns the component-wise difference of two points.
+func (p Point) Minus(q Point) Point {
```
kortschak: `Sub`

kortschak:
LGTM after consideration of names suggestion, waiting for @eaburns.

kortschak:
@jonlawlor I take it these tests aren't being run on travis because of my paranoia?

vg/geom.go
sbinet: done

vg/geom.go
sbinet: done

sbinet:
FTR, I just noticed it as well and went to the travis dashboard for gonum/plot and re-enabled builds on push requests.
Hope that's ok.

vg/draw/canvas.go
```diff
@@ -549,30 +549,30 @@ func clipPoly(in func(Point, Point) bool, clip, norm Point, pts []Point) (clippe
...
 	return p.Y >= clip.Y-slop
 }
 
 // isect returns the intersection of a line p0→p1 with the
 // clipping line specified by the clip point and normal.
-func isect(p0, p1, clip, norm Point) Point {
+func isect(p0, p1, clip, norm vg.Point) vg.Point {
 	// t = (norm · (p0 - clip)) / (norm · (p0 - p1))
-	t := p0.minus(clip).dot(norm) / p0.minus(p1).dot(norm)
+	t := p0.Sub(clip).Dot(norm) / p0.Sub(p1).Dot(norm)
```
eaburns: I prefer naming these nominally, not verbally. They aren't performing an action, they are returning a thing. `Sub` is not subtracting from `p0`. The value of `p0` remains fixed. It's returning the difference of `p0` and `clip`. We call that difference "`p0` minus `clip`", which in code looks like this: `p0.Minus(clip)`.

eaburns: On the other hand, I don't know what to do about Scale.

eaburns: I see that @kortschak told you to change these. I don't want to argue over these names, so just leave them.

eaburns:
LGTM.

vg/draw/canvas.go
sbinet: one could create the following free functions:
```go
func SubPoints(p1,p2 Point) Point {...}
func AddPoints(p1,p2 Point) Point {...}
func DotPoints(p1,p2 Point) Length {...}
func ScalePoint(s Length, p Point) Point {...}
```

eaburns: That makes their use a lot more verbose. These names are fine. They are mostly for internal use anyway.

jonlawlor:
@kortschak I don't know why the commit doesn't seem to have built on travis.  There are other commits and pull requests that are either running right now or ran in the last 20 minutes.  Also checking the settings for the repo and all pushes and pull requests are set to build.

Also I just saw sbinet's commits & tests go through a minute ago, but github is listing them as 5 hours old, maybe there was a big delay between commit and push?

vg/draw/canvas.go
kortschak: The suggestion was based on the naming of time.Time.Sub which returns a time.Time.

sbinet:
ok. 2 LGTMs w/ Sub, Add and friends.

before merging this into `master`, should we send a quick note (at least on `gonum-dev` and perhaps even on `golang-nuts`) about the breakage and the mechanics for the migration ?
or wait for #266 and its associated breakage before merging everything into `master`.

kortschak:
Send a post notifying of impending breakage, but do all the merges together I think.

PR #266, opened 2016-04-12
vg: make Canvas and Path APIs use vg.Points

sbinet:
- improve `vg.Canvas` interface to use `vg.Point` where it makes sense:

```diff
type Canvas interface {
 	// Translate applies a translational transform
 	// to the context.
-	Translate(x, y Length)
+	Translate(pt Point)
 
 	// Scale applies a scaling transform to the
 	// context.
 @@ -73,7 +73,7 @@ type Canvas interface {
 
 	// FillString fills in text at the specified
 	// location using the given font.
-	FillString(f Font, x, y Length, text string)
+	FillString(f Font, pt Point, text string)
}
```

- also improved `vg.Path` and `vg.PathComp` to use `vg.Point`:

```diff
@@ -147,11 +146,11 @@ type PathComp struct {
 	// be meaningless.
 	Type int
 
-	// The X and Y fields are used as the destination
-	// of a MoveComp or LineComp and are the center
-	// point of an ArcComp.  They are not used in
+	// The Pos field is used as the destination
+	// of a MoveComp or LineComp and is the center
+	// point of an ArcComp.  It is not used in
 	// the CloseComp.
-	X, Y Length
+	Pos Point
```

I can split the 2 changes into 2 PRs if deemed preferable.

Needs #261


axis.go
```diff
@@ -248,9 +248,9 @@ func (a *horizontalAxis) GlyphBoxes(*Plot) (boxes []GlyphBox) {
 		w := a.Tick.Label.Width(t.Label)
 		box := GlyphBox{
 			X: a.Norm(t.Value),
-			Rectangle: draw.Rectangle{
-				Min: draw.Point{X: -w / 2},
-				Max: draw.Point{X: w / 2},
+			Rectangle: vg.Rectangle{
```
eaburns: Shouldn't this go into #261? Otherwise, this will be broken when #261 is merged.

plot_test.go
```diff
@@ -57,10 +57,10 @@ func TestLegendAlignment(t *testing.T) {
 		},
 		&recorder.Fill{
 			Path: vg.Path{
-				{Type: vg.MoveComp, X: 80, Y: 30},
-				{Type: vg.LineComp, X: 80, Y: 40},
-				{Type: vg.LineComp, X: 100, Y: 40},
-				{Type: vg.LineComp, X: 100, Y: 30},
+				{Type: vg.MoveComp, Pos: vg.Point{X: 80, Y: 30}},
```
eaburns: We should add `vg.Pt(x, y float64)` so that these don't need to be so verbose.

axis.go
sbinet: it's already part of that PR.
this PR is rebased on top of it, but as #261 hasn't been merged yet, it's shown in the set of CLs.
(perhaps I should have done things differently when creating this PR...)

plot_test.go
sbinet: note that:
```go
{Type: vg.MoveComp, X: 80, Y: 30},
```

doesn't need to be as verbose as:
```go
{Type: vg.MoveComp, Pos: vg.Point{X: 80, Y: 30}},
```

(it's just my script that rewrote it like that)
it could be written as:
```go
{Type: vg.MoveComp, Pos: vg.Point{80, 30}},
```

which isn't that much verbose than:
```go
{Type: vg.MoveComp, Pos: vg.Pt(80, 30)},
```

is it?
I distinctly remember removing that function during the `vg` -> `vg/draw` move :)

eaburns: OK. You've convinced me to leave it as `vg.Point{X: 80, Y: 30}`. It's not so bad, and it's one less function we have to expose.

eaburns:
Would it be possible to use more descriptive names for PRs? When I'm looking at a list of PRs, it's difficult to remember what "vg canvas iface" is.

sbinet:
sure: done.

vg/recorder/recorder.go
```diff
@@ -372,7 +373,7 @@ func (a *Fill) callerLocation() *callerLocation {
 type FillString struct {
 	Font   string
 	Size   vg.Length
-	X, Y   vg.Length
+	X, Y   vg.Length // FIXME: move to a vg.Point
```
eaburns: Why not this one?

axis.go
eaburns: Thanks. Now that I know, I can convince github to only show me the commits that I want.

eaburns:
This LGTM, but it needs to wait for #261.

vg/recorder/recorder.go
sbinet: will do in the next rebase.
(it was quite a large CL already and I was getting a bit tired)

sbinet: done.

axis.go
kortschak: @sbinet be wary of doing that kind of thing, it can get you quite tangled.

sbinet: yep. I don't know whether there is a better workflow for this. do you?

eaburns: That's a pretty standard git workflow.

Next time, I would just not send the follow up PR until the first is in, because github has no way to tell the reviewers what is going on.

sbinet: that's what I usually do but in this instance, it seemed kinder for our users to not break code twice in a row :)

eaburns: Well, it only breaks them twice in a row if they happen to `go get -u` between the time of the two PRs. In the future, maybe we should set a go1 tag on the gitrepo.

PR #267, opened 2016-04-12
vg/vgtex: fix generation of PGF document

sbinet:
- fix a bug introduced by the `New`/`NewDocument` split.
- updated reference file.

this could have been caught if the generation of the PDF was performed during the test but... adding `LaTeX` as a test pre-requisite sounds less than ideal.


eaburns:
LGTM.

PR #269, opened 2016-04-12
plotutil: make legend entries ordered by insertion

### LINKED ISSUE #268: plotutils.AddLines stores names in a map which causes legend ordering issues
https://github.com/gonum/plot/blob/master/plotutil/add.go#L179

When calling plotutils.AddLines the names passed in get stored temporarily in a map which causes the legend ordering to get mixed up. When creating multiple plots it can be important to have the lines/glyphs matched up and in the same order across all plots. If this was a slice instead of a map the order would be preserved.

sbinet:
This CL removes the use of maps to collect legend entries in favor of
slices so the legends are populated by insertion order.
This makes sure the order is stable.

Fixes #268.

### DIFF
--- a/plotutil/add.go
+++ b/plotutil/add.go
@@ -18,6 +18,11 @@ type combineXYs struct{ xs, ys plotter.Valuer }
 func (c combineXYs) Len() int                    { return c.xs.Len() }
 func (c combineXYs) XY(i int) (float64, float64) { return c.xs.Value(i), c.ys.Value(i) }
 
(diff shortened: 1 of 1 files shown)


eaburns:
LGTM. Thanks for doing this.

eaburns:
But I should mention that I didn't try it out, so perhaps wait for @skiesel to given you the OK.

skiesel:
Looks good on my end. Thanks!

PR #270, opened 2016-04-14
vg/vgtex: fix yet another NewDocument-related bufio bug

sbinet:


eaburns:
LGTM. Do we have any to make tests so this doesn't keep happening?

sbinet:
in this particular case, it was mostly a semantics issue but the generated LaTeX file was correct.
(that's why the test was passing.)
it was a bug in waiting though, if some refactoring had happened.

PR #275, opened 2016-04-21
vg/draw: update Canvas.FillText signature for consistency

### LINKED ISSUE #274: draw: Canvas takes an x, y vg.Length pair in FillText
It looks like the vg.Length tuple => vg.Point change was missed here.

kortschak:
Fixes #274.

@sbinet @eaburns Please take a look.

### DIFF
--- a/axis.go
+++ b/axis.go
@@ -205,7 +205,7 @@ func (a *horizontalAxis) draw(c draw.Canvas) {
 	y := c.Min.Y
 	if a.Label.Text != "" {
 		y -= a.Label.Font.Extents().Descent
-		c.FillText(a.Label.TextStyle, c.Center().X, y, -0.5, 0, a.Label.Text)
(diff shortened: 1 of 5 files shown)


sbinet:
LGTM.

it would seem, there are 2 other occurrences of a pair `x,y vg.Length` which could be replaced by a `pt vg.Point`:
- https://github.com/gonum/plot/blob/master/legend.go#L37
- https://github.com/gonum/plot/blob/master/plotter/labels.go#L44
 
not sure whether these should be modified nor if they should be part of that CL.

kortschak:
Not all x,y tuples are points. I'll wait for @eaburns.

eaburns:
This change LGTM.
I think that it's fine to replace the offsets with points too.

kortschak:
PTAL

sbinet:
still LGTM

sbinet:
(don't forget to rebase on top of the latest master)

PR #278, opened 2016-04-22
plotter: use a fixed time zone for tests

### LINKED ISSUE #276: plotter: UnixTimeTicks is not time.Location agnostic
I ran tests today and found that the dates shown on the time series test were off by one. I suspect this is due to timezone since I am running this in Australia.

/cc @sbinet

sbinet:
Fixes #276


kortschak:
It's been [132 years](https://en.wikipedia.org/wiki/International_Meridian_Conference). I think it's time to let go ;P

This does not fix the problem; UTC is just as fixed as Paris time. I think the issue is interplay with the locale that is used by formatting and the locale that is used during parsing, though I can't find a combinations of locales that fixes it.

eaburns:
You need to use the local timezone to create the original time values. The timezone is lost when you convert to unix seconds`; `time.Unix` always assumes its input is in the local timezone. If you create the initial times in the local zone then `time.Unix` will be correct in its assumption that the times are local, and you will get the correct time back.

kortschak:
I tried that when I reported the issue and it doesn't work:

```
$ git diff
diff --git a/plotter/timeseries_test.go b/plotter/timeseries_test.go
index 9954874..dbbfcf6 100644
--- a/plotter/timeseries_test.go
+++ b/plotter/timeseries_test.go
@@ -31,7 +31,7 @@ func Example_timeSeries() {
                )
                pts := make(XYs, n)
                for i := range pts {
-                       date := time.Date(2007+i, month, day, hour, min, sec, nsec, time.UTC).Unix()
+                       date := time.Date(2007+i, month, day, hour, min, sec, nsec, time.Local).Unix()
                        pts[i].X = float64(date)
                        pts[i].Y = float64(pts[i].X+10*rand.Float64()) * 1e-9
                }
$ go test
--- FAIL: TestImagePlot (0.02s)
	general_test.go:74: image mismatch for image_plot.png
	general_test.go:94: golden image has been written out as image_plot_golden.png
--- FAIL: TestTimeSeries (0.02s)
	general_test.go:74: image mismatch for timeseries.png
	general_test.go:94: golden image has been written out as timeseries_golden.png
FAIL
```

There is another failure there, which I have not yet investigated.

eaburns:
Are those two separate thoughts: 1. that you tried it, and 2. that you
found another failure? Or is the second a consequence of the first?

Anyway, I'm still confused. How can it fail if you use local time? If you
create a date in your local timezone, convert it to unix seconds, convert
it back is it a different time? In other words, what does this program
print when you run it locally: http://play.golang.org/p/3AjCyKJsIX ? It
should print true. If so, then I don't see how doing the exact same thing
for the test would fail.

sbinet:
so... the fix would actually be to modify:
https://github.com/gonum/plot/blob/master/axis.go#L483

to return a `time.Time` in `UTC`, right?
```go
t := time.Unix(int64(tick.Value), 0).UTC()
```

then, perhaps the right fix would be to add another field to `UnixTimeTicks`:

```go
type UnixTimeTicks struct {
	// Ticker is used to generate a set of ticks.
	// If nil, DefaultTicks will be used.
	Ticker Ticker

	// Format is the textual representation of the time value.
	// If empty, time.RFC3339 will be used
	Format string

	// Convert takes a float64 value and converts it into a time.Time.
	// If nil, time.Unix will be used.
	Convert func(float64) time.Time
}
```

kortschak:
They are separate things. I didn't look into the second failure because the test golden image mutation gets in the way of debugging the issue.

eaburns:
@sbinet, another option is instead of having Convert, have the user specify a timezone in the ticker, defaulting to Local. However, I still want to know why using Local all of the way through doesn't work.

kostya-sh:
I have looked into this issue and found the following:
- In order to produce a plot with the same date ticks as the golden image the same timestamps and location as were used to produce the golden image must be used. This means that timestamps should be created using UTC timezone but formatted using  (I assume) Europe/Paris.
- I also discovered that when I run tests locally (using the latest versions of all libraries and Go 1.7 from tip) all tests comparing images against the golden files fail. Visually test images look the same though.

I hope this helps.

As far as the code change is concerned I would prefer to have UnixTimeTicks.Location field instead of UnixTimeTicks.Convert function (as suggested by @eaburns).

kostya-sh:
With Go 1.6.2 all tests pass (apart from timestamp test as I am in UTC+8). I suspect that Go 1.7 produces different images comparing to Go 1.6 due to compression algorithms changes in `compress/flate` package.

sbinet:
@kostya-sh thanks for giving it a try (and raising the issue _wrt_ `tip`.)

_wrt_ having a `Convert` field or a `Location` field:
having a `Convert` func field allows the user to have complete control on how the 64bits of a `time.Time` are to be encoded and then interpreted. (perhaps you want `UnixNano()` instead of `Unix()`, perhaps you want `UTC()`, _etc..._)
You could imagine that instead of simply converting the returned `int64` value into a `float64` one, you may have a function that retrieves the `int64` bits payload and encodes it directly (via `math.Float64frombits`) into a `int64`. No rounding issue for big `int64` values.

Just for my own information, did my alternate PR (the #278 one with `Convert`) fix the `UTC` problem?

kostya-sh:
@sbinet, 

`Convert` vs `Location`. There is only one way to convert unix time (number of seconds since epoch) and timezone (`time.Location`) to `time.Time` object. Any other conversion is no longer a unix time conversion. If more flexibility is required then maybe UnixTimeTicks should be renamed to TimeTicks and a few standard conversion functions are provided. E.g.:

    
    type TimeTicks {
        ...
        Conversion: TimeConversion
    }

    type TimeConversion func(float64) time.Time

    func UnixTime(loc *time.Location) TimeConversion {
        ....
    }
?

Your PR with UTC fixes the problem, but if you use Local() both to create timestamps and to convert them then the test fails because the golden file has been created using different data. To fix it you can either:

- regenerate the golden file (using Local() conversion)
- Use the timezone that was used to create golden file in your example (`return time.Unix(int64(v), 0).In(time.FixedZone("Europe/Paris", 120))`)

kortschak:
>If more flexibility is required then maybe UnixTimeTicks should be renamed to TimeTicks and a few standard conversion functions are provided.

This sounds good to me.

sbinet:
PTAL.

sbinet:
interestingly, my first approach of (un)marshaling `time.Time` values into `float64` via `math.Float64frombits` didn't work because (presumably, haven't checked) the `Ticker` interface expects a linear space to find ticks and adjacent `time.Time` values don't result in adjacent `float64` values (generated via `math.Float64frombits`)

too bad.

kostya-sh:
@sbinet I think converting from time.Time to float64 (`TimeFloatConverter.MarshalTime) should not be part of this package because:
1. this functionality is not used in the pacakge itself
2. it is cumbersome to use this functionality from outside of the package (`t.Unix()` is much simler than creating `plot.UnixTimeConverter{}` and calling `converter.MarshalTime(t)`). Ideally I want the time formatting to work reasonably just by setting `Tick.Marker = plot.TimeTicks{}`.

Secondly it is not possible to customize Location that is used to format dates. Currenly it is hardcoded to UTC.

See also some comments inline.

axis.go
```diff
@@ -451,41 +451,70 @@ func (ts ConstantTicks) Ticks(float64, float64) []Tick {
 	return ts
 }
 
-// UnixTimeTicks is suitable for axes representing time values.
-// UnixTimeTicks expects values in Unix time seconds.
-type UnixTimeTicks struct {
+// TimeTicks is suitable for axes representing time values.
+// By default, TimeTicks expects values in (UTC) Unix time seconds.
```
kostya-sh: Unix time represents moment in time. It is time zone independent, so _(UTC) Unix time seconds_ doesn't make sense.

axis.go
```diff
@@ -451,41 +451,70 @@ func (ts ConstantTicks) Ticks(float64, float64) []Tick {
...
+	UnmarshalTime(float64) time.Time
+}
+
+// UnixTimeConverter converts a time.Time value into the float64 representation
+// of Unix time, using the UTC time location.
+type UnixTimeConverter struct{}
+
+// MarshalTime implements TimeFloatConverter.
+func (UnixTimeConverter) MarshalTime(t time.Time) float64 {
+	v := t.UTC().Unix()
```
kostya-sh: calling UTC() here is not necessary `t.Unix()` is always the same as `t.UTC().Unix()`. From https://golang.org/pkg/time/#Time:

> Each Time has associated with it a Location, consulted when computing the presentation form of the time, such as in the Format, Hour, and Year methods. The methods Local, UTC, and In return a Time with a specific location. Changing the location in this way changes only the presentation; it does not change the instant in time being denoted and therefore does not affect the computations described in earlier paragraphs.

axis.go
```diff
@@ -451,41 +451,70 @@ func (ts ConstantTicks) Ticks(float64, float64) []Tick {
...
+	return float64(v)
+}
+
+// UnmarshalTime implements TimeFloatConverter.
+func (UnixTimeConverter) UnmarshalTime(v float64) time.Time {
+	return time.Unix(int64(v), 0).UTC()
```
kostya-sh: Using UTC as default here is fine, but it would be good to make it configurable.

axis.go
```diff
@@ -451,41 +451,70 @@ func (ts ConstantTicks) Ticks(float64, float64) []Tick {
...
-		tick.Label = t.Format(utt.Format)
+		t := tt.Converter.UnmarshalTime(tick.Value)
+		tick.Label = t.Format(tt.Format)
 	}
 	return ticks
 }
 
+// TimeFloatConverter converts back and forth a time.Time value into
+// a float64.
+type TimeFloatConverter interface {
```
kostya-sh: Instead of using interface I would use 

    type TimeConverterFunc func(t float64) time.Time

and have

    func UnixTimeIn(loc *time.Location) TimeConverterFunc {
        return func(t float64) time.Time {
            return time.Unix(int64(t), 0).In(loc)
        }
    }

    var UTCUnixTime = UnixTimeIn(time.UTC) // use this as default converter

kostya-sh:
I left a few comments however despite being the one who proposed the idea of having more flexible TimeTicks type I really like the simplicity of the original UnixTimeTicks type and still think that adding Location field to it was all that required ;)

sbinet:
@kostya-sh , yes, it is somewhat a bit less compact to write.
on the plus side, the protocol to encode/decode `time.Time` values to/from `float64` ones is, IMHO, clearer, with less opportunities to introduce impedance mismatch or incorrect expectations.

I could introduce:

```go
func UnixTimeTicks(format string, cnv TimeFloatConverter) TimeTicks {
    if format == "" {
        format = time.RFC3339
    }
    if cnv == nil {
        cnv = UnixTimeConverter{}
    }
    return TimeTicks{
        Format: format,
        Converter: cnv,
    }
}
```

kostya-sh:
I had another look at ticker related functionality in plot package and I think it might be good idea to split Ticker interface into two: Ticker and Labeler. 

    type Ticker interface {
      func Ticks(min, max float64) []float64
    }

    type Labeler interface {
      func Label(tick float64) string
    }

This will allow to specify ticker rules and formatting rules independently. 

For example LogTicks and DefaultTicks and ConstantTicks use the same formatting rules but different ticker rules. UnixTimeTicks and DefaultTicks have the same ticker rules but different formatting rules.

Also if ConstantTicks is just a []float64 it is much easier to create in code.

It might be good idea in addition to the time specific formatter provide time specific ticker as well (that can prefer ticks at the beginning of major time intervals such as minutes or months depending on range).

Having a seprate Labeler interface will also make it easy to provide FloatLabeler with configurable precision if necessary.

Thoughts? Should I bring this question on the mainling list instead?

sbinet:
SGTM.
I suppose sending a PR wouldn't hurt to see how that would look like.
(small(er) interfaces are good, anyways)

kostya-sh:
Ok, I will send a PR for review soon.

kostya-sh:
See #286

kortschak:
Can we merge this in the interim. Having a failing case here is hampering what I can do.

kortschak:
*ping*

kortschak:
@sbinet Is there a particular hold up here that I can help with?

kortschak:
I'm going to close this. If it is going to be revived, please reopen.

eaburns:
I don't understand that logic. It certainly will not get reviewed if it's closed. Are you waiting for me or someone in particular to review?

kortschak:
It's stale and hasn't moved for months. If I get some time I will try to put together a good.

PR #282, opened 2016-05-25
plotter: make plotter tests idempotent

### LINKED ISSUE #252: plotter: testing is not idempotent with respect to failure

kostya-sh:
This change renames all golden files from plotter/testdata to have
"_golden.png" suffix so running tests do not override them.

Fixes #252


ctessum:
This is more or less what I was planning on doing when I got around to finishing #253. LGTM, thanks.

kostya-sh:
@ctessum. sorry for stealing your PR ;)

ctessum:
No worries! It was making me nervous just sitting there unfinished.

plotter/general_test.go
```diff
@@ -21,30 +21,40 @@ import (
 
 var generateTestData = flag.Bool("regen", false, "Uses the current state to regenerate the test data.")
 
+func goldenPath(path string) string {
+	ext := filepath.Ext(path)
+	noext := strings.TrimSuffix(path, ext)
+	return noext + "_golden" + ext
+}
+
+func readAll(path string) ([]byte, error) {
```
sbinet: AFAICT, this is just `io/ioutil.ReadFile`, isn't it?

.gitignore
```diff
@@ -1,5 +1,7 @@
 **/*.svg
 **/*.png
+!**/testdata/*_golden.png
```
sbinet: do we want to ignore `_golden` files ?
I'd say we do want to see via `git status` that a `_golden` file was modified.
and/or see when a new `_golden` file appeared.

plotter/general_test.go
```diff
@@ -53,45 +63,20 @@ func checkPlot(ExampleFunc func(), t *testing.T, paths ...string) {
...
-	for i, path := range paths {
-		errored := false
-		f, err := os.Open(filepath.Join("testdata", path))
-		if err != nil {
-			t.Error(err)
-			errored = true
-		}
-		have, err := ioutil.ReadAll(f)
+	for _, path := range paths {
+		got, err := readAll(path)
```
sbinet: just use `io/ioutil.ReadFile`

plotter/general_test.go
```diff
@@ -53,45 +63,20 @@ func checkPlot(ExampleFunc func(), t *testing.T, paths ...string) {
...
 		if err != nil {
-			t.Error(err)
-			errored = true
+			t.Errorf("Failed to read %s: %s", path, err)
+			continue
 		}
-		err = f.Close()
+		golden := goldenPath(path)
+		want, err := readAll(golden)
```
sbinet: ditto

sbinet:
is this PR completely superseding #253 ?

.gitignore
kostya-sh: `!` at the beginning of the line tells git to not ignore golden files. Before this change all *.png files were ignored.

plotter/general_test.go
kostya-sh: Good point!

ctessum:
Yes.

kostya-sh:
@sbinet , thanks for the review. I've addressed your comments.

.gitignore
sbinet: ah! sorry, I overlooked that exclamation mark...

sbinet:
LGTM.

sbinet:
@kostya-sh could you send a PR against `gonum/license` so I can merge this PR?
(I don't think you signed it, did you?)

kostya-sh:
@sbinet sent PR against license

sbinet:
needs gonum/license#19

sbinet:
ok, thanks.
I'll wait a day or so for @kortschak to comment (he already LGTM'd #253, so it should be ok.)
@eaburns was also ok with the previous PR...

sbinet:
merged. thanks again.

PR #286, opened 2016-06-01
plot: split Ticker into Ticker and Labeler

kostya-sh:
DO NOT MERGE.

This change is for review and discussion only.

This is backward incompatible change.

### DIFF
--- a/axis.go
+++ b/axis.go
@@ -24,6 +24,11 @@ type Ticker interface {
 	Ticks(min, max float64) []Tick
 }
 
+// Labeler creates lables from Ticks
+type Labeler interface {
+	Labels(ticks []Tick) []string
+}
+
 // Normalizer rescales values from the data coordinate system to the
 // normalized coordinate system.
 type Normalizer interface {
@@ -67,10 +72,14 @@ type Axis struct {
 		// tick marks.
 		Length vg.Length
 
-		// Marker returns the tick marks.  Any tick marks
+		// Ticker returns the tick marks.  Any tick marks
 		// returned by the Marker function that are not in
 		// range of the axis are not drawn.
-		Marker Ticker
+		Ticker
+
+		// Labeler is used to create tick labels from ticks created by
+		// Ticker
+		Labeler
 	}
 
 	// Scale transforms a value given in the data coordinate system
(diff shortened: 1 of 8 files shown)


kortschak:
I like this.

axis.go
```diff
@@ -451,76 +468,100 @@ func (ts ConstantTicks) Ticks(float64, float64) []Tick {
...
 }
 
-var _ Ticker = UnixTimeTicks{}
+var _ Labeler = UnixTimeLabeler{}
 
 // Ticks implements plot.Ticker.
-func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
-	if utt.Ticker == nil {
-		utt.Ticker = DefaultTicks{}
+func (utt UnixTimeLabeler) Labels(ticks []Tick) []string {
```
sbinet: perhaps `s/utt/utl/g` ?

axis.go
```diff
@@ -24,6 +24,11 @@ type Ticker interface {
 	Ticks(min, max float64) []Tick
 }
 
+// Labeler creates lables from Ticks
```
sbinet: `s/lables/labels/`

sbinet:
looks good.
I may even abandon #278 in favor of this change.

kostya-sh:
If the agreement is that the new API is an improvement over the current API and will be merged I am happy to finsh this PR (it requires some cleanup, more docs and tests).

I also found previous discussion about tick generation (https://groups.google.com/forum/#!topic/gonum-dev/g7ZgAVItbbA) and https://github.com/vdobler/plot-1/blob/redesign/design.md by @vdobler. There are a few proposed features. Maybe it is worth checking how to implement them (and if it is possible at all) with the new API?

sbinet:
(apologies for the belated answer)

yes, I guess it would be worth to get a sense of how this might look like.
(I myself have still to update my PR #245 and/or provide a more theme-based oriented facility, like https://godoc.org/golang.org/x/exp/shiny/widget/theme#Theme)

eaburns:
Any news on this? We are trying to clean up our PRs. Folks seemed to like this idea, but it still has some outstanding comments. Do you plan to address them? If not, we plan to close this soon.

kostya-sh:
I am going to close this PR. I am still not sure if the proposed interface is flexible enough. Additionally I do not have much time to work on this at the moment.

PR #290, opened 2016-07-09
plotter: fix XY range and palette indexing calculations

jasonpfox:
First image is prior broken code, only data for X>0, Y>0 shown, 2nd includes full X and Y value ranges, final image shifts palette index to avoid early saturation seen in top right of images.

prior-onlyposx-posyshown
fixed-fullmap
scalefix-noearlysaturationtopright

Issue #289, opened 2016-07-08
Heatmap xmin and ymin incorrect for negative minimum values


jasonpfox:
Given Y(0) == -19 and Y(1) == -18, ymin comes out as -0.5, where it needs to be -19.5.  Looks like scale for xmin and ymin only worked correctly for min endpoint at 0.

jasonpfox:
I am trying to push up a fix in a branch but it looks like I don't have permission to gonum/plot...

sbinet:
yep, you don't.
the normal workflow on github is to fork, fix/enhance your fork and then send a pull request (a PR) against the original project.

sbinet:
see how I did it there: https://github.com/sbinet/gonum-plot/tree/time-loc
(ie a branch `time-loc` on my fork of `gonum/plot` in `sbinet/gonum-plot`)

which a PR'ed here: `https://github.com/gonum/plot/pull/278`

having said that, could you show us what you tried, what `git` told you and what's the output of `git remote -v` ?

jasonpfox:
I was to push up a new branch with the change via SourceTree:

git -c diff.mnemonicprefix=false -c core.quotepath=false push -v --tags --set-upstream origin 289-HeatmapXminYmin:289-HeatmapXminYmin
remote: Permission to gonum/plot.git denied to jasonpfox.

fatal: unable to access 'https://github.com/gonum/plot/': The requested URL returned error: 403
Pushing to https://github.com/gonum/plot

kortschak:
Use github to make the fork. I can see that you don't have a fork of gonum/plot in your repos there. Then pull to your local from that repo, make changes, push back to your github repo, then send the PR.

kortschak:
Would you please send a reproducer. The description here is not complete enough for me to write a fix.

kortschak:
Note that Contour uses the same logic for the domain calculations, so any fix should probably be applied there too.

jasonpfox:
I submitted a pull request with a fix.  Also looked briefly at contour.  The first fix is not applicable, but the second (line 172 in contour.go change the +0.5 to -0.5 in finding index into pal) I believe would apply.  I did not include that in submission since I was not setup to create or test that code.  This is also likely not as noticeable or problematic with contour lines as it is with heatmaps

kortschak:
I'll send a PR for the contour palette fix. Thanks.

kortschak:
[...]

PR #291, opened 2016-07-09
palette: avoid floating point errors with dense palettes

kortschak:
When a user specifies a very dense palette (see #290 for an example), floating point error can accumulate resulting in unexpected color.Color values (zero in the case above). Rather than warning users to avoid overly dense palettes, just avoid the error by using multiplication rather than successive addition.

@sbinet @jasonpfox Please take a look.


kortschak:
Can we do something about the test failures for image matching and time handling: timeseries is failing here (as expected), but also image_plot is failing. These are not due to this change as far as I can see (though travis is being extremely unhelpful in not actually letting me see the build log, so I may be wrong - even in the case I am, I can't regen the golden images because of the time series issue).

btracey:
LGTM

kortschak:
Updates #289.

PR #292, opened 2016-07-09
add myself to authors

ctessum:


kortschak:
I don't think this is necessary - I'm not sure why AUTHORS is still in this repo, but I pretty sure it's largely historical/inertial.

PR #295, opened 2016-07-12
plotter: fix data range calculation for negative X and Y

### LINKED ISSUE #289: Heatmap xmin and ymin incorrect for negative minimum values

kortschak:
Fixes #289.

@jasonpfox @eaburns @sbinet Please take a look.


plotter/heat.go
```diff
@@ -178,16 +178,24 @@ func (h *HeatMap) DataRange() (xmin, xmax, ymin, ymax float64) {
 		xmax = 0.5
 		xmin = -0.5
 	default:
+		// Min and max are the first and last h.GridXYZ.X
+		// values extended beyond by half the distance
+		// between them and their adjacent h.GridXYZ.X
+		// value.
 		xmax = (3*h.GridXYZ.X(c-1) - h.GridXYZ.X(c-2)) / 2
```
vladimir-ch: Make the code do more explicitly what is described in the comment? Then perhaps the comment is not even necessary.
```
xmax = h.GridXYZ.X(c-1) + (h.GridXYZ.X(c-1) - h.GridXYZ.X(c-2)) / 2
xmin = h.GridXYZ.X(0) - (h.GridXYZ.X(1) - h.GridXYZ.X(0)) / 2
```

kortschak: Yeah. It's sort of habit for me to reduce the number of interface method calls. This doesn't get called in a loop, so probably it should be expanded.

kortschak: Done

jasonpfox:
Sorry I hadn't gotten back to finish this off, thanks for working it.  I have been running into other strangeness that I wanted tor ack down and include if needed.
I think we still need to change the +0.5 to -0.5 on line 162 for bin retrieval

kortschak:
>I think we still need to change the +0.5 to -0.5 on line 162 for bin retrieval

The black bin for you issue is fixed by #291.

The 0.5 addition is for correct rounding - using a subtraction will cause a panic for the minimum Z value.

jasonpfox:
When I looked, starting with minimum value (0) -0.5 would round up to 0 and start at the 0th bin correctly,  With +0.5, you have a minimum start point of 0+0.5 and you always round up from there to bin 1 and never use bin 0 color.  Not a big deal either way, the panic I found was caused by another case...  Trying to get the process correct for a few additional bugs, hopefully you won't have to come in and rework it for me again... :)

kortschak:
Try playing with pal, v, min and max [here](https://play.golang.org/p/YZwwQLcli7). You'll see that the plus case always uses all the bins, the minus case does not.

btracey:
LGTM

PR #298, opened 2016-07-15
Allow solid colored glyphs have a different colored outline

zaddok:
Here is an initial go at a patch that allows solid coloured Glyphs to have an outline.

A keen observer may note that an update to allow a `PyramidGlyph` to be solid coloured with an outline, is almost the same as if we had allowed `TriangleGlyph` to have a fill. (ditto for Box/Square, and Circle,Round).

There is probably a case for arguing that we don't need `PyramidGlyph` and `TriangleGlyph` as they are both the same except with a different fill or stroke colour, but I don't want such arguments to hold up a useful patch :D Can we patch first and argue how to present the API later?

Where in the past we may have done:

         p, err := plot.New()
         ....
         lpPoints.Shape = draw.CircleGlyph{}
         lpPoints.Color = color.RGBA{R: 255, G: 200, B: 200, A: 255}
         p.Add(lpLine, lpPoints)

We can now do:

         p, err := plot.New()
         ....
         lpPoints.Shape = draw.CircleGlyph{}
         lpPoints.Color = color.RGBA{R: 255, G: 200, B: 200, A: 255}
         lpPoints.Width = vg.Points(1.0)
         lpPoints.LineColor = lpLine.Color
         p.Add(lpLine, lpPoints)

### DIFF
--- a/vg/draw/canvas.go
+++ b/vg/draw/canvas.go
@@ -55,6 +55,12 @@ type GlyphStyle struct {
 	// Radius specifies the size of the glyph's radius.
 	Radius vg.Length
 
+	// Width is set, LineColour is used to outline the glyph
+	Width vg.Length
+
+	// Outline glyph with this color if Width is set.
+	LineColor color.Color
+
 	// Shape draws the shape of the glyph.
 	Shape GlyphDrawer
 }
@@ -106,6 +112,11 @@ func (CircleGlyph) DrawGlyph(c *Canvas, sty GlyphStyle, pt vg.Point) {
 	p.Arc(pt, sty.Radius, 0, 2*math.Pi)
 	p.Close()
 	c.Fill(p)
+
+	if sty.Width > 0 {
+		c.SetLineStyle(LineStyle{Color: sty.LineColor, Width: sty.Width})
+		c.Stroke(p)
+	}
 }
 
 // RingGlyph is a glyph that draws the outline of a circle.
@@ -156,6 +167,11 @@ func (BoxGlyph) DrawGlyph(c *Canvas, sty GlyphStyle, pt vg.Point) {
 	p.Line(vg.Point{X: pt.X - x, Y: pt.Y + x})
 	p.Close()
 	c.Fill(p)
+
+	if sty.Width > 0 {
+		c.SetLineStyle(LineStyle{Color: sty.LineColor, Width: sty.Width})
+		c.Stroke(p)
+	}
 }
 
 // TriangleGlyph is a glyph that draws the outline of a triangle.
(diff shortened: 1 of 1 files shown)


vg/draw/canvas.go
```diff
@@ -55,6 +55,12 @@ type GlyphStyle struct {
 	// Radius specifies the size of the glyph's radius.
 	Radius vg.Length
 
+	// Width is set, LineColour is used to outline the glyph
```
kortschak: How about using a `draw.LineStyle` here? Comment would be:

```
// LineStyle is the style of the glyph outline.
// If LineStyle.Width is zero no outline is drawn.
```

zaddok: Great suggestion, I would like that idea, except for the fact that `LineStyle` also has `Dashes` and `DashOffs` fields, they would be effectively unused. Im happy to switch the patch over to use `LineStyle` if that is what people prefer.

kortschak:
I'm not sure about this. I'm inclined to think the place for this kind of configuration is in the GlyphDrawer implementations rather than in a style.

vg/draw/canvas.go
kortschak: There's no reason not to allow dashed outlines for glyphs.

zaddok:
Shouldmt the object that holds the full color also hold the line/stroke color? 

Perhaps I misunderstand? Are you saying the fill color should go in the GlyphStyle object and the line/stroke color should go in the GlyphDrawer object?

kortschak:
What I'm saying is that you could write an `OutlineCircle` implementation of `GlyphDrawer` that has all the relevant style information in the type.

zaddok:
Aaah, ok so your suggesting we create either:

1. `OutlineCircleGlyph`, `OutlineBoxGlyph` and a `OutlinePyramidGlyph` or
2. `FilledRoundGlyph`, `FilledSquareGlyph` and a `FilledTriangleGlyph`

That I could also do, that would be more following the spirit of the existing code. Would that get the pull request approved

If we go down that route, someone could argue that `OutlineCircleGlyph`, `CircleGlyph` and `RoundGlyph` are technically all just circles with different style attributes.

kortschak:
The argument could be made that there is an obvious redundancy if the glyphs are configurable. The issue is that they just work. Functions of similar names that return appropriately configured GlyphDrawers would probably be satisfactory though.

All up, this is more complicated than it first looks. I'd like to wait until @eaburns has a chance to comment.

zaddok:
In summary the choice is between

1. existing patch, Make `CircleGlyph`, `BoxGlyph`, `PyramidGlyph` configurable
1. Add new types of glyphs for each shape, i.e. `OutlineCircleGlyph`, `CircleGlyph`, and `RoundGlyph`
2. Merge `CircleGlyph` and `RoundGlyph` into a single object where you can specify both stroke and fill.

eaburns:
I haven't had a chance to look at the code yet, but from what I've read here, I prefer option 3.

zaddok:
Great. In the morning I'll fgo ahead and create another patch that it implements option three.  

Do we want to merge CircleGlyph into RoundGlyph, or visa versa? (Same goes for square/box and Triangle/Pyramid)

eaburns:
Any news on this? We are trying to clean up our PRs. It seems like a nice simple change, but still has some outstanding comments. Do you plan to address them? If not, we plan to close this soon.

eaburns:

PR #299, opened 2016-07-18
Plotter: Fix 0 depth heatmap issue

### LINKED ISSUE #302: Creating a heatmap with 0 depth (min value == max value) causes panic

jasonpfox:
Resolves #302 
If a data set happens to have 0 depth (min==max) the color scaling calculations blow up to NaN due to a divide by 0.  Catch this case to allow successfully returning a flat heatmap.

### DIFF
--- a/plotter/heat.go
+++ b/plotter/heat.go
@@ -179,7 +179,7 @@ func (h *HeatMap) DataRange() (xmin, xmax, ymin, ymax float64) {
 		xmin = -0.5
 	default:
 		xmax = (3*h.GridXYZ.X(c-1) - h.GridXYZ.X(c-2)) / 2
-		xmin = (h.GridXYZ.X(0) - h.GridXYZ.X(1)) / 2
+		xmin = (3*h.GridXYZ.X(0) - h.GridXYZ.X(1)) / 2
 	}
 	switch r {
 	case 1: // Make a unit length when there is no neighbour.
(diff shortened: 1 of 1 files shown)


plotter/heat.go
```diff
@@ -159,7 +164,7 @@ func (h *HeatMap) Plot(c draw.Canvas, plt *plot.Plot) {
 			case v > h.Max:
 				col = h.Overflow
 			default:
-				col = pal[int((v-h.Min)*ps+0.5)] // Apply palette scaling.
+				col = pal[int((v-h.Min)*ps-0.5)] // Apply palette scaling.
```
kortschak: This is still incorrect. Please leave it.

plotter/heat.go
```diff
@@ -98,6 +98,11 @@ func (h *HeatMap) Plot(c draw.Canvas, plt *plot.Plot) {
 	}
 	// ps scales the palette uniformly across the data range.
 	ps := float64(len(pal)-1) / (h.Max - h.Min)
+	// for some degenerate data sets with h.Max == h.Min, this blows
+	// up and needs to be caught
+	if math.IsNaN(ps) {
```
kortschak: If Min == Max, the plot is meaningless. I would argue the best thing to do here is panic. Maybe this should be more explicit, but this is not the correct answer.

jasonpfox: A flat heatmap itself isn't meaningful by itself, but as one of a set of heatmaps that show differences in data it shows the case of perfectly flat data and is a valid heatmap to compare to others.  Additionally, updating a heatmap continuously based on data flowing through can see cases where all data values are identically equal which is still a valid and well-formed case so i wouldn't expect this to cause a panic

kortschak: `h.Min` and `h.Max` can be set prior to plotting. If a selection of heatmaps are being rendered, they should be on the same scale and so should be being set to have the same range.

kortschak:
#302 has been fixed by #310.

PR #300, opened 2016-07-18
Plotter:  Add tick label precision to dintinguish adjacent labels

### LINKED ISSUE #301: Tick label are indisintinguishable for small differences
Current precision requirement for tick values allows resolving min and max tick values, but does not provide ability to resolve differences between adjacent tick values when when they are close spaced.  Note all tick labels on colorbar heatmap exactly equal at 0.0002

tickprecisionambiguity

jasonpfox:
Resolves #301

Previous precision limits guaranteed seeing at least one digit of major tick mark values; however, this was not enough to tell them apart, and all major tick marks could display the same value without precision to tell them apart.  This change increases minimum precision based on distance between the major tick marks to ensure that adjacent major labels can be distinguished from each other. 
tickprecisionambiguity
tickprecisionimproved

Note the colorbar tick labels for similar random test data sets before and after the change

### DIFF
--- a/axis.go
+++ b/axis.go
@@ -373,7 +373,12 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
 	}
 	majorDelta := float64(majorMult) * tens
 	val := math.Floor(min/majorDelta) * majorDelta
-	prec := maxInt(precisionOf(min), precisionOf(max))
+	// Precision required is not just to see a value for the ticks, but to
+	// measure and distinguish the tick points from each other accurately,
+	// limit is not to abrely resolve the min and max points, but to
+	// see the difference between points that are majorDelta away from each other
+	// /10 is more than needed, /2 probalby not enouth in all cases, so go with /3
+	prec := precisionOf(majorDelta / 3)
 	for val <= max {
 		if val >= min && val <= max {
 			ticks = append(ticks, Tick{Value: val, Label: formatFloatTick(val, prec)})
--- a/axis.go
+++ b/axis.go
@@ -374,11 +374,8 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
 	majorDelta := float64(majorMult) * tens
 	val := math.Floor(min/majorDelta) * majorDelta
 	// Precision required is not just to see a value for the ticks, but to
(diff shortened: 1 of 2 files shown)


kortschak:
Please file an issue for this. There are other PRs live that touch this and I think we should get them in first. This is a valid issue, but I don't think this is the correct solution.

axis.go
```diff
@@ -373,7 +373,12 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
 	}
 	majorDelta := float64(majorMult) * tens
 	val := math.Floor(min/majorDelta) * majorDelta
-	prec := maxInt(precisionOf(min), precisionOf(max))
+	// Precision required is not just to see a value for the ticks, but to
+	// measure and distinguish the tick points from each other accurately,
+	// limit is not to abrely resolve the min and max points, but to
+	// see the difference between points that are majorDelta away from each other
+	// /10 is more than needed, /2 probalby not enouth in all cases, so go with /3
+	prec := precisionOf(majorDelta / 3)
```
kortschak: `prec := precisionOf(majorDelta)` is good enough. The comment is not necessary.

kortschak: maxInt is no longer used when this is gone, so should be deleted.

jasonpfox:
I looked but did not find a related issue , and didn't see it when i went back through (only checked titles vice full issue threads so I could ahve missed it somewhere)...

Went back and tested and you are right that majorDelta alone is enough to resolve the differences.  Updated minimum and comment to only require majorDelta precision.

axis.go
```diff
@@ -373,7 +373,9 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
 	}
 	majorDelta := float64(majorMult) * tens
 	val := math.Floor(min/majorDelta) * majorDelta
-	prec := maxInt(precisionOf(min), precisionOf(max))
+	// Precision required is not just to see a value for the ticks, but to
+	// measure and distinguish the tick points from each other accurately
```
kortschak: Please delete the comment. The code tells us what is going on here.

Also, maxInt needs to be deleted.

kortschak:
Needs tests.

kortschak:
Counter at #303 with test.

PR #303, opened 2016-07-19
plot: make tick labels non-ambiguous

### LINKED ISSUE #301: Tick label are indisintinguishable for small differences

kortschak:
Fixes #301.

@jasonpfox @sbinet @eaburns Please take a look.

### DIFF
--- a/axis.go
+++ b/axis.go
@@ -384,7 +384,7 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
 	}
 	majorDelta := float64(majorMult) * tens
 	val := math.Floor(min/majorDelta) * majorDelta
-	prec := maxInt(precisionOf(min), precisionOf(max))
+	prec := precisionOf(majorDelta)
 	for val <= max {
 		if val >= min && val <= max {
 			ticks = append(ticks, Tick{Value: val, Label: formatFloatTick(val, prec)})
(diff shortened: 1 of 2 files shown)


jasonpfox:
Looks great, thanks.

axis.go
```diff
@@ -384,7 +384,7 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
 	}
 	majorDelta := float64(majorMult) * tens
 	val := math.Floor(min/majorDelta) * majorDelta
-	prec := maxInt(precisionOf(min), precisionOf(max))
+	prec := precisionOf(majorDelta)
```
btracey: And the logic is simpler!

btracey:
LGTM

axis.go
kortschak: Yes, thanks to @jasonpfox for that observation. I don't really know what I was thinking when I wrote the original.

axis_test.go
```diff
@@ -35,19 +35,25 @@ func TestAxisSmallTick(t *testing.T) {
...
+			Max:    0.00015,
+			Labels: []string{"0.0001", "0.00011", "0.00012", "0.00013", "0.00014"},
+		},
 	} {
 		ticks := d.Ticks(test.Min, test.Max)
 		var count int
 		for _, tick := range ticks {
 			if tick.Label != "" {
+				if count >= len(test.Labels) {
+					t.Errorf("Too many tick labels")
```
sbinet: please provide a "got=%d want=%d" information text so we know which test failed.

eaburns:
Sorry that I'm late, LGTM.

PR #308, opened 2016-08-15
Add Sankey diagram

ctessum:


ctessum:
PTAL. Let me know your thoughts about whether this fits here.

kortschak:
I think this is worthwhile having, but a couple of questions.

Why is Flow retained as a pointer and with an inUse field? It seems to me that it's not mutated except for the Group field when there is a defaulting behaviour. Would it not be better to just use a concrete Flow, not mutate it and either understand that "" is default, or pass "Default" (I don't like this) when f.Group is "".

Would you accept a bezier curve instead of a cubic spline. I have a bezier package that I can donate - I prefer to not increase imports given the trouble we have had here with moving infrastructure.

ctessum:
1. When I was starting to write it I was thinking that it would need to be mutated but that turned to not be the case. It could also work to have the default group be "", but I thought it might be confusing for beginning users that try to make a legend and find that it is a grey box with no label. This may not be an important concern though.

2. I don't have a strong knowledge of the difference between a bezier curve and a spline, but it seems like a change I would be fine with. Would this involve creating a new bezier package in gonum, copying the code into sankey.spline, or something in between?

kortschak:
1. I think it might be reasonable to prohibit "" named groups. Ascribing a default makes the package user locale opinionated, which is probably not OK.

2. The package is [here](https://godoc.org/github.com/biogo/graphics/bezier). I'm the sole author of this code, so we can bring it into gonum. Either as a unexported version of what exists there (probably the right thing to do in the short term) or as a (probably internal) package. In the first instance see if you can get your code to play with it where it is and then I can commit it as the author afterwards.

plotter/sankey.go
```diff
@@ -0,0 +1,454 @@
...
+func (s stockSorter) Len() int      { return len(s) }
+func (s stockSorter) Swap(i, j int) { s[i], s[j] = s[j], s[i] }
+func (s stockSorter) Less(i, j int) bool {
+	if s[i].category != s[j].category {
+		return s[i].category < s[j].category
+	}
+	if s[i].order != s[j].order {
+		return s[i].order < s[j].order
+	}
+	panic(fmt.Errorf("can't sort stocks:\n%+v\n%+v", s[i], s[j]))
```
sbinet: prepend error message with `"plotter:"`

plotter/sankey.go
```diff
@@ -0,0 +1,454 @@
...
+// flows and stocks.
+func NewSankey(flows ...*Flow) (*Sankey, error) {
+	s := new(Sankey)
+
+	s.stocks = make(map[int]map[string]*stock)
+
+	s.flows = flows
+	for i, f := range flows {
+		if f.inUse {
+			panic("this Flow is already in use in another Plotter")
```
sbinet: prepend error message with `"plotter:"`

also, perhaps replace with: `"plotter: Flow already in use in another Sankey diagram"`

ctessum:
I just removed the requirement for the group to not be an empty string. There is currently no requirement that the stock labels not be an empty string, so I figured it would be best for the behavior to be consistent.

plotter/sankey.go
```diff
@@ -0,0 +1,440 @@
...
+	// the plotting process.
+	sourceFlowPlaceholder, receptorFlowPlaceholder float64
+}
+
+// A Flow represents the amount of an entity flowing between two stocks.
+type Flow struct {
+	// SourceStockLabel and ReceptorStockLabel are the labels
+	// of the stocks that originate and receive the flow,
+	// respectively.
+	SourceStockLabel, ReceptorStockLabel string
```
kortschak: SourceLabel, DestinationLabel? These diagrams are used in many places where the entities are not financial. Same for the categories below.

plotter/sankey.go
```diff
@@ -0,0 +1,440 @@
...
+		// check stock category order
+		if f.SourceStockCategory >= f.ReceptorStockCategory {
+			return nil, fmt.Errorf("plotter.NewSankey: Flow %d SourceStockCategory (%d) "+
+				">= ReceptorStockCategory (%d)", i, f.SourceStockCategory, f.ReceptorStockCategory)
+		}
+		if f.Value < 0 {
+			return nil, fmt.Errorf("plotter.NewSankey: Flow %d value (%g) < 0", i, f.Value)
+		}
+
+		// initialize stock holders
```
kortschak: Proper sentence.

plotter/sankey.go
```diff
@@ -0,0 +1,440 @@
...
+		if _, ok := s.stocks[f.SourceStockCategory]; !ok {
+			s.stocks[f.SourceStockCategory] = make(map[string]*stock)
+		}
+		if _, ok := s.stocks[f.ReceptorStockCategory]; !ok {
+			s.stocks[f.ReceptorStockCategory] = make(map[string]*stock)
+		}
+
+		// figure out plotting order of stocks
```
kortschak: Ditto here.

plotter/sankey.go
```diff
@@ -0,0 +1,440 @@
...
+			}
+		}
+
+		// add to total value of stocks
+		s.stocks[f.SourceStockCategory][f.SourceStockLabel].sourceValue += f.Value
+		s.stocks[f.ReceptorStockCategory][f.ReceptorStockLabel].receptorValue += f.Value
+	}
+
+	s.LineStyle = DefaultLineStyle
+	s.LineStyle.Color = color.NRGBA{R: 0, G: 0, B: 0, A: 150}
```
kortschak: Do we impose specific colour types elsewhere?  If we do, we mostly shouldn't.

plotter/sankey.go
```diff
@@ -0,0 +1,440 @@
...
+	for _, ss := range s.stocks {
+		for _, sss := range ss {
+			stocks = append(stocks, sss)
+		}
+	}
+	sort.Sort(stockSorter(stocks))
+	return stocks
+}
+
+type stockSorter []*stock
```
kortschak: Doc comment here please.

plotter/sankey.go
```diff
@@ -0,0 +1,440 @@
...
+
+	// stocks arranges the stocks by category.
+	// The first key is the category and the seond
+	// key is the label.
+	stocks map[int]map[string]*stock
+}
+
+// stock represents the amount of a stock and its plotting order.
+type stock struct {
+	receptorValue, sourceValue float64
```
kortschak: Comments on these please.

plotter/sankey.go
```diff
@@ -0,0 +1,440 @@
...
+func NewSankey(flows ...Flow) (*Sankey, error) {
+	s := new(Sankey)
+
+	s.stocks = make(map[int]map[string]*stock)
+
+	s.flows = flows
+	for i, f := range flows {
+		// check stock category order
+		if f.SourceStockCategory >= f.ReceptorStockCategory {
+			return nil, fmt.Errorf("plotter.NewSankey: Flow %d SourceStockCategory (%d) "+
```
kortschak: A single line here is fine.

plotter/sankey.go
```diff
@@ -0,0 +1,440 @@
...
+}
+
+// Plot implements the plot.Plotter interface.
+func (s *Sankey) Plot(c draw.Canvas, plt *plot.Plot) {
+	stocks := s.stockList()
+	s.setStockMinMax(&stocks)
+
+	trCat, trVal := plt.Transforms(&c)
+
+	// draw the flows
```
kortschak: Complete sentence.

plotter/sankey.go
ctessum: I was thinking stocks and flows in the sense used [here](https://en.wikipedia.org/wiki/Stock_and_flow#More_general_uses). It is true that the terms are most commonly used in an economic context, but as the wikipedia page linked above mentions, the amount of CO2 in the atmosphere can also be thought of as a stock. If these terms are deemed not sufficiently general, though, we can use something else.

plotter/sankey.go
```diff
@@ -0,0 +1,440 @@
...
+	"fmt"
+	"image/color"
+	"math"
+	"sort"
+
+	"github.com/gonum/floats"
+	"github.com/gonum/plot"
+	"github.com/gonum/plot/vg"
+	"github.com/gonum/plot/vg/draw"
+	"github.com/ready-steady/spline"
```
kortschak: Did you try with a bezier?

plotter/sankey.go
ctessum: Lines in plots elsewhere in the package are all black by default. In this case it will typically only be useful if there is some degree of transparency because there will typically be overlapping lines that would be difficult (often impossible) to follow if they are opaque. (As an aside, the pdf backend doesn't appear to support transparency.) Having a completely transparent fill by default could be an option.

plotter/sankey.go
kortschak: Fair enough, the issue for me is more that it makes the field names java-esque. If Source and Destination are meaningful and short. SourceCategory and DestinationCategory are unfortunately longer, but an additional type to allay this doesn't seem warranted.

plotter/sankey.go
```diff
@@ -0,0 +1,432 @@
...
+)
+
+// A Sankey diagram presents stock and flow data as rectangles representing
+// the amount of each stock and lines between the stocks representing the
+// amount of each flow.
+type Sankey struct {
+	// Color specifies the default fill
+	// colors for the stocks and flows. Colors can be
+	// modified for individual stocks and flows.
+	color.Color
```
kortschak: Color should not be embedded.

plotter/sankey.go
kortschak: The lines don't need to be transparent, but the fill does. This can be resolved by taking a parameter for the fill and leaving the decision up to the user.

kortschak: An alternative that you may want to look at is how I handle the same issue in [rings](https://github.com/biogo/graphics/blob/master/rings/ribbons.go).

plotter/sankey.go
```diff
@@ -0,0 +1,440 @@
...
+
+	// StockBarWidth is the widths of the bars representing
+	// the stocks. The default value is 15% larger than the
+	// height of the stock label text.
+	StockBarWidth vg.Length
+
+	// LineStyle specifies the default border
+	// line style for the stocks and flows. Styles can be
+	// modified for individual stocks and flows.
+	draw.LineStyle
```
kortschak: LineStyle and TextStyle should not really be embedded either, particularly TextStyle since it has methods.

kortschak:
Could you remove the changes from other branches (possibly rebase onto master if you want, but the other commits confound the PR).

plotter/sankey.go
ctessum: I removed the fill, but it does make it somewhat harder for the eye to follow the flows from one side to the other.

ctessum:
PTAL. I'm still not sure about the proper usage of ClipLinesX. @eaburns ?

plotter/sankey.go
```diff
@@ -0,0 +1,472 @@
...
+
+	// TextStyle specifies the default stock label
+	// text style. Styles can be modified for
+	// individual stocks.
+	TextStyle draw.TextStyle
+
+	flows []Flow
+
+	// FlowStyle is a function that specifies the
+	// background colar and border line style of the
```
kortschak: s/colar/color/

plotter/sankey.go
```diff
@@ -0,0 +1,472 @@
...
+	// Group specifies the group that a flow belongs
+	// to. It is used in assigning styles to groups
+	// and creating legends.
+	Group string
+}
+
+// NewSankey creates a new Sankey diagram with the specified
+// flows and stocks.
+func NewSankey(flows ...Flow) (*Sankey, error) {
+	s := new(Sankey)
```
kortschak: var s Sankey

plotter/sankey.go
```diff
@@ -0,0 +1,472 @@
...
+
+	s.FlowStyle = func(_ string) (color.Color, draw.LineStyle) {
+		return s.Color, s.LineStyle
+	}
+
+	s.StockStyle = func(label string, category int) (string, draw.TextStyle, vg.Length, vg.Length, color.Color, draw.LineStyle) {
+		return label, s.TextStyle, 0, 0, s.Color, s.LineStyle
+	}
+
+	return s, nil
```
kortschak: `return &s, nil`

plotter/sankey.go
kortschak: I think the fill should be there for the reason that you give. I just don't think the way that you are specifying the alpha is right (it ties the user to NRGBA) and how the default styles are set. Taking the fill from the package default feels wrong. What's worse is that it then gets captured in a closure and so can't be altered without replacing the closure.

It seems to me that much of this could be done in Plot so that the user has had a chance to alter fields. Unfortunately because of the plot.Plot.Draw error handling the font handling here can't be done there - it would be good if it could so you don't need to foist the default font onto the user - API toggles by global on a per use basis is pretty smelly. Really it seems to me that the TextStyle field should be left zero and handled by the user. Once you do that, L170-190 go into Plot.

plotter/sankey.go
```diff
@@ -0,0 +1,472 @@
...
+	s.stocks = make(map[int]map[string]*stock)
+
+	s.flows = flows
+	for i, f := range flows {
+		// Here we make sure the stock categories are in the proper order.
+		if f.SourceCategory >= f.ReceptorCategory {
+			return nil, fmt.Errorf("plotter.NewSankey: Flow %d SourceCategory (%d) >= ReceptorCategory (%d)", i, f.SourceCategory, f.ReceptorCategory)
+		}
+		if f.Value < 0 {
+			return nil, fmt.Errorf("plotter.NewSankey: Flow %d value (%g) < 0", i, f.Value)
```
kortschak: s/.NewSankey//g

kortschak:
Where are we up to on this?

ctessum:
Sorry for the delay. I have responded to the remaining comments and made some additional changes. PTAL.

plotter/sankey.go
```diff
@@ -276,7 +276,6 @@ func (s *Sankey) Plot(c draw.Canvas, plt *plot.Plot) {
 			{catMin, valMin},
 			{catMax, valMin},
 		}
-		// outline := c.ClipLinesX(pts) // This causes half of the lines to disappear.
 		c.StrokeLines(lineStyle, pts) //outline...)
```
kortschak: `//outline...)` and below?

ctessum:
done. @eaburns PTAL

eaburns:
Very pretty plots, though I don't really understand them.

My one issue is that I don't like the outline-box drawn around the plot *within* the axes. In my opinion it violates the idea of maximizing the data:ink ratio. The box adds too much ink and no additional information. It's on purpose that Plotinum didn't have a box around the entire plot like many other plotting systems do. Anyway, if you all prefer that, then I don't want to block anything.

I skimmed the code. It looks very nice. LGTM after a squash.

ctessum:
The idea behind sankey diagrams is to visualize stocks and flows, where stocks are some quantity and each flow some transfer function that relates two stocks. Perhaps the most famous sankey diagram is [this one](https://upload.wikimedia.org/wikipedia/commons/2/29/Minard.png) where the stocks are numbers of soldiers and the flows are loss rates between events. A more common use of the diagrams, is to [visualize energy utilization](http://www.sankey-diagrams.com/wp-content/gallery/o_gallery_200/gea_sankey_02.png), where the stocks are different types of energy (either waste or useful) and the flows are conversions between the different types. 

Both of the examples linked above used solid colors with no outlines to delimit the different stocks and flows, but the implementation here uses outlines with no fill color by default. So the outline box around the whole plotter does represent data: the y-height of the box gives the total value of the stock in the system at a given point on the x axis (just like in a stacked bar plot). So in [this example](https://github.com/gonum/plot/blob/79244fc3dfdc0ba090efeb540636a665e1b6de91/plotter/testdata/sankeySimple_golden.png), the outlines show that the trees produce 18 apples total but we only know the fate of 17 of them. If you look closely, you can see that this is due to a difference between the number of apples Sofia recieves and the number of apples she does something with. These types of accounting mismatches are common in the types of data used in these plots, and different implementations handle them differently.

eaburns:
I see. So is it sufficient to only have the line on the top? Just curious.

ctessum:
We could, but I think that would be the equivalent of not outlining the bottom of the bar in a bar chart, but bar charts currently [do](https://github.com/gonum/plot/blob/master/plotter/barchart.go#L149) currently have lines at bottom.

eaburns:
Yeah the bar charts are kind of noisy. Oh well. To late to fix them.

PR #310, opened 2016-08-26
plotter: explicitly panic when HeatMap.Min >= HeatMap.Max

### LINKED ISSUE #302: Creating a heatmap with 0 depth (min value == max value) causes panic

kortschak:
Fixes #302.


eaburns:
lgtm

jasonpfox:
Have you created an issue to ensure that X- and Y- axis behavior is consistent with this approach?

kortschak:
No, and I don't intend to for reasons given in the discussion in the issue and your PR. Consistency for consistency's sake is not a goal; context is important.

PR #313, opened 2016-09-12
Implemented an AutoUnixTimeTicks

abourget:
Could probably be the default eventually.

It allows configurability of density of ticks with the `Width`, and aims
to be legible on a per-inch basis. Normally, if you set Width to be
the same width you render the graph with, things should look good.

Check tests for a feel of what it would look like once rendered (for 4 inch-wide graph).

### DIFF
--- a/axis.go
+++ b/axis.go
@@ -7,7 +7,6 @@ package plot
 import (
 	"image/color"
 	"math"
-	"os"
 	"strconv"
 	"time"
 
@@ -512,113 +511,85 @@ type AutoUnixTimeTicks struct {
 var _ Ticker = AutoUnixTimeTicks{}
 
 // Inspired by https://github.com/d3/d3-scale/blob/master/src/time.js
-var tickIntervals = []tickRule{
-	{time.Millisecond,
-		"15:04:05.000", "15:04:05", ".000", time.Millisecond},
-	{200 * time.Millisecond,
-		"15:04:05.000", "15:04:05", ".000", 200 * time.Millisecond},
-	{500 * time.Millisecond,
-		"15:04:05.000", "15:04:05", ".000", 500 * time.Millisecond},
-	{time.Second,
-		"15:04:05", "15:04", ":05", time.Second},
-	{2 * time.Second,
-		"15:04:05", "15:04", ":05", 2 * time.Second},
-	{5 * time.Second,
-		"15:04:05", "15:04", ":05", 5 * time.Second},
-	{15 * time.Second,
-		"Jan 02, 15:04", "Jan 02", "15:04", 15 * time.Second},
-	{30 * time.Second,
-		"15:04:05", "15:04", ":05", 30 * time.Second},
-	{time.Minute,
-		"15:04:05", "15:04", ":05", time.Minute},
-	{2 * time.Minute,
-		"Jan 02, 3:04pm", "Jan 02", "3:04pm", 2 * time.Minute},
-	{5 * time.Minute,
-		"Jan 02, 3:04pm", "Jan 02", "3:04pm", 5 * time.Minute},
-	{15 * time.Minute,
-		"Jan 02, 3:04pm", "Jan 02", "3:04pm", 15 * time.Minute},
-	{30 * time.Minute,
-		"Jan 02, 3:04pm", "Jan 02", "3:04pm", 30 * time.Minute},
-	{time.Hour,
-		"Jan 2, 3pm", "Jan 2", "3pm", time.Hour},
-	{3 * time.Hour,
-		"Jan 2, 3pm", "Jan 2", "3pm", 3 * time.Hour},
-	{6 * time.Hour,
-		"Jan 2, 3pm", "Jan 2", "3pm", 6 * time.Hour},
-	{12 * time.Hour,
-		"Jan 2", "Jan 2", "3pm", 12 * time.Hour},
-	{24 * time.Hour,
-		"Jan 2", "Jan", "2", 24 * time.Hour},
-	{48 * time.Hour,
-		"Jan 2", "Jan", "2", 48 * time.Hour},
-	{7 * 24 * time.Hour,
-		"Jan 2", "Jan", "2", 7 * 24 * time.Hour},
-	{aMonth, // 1 month
-		"Jan 2006", "2006", "Jan", aMonth},
-	{3 * aMonth, // 3 months
-		"Jan 2006", "2006", "Jan", 3 * aMonth},
-	{6 * aMonth, // 6 months
-		"Jan 2006", "2006", "Jan", 6 * aMonth},
-	{12 * aMonth, // 1 year
-		"2006", "", "2006", 12 * aMonth},
-	{2 * 12 * aMonth, // 2 years
-		"2006", "", "2006", 2 * 12 * aMonth},
-	{5 * 12 * aMonth, // 5 years
-		"2006", "", "2006", 5 * 12 * aMonth},
-	{10 * 12 * aMonth, // 10 years
-		"2006", "", "2006", 10 * 12 * aMonth},
+var tickRules = []tickRule{
+	{time.Millisecond, "15:04:05.000", "15:04:05", ".000"},
+	{200 * time.Millisecond, "15:04:05.000", "15:04:05", ".000"},
+	{500 * time.Millisecond, "15:04:05.000", "15:04:05", ".000"},
+	{time.Second, "15:04:05", "15:04", ":05"},
+	{2 * time.Second, "15:04:05", "15:04", ":05"},
+	{5 * time.Second, "15:04:05", "15:04", ":05"},
+	{15 * time.Second, "Jan 02, 15:04", "Jan 02", "15:04"},
+	{30 * time.Second, "15:04:05", "15:04", ":05"},
+	{time.Minute, "15:04:05", "15:04", ":05"},
+	{2 * time.Minute, "Jan 02, 3:04pm", "Jan 02", "3:04pm"},
+	{5 * time.Minute, "Jan 02, 3:04pm", "Jan 02", "3:04pm"},
+	{15 * time.Minute, "Jan 02, 3:04pm", "Jan 02", "3:04pm"},
+	{30 * time.Minute, "Jan 02, 3:04pm", "Jan 02", "3:04pm"},
+	{time.Hour, "Jan 2, 3pm", "Jan 2", "3pm"},
+	{3 * time.Hour, "Jan 2, 3pm", "Jan 2", "3pm"},
+	{6 * time.Hour, "Jan 2, 3pm", "Jan 2", "3pm"},
+	{12 * time.Hour, "Jan 2", "Jan 2", "3pm"},
+	{24 * time.Hour, "Jan 2", "Jan", "2"},
+	{48 * time.Hour, "Jan 2", "Jan", "2"},
+	{7 * 24 * time.Hour, "Jan 2", "Jan", "2"},
+	{month, "Jan 2006", "2006", "Jan"},
+	{3 * month, "Jan 2006", "2006", "Jan"},
+	{6 * month, "Jan 2006", "2006", "Jan"},
+	{12 * month, "2006", "", "2006"},
+	{2 * year, "2006", "", "2006"},
+	{5 * year, "2006", "", "2006"},
+	{10 * year, "2006", "", "2006"},
 }
 
-var aMonth = 31 * 24 * time.Hour
+const month = 31 * 24 * time.Hour
+const year = 12 * month
 
+// tickRule defines a time display format for a given time window (per
+// inch).
+//
+// This assumes a tick about each `durationPerInch`.  The long format is
+// shown each time the timestamp goes over a certain boundary
+// (verified through `watchFormat`).  This way you can show `Sep 2,
+// 12pm` when you pass midnight after `11pm` on `Sep 1`.
 type tickRule struct {
-	DurationPerInch time.Duration // use this rule for a maximum Duration per inch
-	LongFormat      string        // longer format
-	WatchFormat     string        // show long format when WatchFormat changes between ticks
-	ShortFormat     string        // incremental format, shorter
-	TimeWindow      time.Duration // interval for ticks
+	durationPerInch time.Duration // use this rule for a maximum Duration per inch, it is also used as an interval per ticks.
+	longFormat      string        // longer format
+	watchFormat     string        // show long format when watchFormat changes between ticks
+	shortFormat     string        // incremental format, shorter
 }
 
-// Ticks implements plot.Ticker.
+// Ticks implements plot.Ticker and displays appropriately spaced and
+// formatted time labels.
 func (t AutoUnixTimeTicks) Ticks(min, max float64) []Tick {
 	width := t.Width
 	if width == 0 {
-		width = 4 * vg.Inch
+		width = 10 * vg.Centimeter
 	}
 
 	minT := time.Unix(int64(min), 0).UTC()
 	maxT := time.Unix(int64(max), 0).UTC()
 	durationPerInch := maxT.Sub(minT) / time.Duration(width/vg.Inch)
 
-	rule := tickIntervals[len(tickIntervals)-1]
-	for idx := range tickIntervals[:len(tickIntervals)-1] {
-		if durationPerInch < tickIntervals[idx+1].DurationPerInch {
-			rule = tickIntervals[idx]
+	lastElement := len(tickRules) - 1
+	rule := tickRules[lastElement]
+	for idx, tickRule := range tickRules[:lastElement] {
+		if durationPerInch < tickRules[idx+1].durationPerInch {
+			rule = tickRule
 			break
 		}
 	}
 
-	ticks := []Tick{
-	//	{Value: min, Label: minT.Format(rule.BeginFormat)},
-	}
-
-	monthsDelta := time.Month(rule.TimeWindow / aMonth)
-
-	//fmt.Println("Months delta", int(monthsDelta))
-	start := minT.Truncate(rule.TimeWindow)
-	count := 0
-	lastWatch := ""
+	timeWindow := rule.durationPerInch
+	delta := time.Month(timeWindow / month) // in months
+	start := minT.Truncate(timeWindow)
+	var lastWatch string
+	var ticks []Tick
 	for {
-		count++
-		if monthsDelta > 0 {
+		if delta > 0 {
 			// Count in Months now
-			start = time.Date(start.Year(), start.Month()+monthsDelta, 1, 0, 0, 0, 0, time.UTC)
+			start = time.Date(start.Year(), start.Month()+delta, 1, 0, 0, 0, 0, time.UTC)
 		} else {
-			start = start.Add(rule.TimeWindow)
-		}
-
-		if count > 20 {
-			os.Exit(0)
+			start = start.Add(timeWindow)
 		}
 
 		if start.Before(minT) {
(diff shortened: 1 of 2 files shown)


axis.go
```diff
@@ -497,6 +498,158 @@ func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
...
+		count++
+		if monthsDelta > 0 {
+			// Count in Months now
+			start = time.Date(start.Year(), start.Month()+monthsDelta, 1, 0, 0, 0, 0, time.UTC)
+		} else {
+			start = start.Add(rule.TimeWindow)
+		}
+
+		if count > 20 {
+			os.Exit(0)
```
sbinet: that's pretty ugly :)
at the very least, this should be a `panic("plot: run away procedure")`.

axis.go
```diff
@@ -497,6 +498,158 @@ func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
...
+
+	rule := tickIntervals[len(tickIntervals)-1]
+	for idx := range tickIntervals[:len(tickIntervals)-1] {
+		if durationPerInch < tickIntervals[idx+1].DurationPerInch {
+			rule = tickIntervals[idx]
+			break
+		}
+	}
+
+	ticks := []Tick{
```
sbinet: replace with just:
```go
var ticks []Tick
```

axis_test.go
```diff
@@ -59,3 +63,176 @@ func labelsOf(ticks []Tick) []string {
...
+			max:  "2020-08-28 13:59:00",
+			want: []string{"2016", "2017", "2018", "2019", "2020"},
+		},
+		{
+			min:  "2016-01-01 12:01:05",
+			max:  "2048-08-28 13:59:00",
+			want: []string{"2017", "2022", "2027", "2032", "2037", "2042", "2047"},
+		},
+	} {
+		fmt.Println("For dates", test.min, test.max)
```
sbinet: remove this line

axis.go
```diff
@@ -497,6 +498,158 @@ func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
...
+		"2006", "", "2006", 12 * aMonth},
+	{2 * 12 * aMonth, // 2 years
+		"2006", "", "2006", 2 * 12 * aMonth},
+	{5 * 12 * aMonth, // 5 years
+		"2006", "", "2006", 5 * 12 * aMonth},
+	{10 * 12 * aMonth, // 10 years
+		"2006", "", "2006", 10 * 12 * aMonth},
+}
+
+var aMonth = 31 * 24 * time.Hour
```
sbinet: could probably be a `const` instead of a `var` ?

axis.go
```diff
@@ -497,6 +498,158 @@ func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
...
+
+type tickRule struct {
+	DurationPerInch time.Duration // use this rule for a maximum Duration per inch
+	LongFormat      string        // longer format
+	WatchFormat     string        // show long format when WatchFormat changes between ticks
+	ShortFormat     string        // incremental format, shorter
+	TimeWindow      time.Duration // interval for ticks
```
sbinet: not sure why any of these fields are exported.
couldn't they just be private?

axis.go
```diff
@@ -497,6 +498,158 @@ func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
...
+}
+
+// Ticks implements plot.Ticker.
+func (t AutoUnixTimeTicks) Ticks(min, max float64) []Tick {
+	width := t.Width
+	if width == 0 {
+		width = 4 * vg.Inch
```
sbinet: pet peeve: could this be (a round value of) centimeters instead of inches?
(`gonum` in general is trying to follow SI units)

axis.go
sbinet: perhaps just return nil and document?

sbinet:
ok, first pass done :)
I'll dig deeper this evening (Paris time).

looks interesting, thanks!

axis.go
abourget: woohaahh! that was just for debugging :) sorry it went through.. should just take that out.. if you have a sane time range, you won't be hitting that.

axis.go
abourget: yeah they could.. do you think it'd be useful to let people tweak certain rules ? change the formatting..

right now, there's zero internationalization of the labels, etc.. like `d3-times` scales, it's just English.. I don't currently mind.. but it's very hard-coded..

axis.go
abourget: yeah, more left overs :)

axis.go
kortschak: Called `month`.

axis.go
```diff
@@ -497,6 +498,158 @@ func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
...
+	// the number of ticks that can fit properly with their time
+	// shown.
+	Width vg.Length
+}
+
+var _ Ticker = AutoUnixTimeTicks{}
+
+// Inspired by https://github.com/d3/d3-scale/blob/master/src/time.js
+var tickIntervals = []tickRule{
+	{time.Millisecond,
```
kortschak: Don't break these over two lines.

axis.go
```diff
@@ -497,6 +498,158 @@ func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
 	return ticks
 }
 
+// AutoUnixTimeTicks is suitable for axes representing time values.
+// AutoUnixTimeTicks expects values in Unix time seconds.  It will
+// adjust the number of ticks according to the specified Width.  If
+// not specified, Width defaults to 4 inches.
+type AutoUnixTimeTicks struct {
```
kortschak: I'm not convinced by the name.

axis.go
kortschak: Unexport them for now.

axis.go
kortschak: Yes, please.

axis.go
```diff
@@ -497,6 +498,158 @@ func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
...
+}
+
+// Ticks implements plot.Ticker.
+func (t AutoUnixTimeTicks) Ticks(min, max float64) []Tick {
+	width := t.Width
+	if width == 0 {
+		width = 4 * vg.Inch
+	}
+
+	minT := time.Unix(int64(min), 0).UTC()
```
kortschak: There is no need for the `T` suffix, just `min` and `max`.

axis.go
```diff
@@ -497,6 +498,158 @@ func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
...
+	{5 * 12 * aMonth, // 5 years
+		"2006", "", "2006", 5 * 12 * aMonth},
+	{10 * 12 * aMonth, // 10 years
+		"2006", "", "2006", 10 * 12 * aMonth},
+}
+
+var aMonth = 31 * 24 * time.Hour
+
+type tickRule struct {
+	DurationPerInch time.Duration // use this rule for a maximum Duration per inch
```
kortschak: Please either an SI or typesetting unit. SI would be nice, but we have what we have. Maybe express it as the reciprocal and use `vg.Length`.

axis.go
```diff
@@ -497,6 +498,158 @@ func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
...
+		"2006", "", "2006", 2 * 12 * aMonth},
+	{5 * 12 * aMonth, // 5 years
+		"2006", "", "2006", 5 * 12 * aMonth},
+	{10 * 12 * aMonth, // 10 years
+		"2006", "", "2006", 10 * 12 * aMonth},
+}
+
+var aMonth = 31 * 24 * time.Hour
+
+type tickRule struct {
```
kortschak: Doc comments here as complete grammatical sentences please.

axis.go
```diff
@@ -497,6 +498,158 @@ func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
...
+			rule = tickIntervals[idx]
+			break
+		}
+	}
+
+	ticks := []Tick{
+	//	{Value: min, Label: minT.Format(rule.BeginFormat)},
+	}
+
+	monthsDelta := time.Month(rule.TimeWindow / aMonth)
```
kortschak: s/monthsDelta/delta/g

axis.go
```diff
@@ -497,6 +498,158 @@ func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
...
+
+	//fmt.Println("Months delta", int(monthsDelta))
```
kortschak: Delete.

axis.go
```diff
@@ -497,6 +498,158 @@ func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
...
+	start := minT.Truncate(rule.TimeWindow)
+	count := 0
+	lastWatch := ""
```
kortschak: `var last string`

axis.go
```diff
@@ -497,6 +498,158 @@ func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
...
+
+	ticks := []Tick{
+	//	{Value: min, Label: minT.Format(rule.BeginFormat)},
+	}
+
+	monthsDelta := time.Month(rule.TimeWindow / aMonth)
+
+	//fmt.Println("Months delta", int(monthsDelta))
+	start := minT.Truncate(rule.TimeWindow)
+	count := 0
```
kortschak: `var n int`

axis.go
```diff
@@ -497,6 +498,158 @@ func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
...
+		}
+
+		if start.Before(minT) {
+			continue
+		}
+		if start.After(maxT) {
+			break
+		}
+
+		label := ""
```
kortschak: `var label string`

axis.go
```diff
@@ -497,6 +498,158 @@ func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
...
+			//showing the longform first.
+			label = start.Format(rule.LongFormat)
+		}
+		lastWatch = newWatch
+
+		ticks = append(ticks, Tick{
+			Value: float64(start.UnixNano()) / float64(time.Second),
+			Label: label,
+		})
+
```
kortschak: Delete blank line.

axis.go
```diff
@@ -497,6 +498,158 @@ func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
...
+	if width == 0 {
+		width = 4 * vg.Inch
+	}
+
+	minT := time.Unix(int64(min), 0).UTC()
+	maxT := time.Unix(int64(max), 0).UTC()
+	durationPerInch := maxT.Sub(minT) / time.Duration(width/vg.Inch)
+
+	rule := tickIntervals[len(tickIntervals)-1]
+	for idx := range tickIntervals[:len(tickIntervals)-1] {
```
kortschak: Get the value here so you have it by the time you get to L596.

axis.go
```diff
@@ -497,6 +498,158 @@ func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
...
+	// Width is the width of the underlying graph, used to calculate
+	// the number of ticks that can fit properly with their time
+	// shown.
+	Width vg.Length
+}
+
+var _ Ticker = AutoUnixTimeTicks{}
+
+// Inspired by https://github.com/d3/d3-scale/blob/master/src/time.js
+var tickIntervals = []tickRule{
```
kortschak: `tickRules`

axis.go
```diff
@@ -497,6 +498,158 @@ func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
...
+
+		if start.Before(minT) {
+			continue
+		}
+		if start.After(maxT) {
+			break
+		}
+
+		label := ""
+		newWatch := start.Format(rule.WatchFormat)
```
kortschak: What is the meaning of "watch" here?

abourget:
Folks I'm so sorry to have submitted a half-baked PR like this.  I was in a hurry and haven't gotten time to clean it up.  I'll do it the moment things calm down over here.

axis.go
abourget: `watch` is probably not the right word... it is the format we check for changes, to decide if we display the long form (like when we change from `Sep 1` to `Sep 2` when showing hours), or the short (like `11pm`, then `Sep 2, 12pm` when `watch` changes)

If you look into tests, you'll see there are 2 sorts of labels per graph.. long and short forms.

axis.go
abourget: I don't mind changing the default value to `10 * vg.Centimer` .. the user can however specify that how he wishes in the `Width` field on the `AutoUnixTimeTicks` (which is a `vg.Length`).

The other place where inches are used is in conjunction with `DurationPerInch`, and these are used solely in internal definitions.  I come from Canada where we use SI everywhere. But I think an inch is about the right spacing between ticks, so defining the values like:
```
	{time.Millisecond, "15:04:05.000", "15:04:05", ".000", time.Millisecond},
	{200 * time.Millisecond, "15:04:05.000", "15:04:05", ".000", 200 * time.Millisecond},
	{500 * time.Millisecond, "15:04:05.000", "15:04:05", ".000", 500 * time.Millisecond},
```

seems more fitting.. where `500 * time.Millisecond` is per inch.. it _could_ be `DurationPer2dot5Centimer` .. just more awkward :)  We'd need to redefine all those rules to some other values.. and use either `Centimer` or `Decimeter`.. both of which don't align easily on ticks.  Finding visually appealing values for those rules is also tricky.. so I borrowed a bunch of them from `d3`.. recomputing and realigning them would be a pain, only to internally not use inches.

Are you okay with keeping inches for the rules ?

axis.go
abourget: heh.. I know it isn't pretty :) `UnixTimeTicks` was already taken.. and `Auto` was used to distinguish the added feature..

Could we make it `TimeTicks`  ? and document the fact that you need Unix timestamps, or provide a function to transform a `[]time.Time` into whatever this thing needs ?

axis.go
abourget: It's there to distinguish from the function parameters `min` and `max`, shortly indicating they're `time.Time` instead of a `float64`.  Would you prefer some other names ?

eaburns:

abourget:
Folks, could we reopen this ? It's still valid.. don't know why it fell into the cracks.. I'm still using that and I'd be grateful to have it upstream !

axis.go
```diff
@@ -497,6 +497,129 @@ func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
 	return ticks
 }
 
+// TimeTicks is suitable for axes representing time values.
+// TimeTicks expects values in Unix time seconds.  It will
+// adjust the number of ticks according to the specified Width.  If
+// not specified, Width defaults to 10 centimeters.
+type TimeTicks struct {
+	// Width is the width of the underlying graph, used to calculate
```
kortschak: s/graph/plot/

axis.go
```diff
@@ -497,6 +497,129 @@ func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
...
+	{month, "Jan 2006", "2006", "Jan"},
+	{3 * month, "Jan 2006", "2006", "Jan"},
+	{6 * month, "Jan 2006", "2006", "Jan"},
+	{12 * month, "2006", "", "2006"},
+	{2 * year, "2006", "", "2006"},
+	{5 * year, "2006", "", "2006"},
+	{10 * year, "2006", "", "2006"},
+}
+
+const month = 31 * 24 * time.Hour
```
kortschak: These can be in a const block. Also, I'm sort of troubled by the choice of length of month; the marginally closer length is 30 days.

axis.go
```diff
@@ -497,6 +497,129 @@ func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
...
+const year = 12 * month
+
+// tickRule defines a time display format for a given time window (per
+// inch).
+//
+// This assumes a tick about each `durationPerInch`.  The long format is
+// shown each time the timestamp goes over a certain boundary
+// (verified through `watchFormat`).  This way you can show `Sep 2,
+// 12pm` when you pass midnight after `11pm` on `Sep 1`.
+type tickRule struct {
```
kortschak: This type does not need to exist. You can define `tickRules` on the anonymous struct.

axis.go
```diff
@@ -497,6 +497,129 @@ func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
...
+	watchFormat     string        // show long format when watchFormat changes between ticks
+	shortFormat     string        // incremental format, shorter
+}
+
+// Ticks implements plot.Ticker and displays appropriately spaced and
+// formatted time labels.
+func (t TimeTicks) Ticks(min, max float64) []Tick {
+	width := t.Width
+	if width == 0 {
+		width = 10 * vg.Centimeter
```
kortschak: What is the basis for this?

axis.go
```diff
@@ -497,6 +497,129 @@ func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
...
+
+// tickRule defines a time display format for a given time window (per
+// inch).
+//
+// This assumes a tick about each `durationPerInch`.  The long format is
+// shown each time the timestamp goes over a certain boundary
+// (verified through `watchFormat`).  This way you can show `Sep 2,
+// 12pm` when you pass midnight after `11pm` on `Sep 1`.
+type tickRule struct {
+	durationPerInch time.Duration // use this rule for a maximum Duration per inch, it is also used as an interval per ticks.
```
kortschak: SI please.

axis.go
```diff
@@ -497,6 +497,129 @@ func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
...
+	{10 * year, "2006", "", "2006"},
+}
+
+const month = 31 * 24 * time.Hour
+const year = 12 * month
+
+// tickRule defines a time display format for a given time window (per
+// inch).
+//
+// This assumes a tick about each `durationPerInch`.  The long format is
```
kortschak: Please use single space after period throughout.

axis.go
```diff
@@ -497,6 +497,129 @@ func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
...
+// shown each time the timestamp goes over a certain boundary
+// (verified through `watchFormat`).  This way you can show `Sep 2,
+// 12pm` when you pass midnight after `11pm` on `Sep 1`.
+type tickRule struct {
+	durationPerInch time.Duration // use this rule for a maximum Duration per inch, it is also used as an interval per ticks.
+	longFormat      string        // longer format
+	watchFormat     string        // show long format when watchFormat changes between ticks
```
kortschak: This still needs a better name, Given that we are talking about time, "watch" has two possible meanings.

axis.go
```diff
@@ -497,6 +497,129 @@ func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
...
+	if width == 0 {
+		width = 10 * vg.Centimeter
+	}
+
+	minT := time.Unix(int64(min), 0).UTC()
+	maxT := time.Unix(int64(max), 0).UTC()
+	durationPerInch := maxT.Sub(minT) / time.Duration(width/vg.Inch)
+
+	lastElement := len(tickRules) - 1
+	rule := tickRules[lastElement]
```
kortschak: Just use `rule := tickRules[len(tickRules)-1]` and similarly in the range.

axis.go
```diff
@@ -497,6 +497,129 @@ func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
...
+	}
+
+	timeWindow := rule.durationPerInch
+	delta := time.Month(timeWindow / month) // in months
+	start := minT.Truncate(timeWindow)
+	var lastWatch string
+	var ticks []Tick
+	for {
+		if delta > 0 {
+			// Count in Months now
```
kortschak: Full stop.

axis.go
```diff
@@ -497,6 +497,129 @@ func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
...
+		if start.After(maxT) {
+			break
+		}
+
+		var label string
+		newWatch := start.Format(rule.watchFormat)
+		if lastWatch == newWatch {
+			label = start.Format(rule.shortFormat)
+		} else {
+			//TODO: overwrite the first tick with the long form if we
```
kortschak: s|//TODO: o|// TODO(name): O|

axis.go
```diff
@@ -497,6 +497,129 @@ func (utt UnixTimeTicks) Ticks(min, max float64) []Tick {
...
+			// haven't shown a lonform at all.. instead of always
```
kortschak: s/lonform/longform/
s/\.\./,/

axis_test.go
```diff
@@ -59,3 +62,176 @@ func labelsOf(ticks []Tick) []string {
...
+			max:  "2020-08-28 13:59:00",
+			want: []string{"2016", "2017", "2018", "2019", "2020"},
+		},
+		{
+			min:  "2016-01-01 12:01:05",
+			max:  "2048-08-28 13:59:00",
+			want: []string{"2017", "2022", "2027", "2032", "2037", "2042", "2047"},
+		},
+	} {
+		//fmt.Println("For dates", test.min, test.max)
```
kortschak: Delete this line.

Issue #284, opened 2016-05-31
Replace plotter.XYs with slices


btracey:
In my experience, most data exists as x, y []float64, and not as []struct{X,Y float64}. plotter should replace these types

See discussion https://groups.google.com/d/msg/gonum-dev/FrGpgA-pZQs/FCy8op2JCwAJ

kortschak:
Now if we had tables...

sbinet:
_wrt_ tables, as far as `gonum/plot` is concerned, everything is a `float64`.
so, perhaps we could get away with a `float64`-based ndim array?
but then, is it really the best option to couple `gonum/plot` with that specific implementation?

should we give @Kunde21's `numgo.Array64` a go?

relying on `[]float64` seems simpler. row-wise data isn't very cache friendly anyways...

kortschak:
That was not a serious suggestion.

ctessum:
Why not just add the code:

``` go
type XYArray struct {
   X, Y []float64
}

func (xy XYArray) Len() int {
   return len(xy.X)
}

func (xy XYArray) XY(i int) (x,y float64) {
   return xy.X[i], xy.Y[i]
}
```

Then one could just do 
``` go
vals := XYArray{X: x, Y: y}
```
and directly add vals to a plotter.

btracey:
https://godoc.org/github.com/btracey/myplot#VecXY :)

Still, I think the slice representation is more natural for gonum/plot. I think it's more common for users of the code, and it's easier to add in more information (say, z data). Frequently the data does need to exist as a slice even internal to plot, see for example the code for NewErrorPoints

ctessum:
So is the question, then, whether to replace the XYs type with VecXY or XYArray, or whether to replace the XYer interface with a concrete type in function arguments? i.e. 
``` go
NewScatter(xys XYer) 
```
would become 
``` go
NewScatter(x, y []float64)
```
To me, the first option seems like a good idea. I see a tradeoff with the second option, where making the change would make things easier for a lot use cases (i.e., where the data to be plotted is already in x and y arrays), but it would make things more difficult for some use cases (e.g., plotting geographic locations, which typically are in the form struct{X, Y float64}). 

I would lean toward keeping the current signature for the function arguments but replacing XYs with VecXY or XYVector as that seems to be most in line with the 'do less, enable more' philosophy, and allows the user to make decisions about their data format rather than having the decisions made for them by the plotting package.

btracey:
[...]

PR #315, opened 2016-10-03
Zip convenience functions

Kunde21:
Convenience functions discussed in #284 implemented.  

We can likely have better discussions about its value with a concrete implementation.

### DIFF
--- a/plotutil/xyer.go
+++ b/plotutil/xyer.go
@@ -1,22 +1,14 @@
+// Copyright ©2016 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
 package plotutil
 
 import "github.com/gonum/plot/plotter"
 
+// xyer implements plotter.XYer for use by ZipXY.
 type xyer struct {
 	x, y []float64
 	len  int
-}
-
-// Len returns the length of the XYer object.
-// fulfills the plotter.XYer interface requirement.
-func (xy *xyer) Len() int {
-	return xy.len
-}
-
-// XY returns the x,y values at index idx.
-// fulfills the plotter.XYer interface requirement.
-func (xy *xyer) XY(idx int) (x, y float64) {
-	return xy.x[idx], xy.y[idx]
 }
 
 // ZipXY is a convenience function to build a plotter.XYer from slices.
@@ -36,21 +28,22 @@ func ZipXY(x, y []float64) plotter.XYer {
 	return &xyer{x: x, y: y, len: ln}
 }
 
+// Len returns the length of the XYer object.
+// fulfills the plotter.XYer interface requirement.
+func (xy *xyer) Len() int {
+	return xy.len
+}
+
+// XY returns the x,y values at index idx.
+// fulfills the plotter.XYer interface requirement.
+func (xy *xyer) XY(idx int) (x, y float64) {
+	return xy.x[idx], xy.y[idx]
+}
+
+// xyzer implements plotter.XYZer for use by ZipXYZ.
 type xyzer struct {
 	x, y, z []float64
 	len     int
-}
-
-// Len returns the length of the XYZer object.
-// fulfills the plotter.XYZer interface requirement.
-func (xyz *xyzer) Len() int {
-	return xyz.len
-}
-
-// XYZ returns the x,y,z values at index idx.
-// fulfills the plotter.XYZer interface requirement.
-func (xyz *xyzer) XYZ(idx int) (x, y, z float64) {
-	return xyz.x[idx], xyz.y[idx], xyz.z[idx]
 }
 
 // ZipXYZ is a convenience function to build a plotter.XYZer from slices.
@@ -74,4 +67,16 @@ func ZipXYZ(x, y, z []float64) plotter.XYZer {
 	}
 
 	return &xyzer{x: x, y: y, z: z, len: ln}
+}
+
+// Len returns the length of the XYZer object.
+// fulfills the plotter.XYZer interface requirement.
+func (xyz *xyzer) Len() int {
+	return xyz.len
+}
+
+// XYZ returns the x,y,z values at index idx.
+// fulfills the plotter.XYZer interface requirement.
+func (xyz *xyzer) XYZ(idx int) (x, y, z float64) {
+	return xyz.x[idx], xyz.y[idx], xyz.z[idx]
 }
--- a/plotutil/xyer.go
+++ b/plotutil/xyer.go
@@ -1,82 +1,71 @@
 // Copyright ©2016 The gonum Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
+
 package plotutil
 
-import "github.com/gonum/plot/plotter"
+import (
+	"fmt"
 
-// xyer implements plotter.XYer for use by ZipXY.
-type xyer struct {
-	x, y []float64
-	len  int
+	"github.com/gonum/plot/plotter"
+)
+
+// VecXY implements plotter.XYer.
+type VecXY struct {
+	X, Y []float64
 }
 
-// ZipXY is a convenience function to build a plotter.XYer from slices.
-// Length will be set to the shortest slice length.
-// Panics on empty or nil slices.
+// ZipXY is a convenience function to build a VecXY from slices with length checks.
+// Panics on length mismatch.
 func ZipXY(x, y []float64) plotter.XYer {
-	switch {
-	case len(x) == 0:
-		panic("x is an empty or nil slice")
-	case len(y) == 0:
-		panic("y is an empty or nil slice")
+	if len(y) != len(x) {
+		panic(fmt.Sprintf("VecXY length mismatch %d != %d", len(x), len(y)))
 	}
-	ln := len(x)
-	if len(y) < ln {
-		ln = len(y)
-	}
-	return &xyer{x: x, y: y, len: ln}
+	return &VecXY{X: x, Y: y}
 }
 
-// Len returns the length of the XYer object.
-// fulfills the plotter.XYer interface requirement.
-func (xy *xyer) Len() int {
-	return xy.len
+// Len returns the length of the VecXY object.
+// Panics on slice length mismatch.
+// Fulfills the plotter.XYer interface requirement.
+func (xy *VecXY) Len() int {
+	if len(xy.X) != len(xy.Y) {
+		panic(fmt.Sprintf("VecXY length mismatch %d != %d", len(xy.X), len(xy.Y)))
+	}
+	return len(xy.X)
 }
 
 // XY returns the x,y values at index idx.
 // fulfills the plotter.XYer interface requirement.
-func (xy *xyer) XY(idx int) (x, y float64) {
-	return xy.x[idx], xy.y[idx]
+func (xy *VecXY) XY(idx int) (x, y float64) {
+	return xy.X[idx], xy.Y[idx]
 }
 
 // xyzer implements plotter.XYZer for use by ZipXYZ.
-type xyzer struct {
-	x, y, z []float64
-	len     int
+type VecXYZ struct {
+	X, Y, Z []float64
 }
 
 // ZipXYZ is a convenience function to build a plotter.XYZer from slices.
 // Length will be set to the shortest slice length.
 // Panics on empty or nil slices.
 func ZipXYZ(x, y, z []float64) plotter.XYZer {
-	switch {
-	case len(x) == 0:
-		panic("x is an empty or nil slice")
-	case len(y) == 0:
-		panic("y is an empty or nil slice")
-	case len(z) == 0:
-		panic("z is an empty or nil slice")
+	if len(y) != len(x) || len(y) != len(z) {
+		panic(fmt.Sprintf("VecXYZ length mismatch %d != %d != %d", len(x), len(y), len(z)))
 	}
-	ln := len(x)
-	if len(y) < ln {
-		ln = len(y)
-	}
-	if len(z) < ln {
-		ln = len(z)
-	}
-
-	return &xyzer{x: x, y: y, z: z, len: ln}
+	return &VecXYZ{X: x, Y: y, Z: z}
 }
 
 // Len returns the length of the XYZer object.
 // fulfills the plotter.XYZer interface requirement.
-func (xyz *xyzer) Len() int {
-	return xyz.len
+func (xyz *VecXYZ) Len() int {
+	if len(xyz.Y) != len(xyz.X) || len(xyz.Y) != len(xyz.Z) {
+		panic(fmt.Sprintf("VecXYZ length mismatch %d != %d != %d", len(xyz.X), len(xyz.Y), len(xyz.Z)))
+	}
+	return len(xyz.X)
 }
 
 // XYZ returns the x,y,z values at index idx.
 // fulfills the plotter.XYZer interface requirement.
-func (xyz *xyzer) XYZ(idx int) (x, y, z float64) {
-	return xyz.x[idx], xyz.y[idx], xyz.z[idx]
+func (xyz *VecXYZ) XYZ(idx int) (x, y, z float64) {
+	return xyz.X[idx], xyz.Y[idx], xyz.Z[idx]
 }
--- a/plotutil/xyer.go
+++ b/plotutil/xyer.go
@@ -1,0 +1,77 @@
+package plotutil
+
+import "github.com/gonum/plot/plotter"
+
+type xyer struct {
+	x, y []float64
+	len  int
+}
+
+// Len returns the length of the XYer object.
+// fulfills the plotter.XYer interface requirement.
+func (xy *xyer) Len() int {
+	return xy.len
+}
+
+// XY returns the x,y values at index idx.
+// fulfills the plotter.XYer interface requirement.
+func (xy *xyer) XY(idx int) (x, y float64) {
+	return xy.x[idx], xy.y[idx]
+}
+
+// ZipXY is a convenience function to build a plotter.XYer from slices.
+// Length will be set to the shortest slice length.
+// Panics on empty or nil slices.
+func ZipXY(x, y []float64) plotter.XYer {
+	switch {
+	case len(x) == 0:
+		panic("x is an empty or nil slice")
+	case len(y) == 0:
+		panic("y is an empty or nil slice")
+	}
+	ln := len(x)
+	if len(y) < ln {
+		ln = len(y)
+	}
+	return &xyer{x: x, y: y, len: ln}
+}
+
+type xyzer struct {
+	x, y, z []float64
+	len     int
+}
+
+// Len returns the length of the XYZer object.
+// fulfills the plotter.XYZer interface requirement.
+func (xyz *xyzer) Len() int {
+	return xyz.len
+}
+
+// XYZ returns the x,y,z values at index idx.
+// fulfills the plotter.XYZer interface requirement.
+func (xyz *xyzer) XYZ(idx int) (x, y, z float64) {
+	return xyz.x[idx], xyz.y[idx], xyz.z[idx]
+}
+
+// ZipXYZ is a convenience function to build a plotter.XYZer from slices.
+// Length will be set to the shortest slice length.
+// Panics on empty or nil slices.
+func ZipXYZ(x, y, z []float64) plotter.XYZer {
+	switch {
+	case len(x) == 0:
+		panic("x is an empty or nil slice")
+	case len(y) == 0:
+		panic("y is an empty or nil slice")
+	case len(z) == 0:
+		panic("z is an empty or nil slice")
+	}
+	ln := len(x)
+	if len(y) < ln {
+		ln = len(y)
+	}
+	if len(z) < ln {
+		ln = len(z)
+	}
+
+	return &xyzer{x: x, y: y, z: z, len: ln}
+}
--- a/plotutil/xyer_test.go
+++ b/plotutil/xyer_test.go
@@ -1,3 +1,6 @@
+// Copyright ©2016 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
 package plotutil
 
 import (
@@ -74,36 +77,36 @@ var tests = []struct {
 		xyzpanic: "",
 	},
 	{
-		x:        randSlice(5),
-		y:        randSlice(5),
-		z:        randSlice(5),
+		x:        genSlice(5),
+		y:        genSlice(5),
+		z:        genSlice(5),
 		xylen:    5,
 		xyzlen:   5,
 		xypanic:  "",
 		xyzpanic: "",
 	},
 	{
-		x:        randSlice(4),
-		y:        randSlice(5),
-		z:        randSlice(5),
+		x:        genSlice(4),
+		y:        genSlice(5),
+		z:        genSlice(5),
 		xylen:    4,
 		xyzlen:   4,
 		xypanic:  "",
 		xyzpanic: "",
 	},
 	{
-		x:        randSlice(5),
-		y:        randSlice(4),
-		z:        randSlice(5),
+		x:        genSlice(5),
+		y:        genSlice(4),
+		z:        genSlice(5),
 		xylen:    4,
 		xyzlen:   4,
 		xypanic:  "",
 		xyzpanic: "",
 	},
 	{
-		x:        randSlice(5),
-		y:        randSlice(5),
-		z:        randSlice(4),
+		x:        genSlice(5),
+		y:        genSlice(5),
+		z:        genSlice(4),
 		xylen:    5,
 		xyzlen:   4,
 		xypanic:  "",
@@ -111,7 +114,7 @@ var tests = []struct {
 	},
 }
 
-func randSlice(len int) []float64 {
+func genSlice(len int) []float64 {
 	sl := make([]float64, len)
 	for i := range sl {
 		sl[i] = float64(i)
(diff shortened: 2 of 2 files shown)


plotutil/xyer.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+// XY returns the x,y values at index idx.
+// fulfills the plotter.XYer interface requirement.
+func (xy *xyer) XY(idx int) (x, y float64) {
+	return xy.x[idx], xy.y[idx]
+}
+
+// ZipXY is a convenience function to build a plotter.XYer from slices.
+// Length will be set to the shortest slice length.
```
kortschak: It think this is the wrong behaviour. Everywhere else we require a length match - doing that here and below would reduce the number of rules from 2 to 1 and the number of fields by 1.

plotutil/xyer_test.go
```diff
@@ -0,0 +1,172 @@
+package plotutil
```
kortschak: License header.

plotutil/xyer.go
```diff
@@ -0,0 +1,77 @@
+package plotutil
```
kortschak: License header.

plotutil/xyer.go
```diff
@@ -0,0 +1,77 @@
...
+// XY returns the x,y values at index idx.
+// fulfills the plotter.XYer interface requirement.
+func (xy *xyer) XY(idx int) (x, y float64) {
+	return xy.x[idx], xy.y[idx]
+}
+
+// ZipXY is a convenience function to build a plotter.XYer from slices.
+// Length will be set to the shortest slice length.
+// Panics on empty or nil slices.
+func ZipXY(x, y []float64) plotter.XYer {
```
kortschak: Move this to just below the type definition (this reflect the godoc rendering) - also below.

I'm not a fan of the function name; an idiomatic name would be `XYerFrom()` or similar.

plotutil/xyer.go
```diff
@@ -0,0 +1,77 @@
+package plotutil
+
+import "github.com/gonum/plot/plotter"
+
+type xyer struct {
```
kortschak: Not exported, but add a doc comment anyway.

plotutil/xyer.go
```diff
@@ -0,0 +1,82 @@
...
+// Length will be set to the shortest slice length.
+// Panics on empty or nil slices.
+func ZipXY(x, y []float64) plotter.XYer {
+	switch {
+	case len(x) == 0:
+		panic("x is an empty or nil slice")
+	case len(y) == 0:
+		panic("y is an empty or nil slice")
+	}
+	ln := len(x)
```
btracey: I think @kortschak  said this earlier, but this should panic on unequal lengths.

plotutil/xyer.go
btracey: Another option is to do 

````
type VecXY struct {
   X, Y []float64
}

func (v VecXY) Len() int {
    if len(v.X) != len(v.Y) { 
        panic("length mismatch")
   }
   return len(x)
}

func (v VecXY) XY(i int) (x, y float64){
    return x[i], y[i]
}
````

This shifts where the panic happens, but I don't think that's a huge problem. It removes having an unexported type as part of a package, which gets tricky for things like json. This is easy enough to construct, plotutil.VecXY{x,y}, and it's not clear to me that we really need to panic in the len == 0 case.  Name can be modified obviously.

Comment (Kunde21) [inline plotutil/xyer.go]: The implementation decisions were focused as much as possible on the "convenience" aspect.  It's not meant to be a replacement data structure for `plotter.XYs` or anything like that.  Each object is only expected to exist long enough to plot the target data.  That's why the return type is the interface itself, users should make no assumptions about the underlying implementation.  The struct type isn't exported, so there's no way to type assert from the interface without unsafe/reflect hacks.  That's also why I used the 'Zip' name, so users will expect the result will be zipped touples (x[i], y[i], <z[i]>) from slice parameters.  There's no need and limited ability to dig into the guts of the resulting object.  If that's what they need, a call to `plotter.CopyXYs(plotutil.ZipXY(x,y))` does the job with the objects that already exist.

The target is where a `plotter.XYer` or `plotter.XYZer` interface type is required and the data isn't necessarily in a data structure that easily implements the interface for each combination of slices.  

Something like:
```go
var data *mat64.Dense = loadData("file.csv")
times := data.RawRowView(0)
rowlen,_ := data.Dims()

for i:=1; i < rowlen; i++ {
        line, _ := plotter.NewLine(plotutil.ZipXY(times, data.RawRowView(i)))
        line.Plot(canvas, plot)
}
```

I used the shortest length in order to allow for ragged matrix structures without needing to juggle lengths.  I was trying to make it as convenient/flexible as possible, while only `panic`-ing on empty (0-length) returns that wouldn't show up on the resulting plot.

kortschak: I prefer the approach suggested by @btracey.

btracey: @Kunde21  I don't see any reason we would change the implementation in a meaningful way, so I don't think there's any need to hide the implementation behind an interface. The explicit approach I suggested is simpler, more transparent, and makes it obvious that the data is not copied.

As to shortening the length automatically, that's against the spirit of gonum APIs. Inputs are either correct or we panic. We don't silently make corrections to make an output happen; that's the easiest way to have bugs which are never discovered.

plotutil/xyer.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+// VecXY implements plotter.XYer.
+type VecXY struct {
+	X, Y []float64
+}
+
+// ZipXY is a convenience function to build a VecXY from slices with length checks.
+// Panics on length mismatch.
+func ZipXY(x, y []float64) plotter.XYer {
```
kortschak: I still don't understand the need for this function.

If eventually it does get included though, it should return z VecXY (or whatever it is called - renaming it `XYSlices` or `XYVecs` seems sensible since it would then be grouped with `XYer` and `XYs` in the godoc - same for the XYZ case).

plotutil/xyer.go
```diff
@@ -0,0 +1,70 @@
...
+package plotutil
+
+import (
+	"fmt"
+
+	"github.com/gonum/plot/plotter"
+)
+
+// VecXY implements plotter.XYer.
+type VecXY struct {
```
kortschak: I think the naming could be improved (below).

eaburns:
Any news on this? We are trying to clean up our PRs. It seems like this change has some outstanding comments. Do you plan to address them? If not, we plan to close this soon.

plotutil/xyer.go
eaburns: Please make sure there is a blank line between the license comment and the `package` declaration statement, or else godoc will think that the license is a doc comment.

eaburns:
This keeps coming up. Many folks want it. We need to add it. There's no real good excuse not to have it.

… but allow me a minor rant. My hope for Plotinum was that a community would form around it, building on it as a scaffolding to create their own tools, styles, and packages. The `plotter` package that is checked into this repository has *no special privledge* over any other package that anyone else writes. This was by design. I hoped this would allow lots of freedom to try out new things, without being bottlenecked on some single repo and some single person's (or small group's, as it's turned out) ideas. `plotter` was always intended partially as an example for how one could write their own tools on top of `plot`. Maybe I never made that clear enough. Anyway, now folks seem to think that their code, their `Plotter`, their data format, their whatever, is somehow second-class if it's not checked in here. That's simply not true…

Anyway, things worked out differently, and we should get this PR merged.

I agree with @kortschak I think that we should make this an exported type like the one described in the comment from July 7th in #284 (but please don't call it `zip`. I like Dan's suggestion of `XYSlices`). The type approach is:
 1. Equally as easy to use (`XYSlices{X: xs, Y: ys}` instead of `XYSlices(xs, ys)`) and
 1. Makes it clear that it is not copying under the hood.

Kunde21:
@eaburns As it stands, I've got both styles built.  

- `VecXY` and `VecXYZ` are exported structs that can be used like `NewScatter(VecXY{X: xs, Y: ys})`.  The names aren't set in stone, but we can paint that bikeshed `VecXY`, `XYSlices`, `XYVec`,  `XYData`, `XY`, etc..  There's no perfect name, so I'm not too hung up on any one.
- `ZipXY` and `ZipXYZ` validate length matches, avoiding panics within the plotting functions (which would give confusing stack traces).  These return `plotter.XYer` and `plotter.XYZer` interface types, respectively, so the slice lengths can't be changed once they're validated.  To me, this also helps inexperienced programmers match the return type to the needed type in the plotter functions, especially those coming from Python/Scala/etc with built-in `zip` functions.

btracey:
I'm with @kortschak, I don't like the Zip functions. All they do is check lengths, which is easily verifiable by the user. 

Something like
````
func SimpleLinePlot(x, y []float64, name string) error {
       if len(x) != len(y) {
              return errors.New("mypkg: plotting length mismatch")
       } 
       v := plotutil.VecXY{x,y}
      // do plot stuff
}
````
Is much better style.

**Comment by Kunde21** [comment]
@btracey Are we really helping the user when the convenience type requires the very boilerplate that they're hoping to avoid?  

It's easily possible for every user to create a `XYer`-implementing struct.  Not providing any  real convenience functions feels like doing less than the bare minimum to me.

**Comment by kortschak** [comment]
>`ZipXY` and `ZipXYZ` validate length matches, avoiding panics within the plotting functions (which would give confusing stack traces).

I don't believe the stack traces would be confusing. In all the `.../plot/plotter` plotter types, the input data are copied in the New function and call the value collection's Len method. So the panic does not happen in `plot.Plot` where is might be confusing, but rather in the creation of the plotter.

**Comment by btracey** [comment]
Code should never be written such that there's a reasonable expectation that it panics (with rare exceptions, but this is not one of them). Thus, ZipXY should never be called when the user is unsure that the lengths are the same. Thus, a length check should happen anyway before calling ZipXY. So, ZipXY doesn't even save a length check.

ZipXY performs two functions. a) It changes where the panic happens when the coder has made a mistake, b) it makes it harder to mess with the lengths of the slices after they've put in the struct.  a) does not seem to merit its cost, since the overall effect on code is basically identical, and it should be a rare event anyway  b) is counterbalanced by the fact that it also obscures whether the slices are copied or not copied, and so obscures the fact that x and y should not be modified afterward. These at best cancel out (though I think the downsides are greater), and we should side toward limited API.

**Comment by Kunde21** [comment]
@kortschak Correct, but that means the stack trace points the user into the `plot.Plotter` type's function call, then back out to the `VecXY.Len` function.  Such an error has nothing to do with the `plot.Plotter` call, but it still shows up in the stack trace, in no way easily differentiated from the actual error.  

@btracey I highly doubt anyone is writing plotter code with the expectation that it panics.  That's not the point of any of this code.  If it's possible, we should assume that it will/does happen with some regularity.  It's impossible to limit these libraries to only the most experienced/talented programmers.  A warning in the Zip documentation can limit the effect of (b).  Conversely, there's no easy way to limit the effect of the easily-made mistake that causes (a) without putting all the boilerplate into `VecXY`'s documentation.  At that point, we're just forcing every user to copy-paste instead of implementing it ourselves.

**Comment by btracey** [comment]
What do you mean by "Without putting all the boilerplate into VecXY's documentation"? You can have

````
// XYSlices implements plotter.XYer using slices of data. It panics during a call to Len() if the slices have different lengths.
````
ZipXY needs to have at least as much documentation. 

How are we forcing everyone to copy/paste? This provides an implementation of `XYSlices`, so that not everyone has to.

**Comment by kortschak** [comment]
No, I don't believe that is true. The following diff implements the minimal vector bind and passes.

```diff
diff --git a/plotter/scatter_test.go b/plotter/scatter_test.go
index 8566e67..1ed36eb 100644
--- a/plotter/scatter_test.go
+++ b/plotter/scatter_test.go
@@ -5,6 +5,7 @@
 package plotter
 
 import (
+       "fmt"
        "image/color"
        "log"
        "math/rand"
@@ -36,7 +37,7 @@ func ExampleScatter() {
        }
 
        n := 15
-       scatterData := randomPoints(n)
+       scatterData := randomSlices(n, rnd)
        lineData := randomPoints(n)
        linePointsData := randomPoints(n)
 
@@ -86,3 +87,40 @@ func ExampleScatter() {
 func TestScatter(t *testing.T) {
        checkPlot(ExampleScatter, t, "scatter.png")
 }
+
+func randomSlices(n int, rnd *rand.Rand) VecXY {
+       v := VecXY{
+               X: make([]float64, n),
+               Y: make([]float64, n),
+       }
+       for i := range v.X {
+               if i == 0 {
+                       v.X[i] = rnd.Float64()
+               } else {
+                       v.X[i] = v.X[i-1] + rnd.Float64()
+               }
+               v.Y[i] = v.X[i] + 10*rnd.Float64()
+       }
+       return v
+}
+
+// VecXY implements plotter.XYer.
+type VecXY struct {
+       X, Y []float64
+}
+
+// Len returns the length of the VecXY object.
+// Panics on slice length mismatch.
+// Fulfills the plotter.XYer interface requirement.
+func (xy VecXY) Len() int {
+       if len(xy.X) != len(xy.Y) {
+               panic(fmt.Sprintf("VecXY length mismatch %d != %d", len(xy.X), len(xy.Y)))
+       }
+       return len(xy.X)
+}
+
+// XY returns the x,y values at index idx.
+// fulfills the plotter.XYer interface requirement.
+func (xy VecXY) XY(idx int) (x, y float64) {
+       return xy.X[idx], xy.Y[idx]
+}
```

Changing `randomSlices` to truncate `v.X` by one element panics with

```
$ go run scatter_main.go 
panic: VecXY length mismatch 14 != 15

goroutine 1 [running]:
panic(0x644a80, 0xc42000aa90)
	/home/daniel/Development/go/src/runtime/panic.go:500 +0x1a1
main.VecXY.Len(0xc42009a580, 0xe, 0xf, 0xc42009a600, 0xf, 0xf, 0xc420013020)
	/home/daniel/Desktop/scatter.go:116 +0x176
main.(*VecXY).Len(0xc420013020, 0xc42000aa70)
	<autogenerated>:1 +0x63
github.com/gonum/plot/plotter.CopyXYs(0x7af180, 0xc420013020, 0xc420013020, 0x30, 0x30, 0x1, 0xc420013020)
	/home/daniel/Development/src/github.com/gonum/plot/plotter/plotter.go:139 +0x35
github.com/gonum/plot/plotter.NewScatter(0x7af180, 0xc420013020, 0xc420013020, 0x7af180, 0xc420013020)
	/home/daniel/Development/src/github.com/gonum/plot/plotter/scatter.go:27 +0x3c
main.main()
	/home/daniel/Desktop/scatter_main.go:55 +0x366
exit status 2
```

After you get out of the runtime, the first func is `CopyXYs` followed by `NewScatter` which is where you would expect it to panic. This is at creation, not use.

Kunde21:
I'll hearken back to my point about novice programmer trying to parse the stack trace.  Sure, VecXY can panic in the call to Len, but is that where they're going to look when they see `plotter.NewScanner`, `plotter.CopyXY`, and `make` in the stack trace?  Are we providing any help/convenience by creating `VecXY` if it forces them to chase down that panic?  

Thus, the only helpful way to guide those novice programmers would be to give an example in the documentation showing a loop that checks lengths before creating that `VecXY`, bringing me back to my "giving boilerplate instead of hoisting it" point.  

Experienced programmers can/will skip the use of the Zip functions altogether, so I'm a bit confused when it sounds like we're assuming expert-level knowledge when discussing them.  These functions were about convenience in the first place, so I feel our base assumption should be that novice vice the pro.

Kunde21:
@kortschak A single plot line is a very minimal example.  

Think of it this way:  User reads a wide-dataset csv file into a `[][]float`, not realizing that one row is truncated.  They attempt to plot each row against a time, but get that error at some random place in their for loop.  Now what?  How are they to know how some unknown slice in the `[][]float` is one element shorter than the rest?  How much anger and frustration is going to build before they find the problem?  How likely is it that they're going to reach for gonum the next time a new dataset shows up?  

An ounce of prevention....

kortschak:
I don't see how the two approaches differ in that example. In the loop they are constructing the plotter or they are constructing the data for the plotter. In one case the New plotter function panics, in the other the new data function panics.

Kunde21:
The new data panic points to the data, whereas the new plotter function gets around to it.  

Keep in mind, all of this is supposed to be about convenience.  Are we really serving that purpose by giving just enough to create convoluted errors?

btracey:
`XYSlices` gives a lot of convenience (I've been using my own implementation of the type for 2 years now). It saves either boilerplate code to implement the type (about 14 lines), or boilerplate code to transform into plotter.XY (about 6, plus a bunch of other ramifications).

The panic that comes out of `XYSlices` is very legible -- the data you told it to plot has mismatched lengths. That should point even a novice coder to check where the length of their data could have gone wrong. Inserting a length check just before constructing the `XYSlices` is equally as informative as replacing the creating of `XYSlices` with a call to `ZipXY`.

Kunde21:
>  a length check just before constructing the XYSlices is equally as informative as replacing the creating of XYSlices with a call to ZipXY.

I'm in complete agreement, but which is more likely in the code of a novice?  You seem to be assuming that every single programmer that might ever touch this package will be a grizzled veteran.  Realistically, the user is more likely to be a botanist/physicist/geologist/<other science field here> that's just trying to hack together a plot to visualize their data.  Helper functions with defensive validation is the least we can do to help them succeed.  If we can't do that, in what way have we really provided convenience in these convenience types?

btracey:
`XYSlices` is a convenience type with defensive validation.

Kunde21:
`XYSlices` has no inherent defensive validation.  Removing the length check from `Len` would provide a similar panic during the copy phase.  Either way, it happens inside of the `plot.plotter` function call, so I can't see how it can be classified as "defensive" at all.

kortschak:
>`XYSlices` has no inherent defensive validation. Removing the length check from Len would provide a similar panic during the copy phase.

By giving an informative panic message it is vastly superior.

>Either way, it happens inside of the `plot.plotter` function call, so I can't see how it can be classified as "defensive" at all.

I'm afraid I don't know what this `plot.plotter` function call is. I've tried to understand this in the thread, but it eludes me. The panic happens in the `plotter.New<PlotterType>` call, which is not very different to the `plotter.VecXY` call - one level of panic indirection.

Kunde21:
>   I've tried to understand this in the thread, but it eludes me. The panic happens in the plotter.New<PlotterType> call, which is not very different to the plotter.VecXY call - one level of panic indirection.  

Bare minimum, two levels of indirection (NewPlotterType and CopyXY).  

That's not the point of this code, though.  If all we're looking for is a slightly more informative panic message, it would be done inside `CopyXY` and we wouldn't even need to create an `XYSlices` struct at all.  Three lines of code could do that, but we wouldn't really be addressing the concerns of #284.

Kunde21:
I'd also like to point out that every user in #284 that isn't a gonum contributor asked for some sort of Zip function.  These users we should be thinking about when working on this PR.  My code might not be prefect, but completely throwing out the requested function should be supported by better rationale than "It's not impossible to do without."

kortschak:
People want and are justified in asking for a way of presenting a pair of slices as an XYer. Without the New function the PR still does that. I would suggest that we merge this PR after cleaning up some remaining issues, mark it as updating #284 and see how people feel after it has been used as a type only.

I'll go through the PR now and comment with this in mind.

plotutil/xyer.go
```diff
@@ -0,0 +1,71 @@
...
+// ZipXY is a convenience function to build a VecXY from slices with length checks.
+// Panics on length mismatch.
+func ZipXY(x, y []float64) plotter.XYer {
+	if len(y) != len(x) {
+		panic(fmt.Sprintf("VecXY length mismatch %d != %d", len(x), len(y)))
+	}
+	return &VecXY{X: x, Y: y}
+}
+
+// Len returns the length of the VecXY object.
```
kortschak: Please use complete sentences: "Len returns the length of the T value." (I would like T to be `XYSlices`)

plotutil/xyer.go
```diff
@@ -0,0 +1,71 @@
...
+// Panics on slice length mismatch.
+// Fulfills the plotter.XYer interface requirement.
+func (xy *VecXY) Len() int {
```
kortschak: This does not need to be on a pointer (also below for other types and methods).

plotutil/xyer.go
```diff
@@ -0,0 +1,71 @@
...
+	if len(xy.X) != len(xy.Y) {
+		panic(fmt.Sprintf("VecXY length mismatch %d != %d", len(xy.X), len(xy.Y)))
```
kortschak: `panic(fmt.Sprintf("plotter: vector length mismatch %d != %d", len(xy.X), len(xy.Y)))`

plotutil/xyer.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	return len(xy.X)
+}
+
+// XY returns the x,y values at index idx.
```
kortschak: "XY returns the x and y values at index i." (Please replace `idx` below with `i`).

plotutil/xyer.go
```diff
@@ -0,0 +1,71 @@
...
+// fulfills the plotter.XYer interface requirement.
+func (xy *VecXY) XY(idx int) (x, y float64) {
+	return xy.X[idx], xy.Y[idx]
+}
+
+// xyzer implements plotter.XYZer for use by ZipXYZ.
```
kortschak: Please fix the name here and omit mention of `ZipXYZ`. The form for `VecXY` is good.

plotutil/xyer.go
kortschak: Please delete this function.

plotutil/xyer.go
```diff
@@ -0,0 +1,71 @@
...
+type VecXYZ struct {
+	X, Y, Z []float64
+}
+
+// ZipXYZ is a convenience function to build a plotter.XYZer from slices.
```
kortschak: Please delete this function.

plotutil/xyer.go
```diff
@@ -0,0 +1,71 @@
...
+// Length will be set to the shortest slice length.
+// Panics on empty or nil slices.
+func ZipXYZ(x, y, z []float64) plotter.XYZer {
+	if len(y) != len(x) || len(y) != len(z) {
+		panic(fmt.Sprintf("VecXYZ length mismatch %d != %d != %d", len(x), len(y), len(z)))
+	}
+	return &VecXYZ{X: x, Y: y, Z: z}
+}
+
+// Len returns the length of the XYZer object.
```
kortschak: Similar wording as above for the XY type's Len method, no need for the implements comment.

plotutil/xyer.go
```diff
@@ -0,0 +1,71 @@
...
+// fulfills the plotter.XYZer interface requirement.
+func (xyz *VecXYZ) Len() int {
```
kortschak: This does not need to be on the pointer.

plotutil/xyer.go
```diff
@@ -0,0 +1,71 @@
...
+	if len(xyz.Y) != len(xyz.X) || len(xyz.Y) != len(xyz.Z) {
+		panic(fmt.Sprintf("VecXYZ length mismatch %d != %d != %d", len(xyz.X), len(xyz.Y), len(xyz.Z)))
```
kortschak: It would be good to be more explicit about which mismatch was the cause. If we have {1,2,3}, {1,2,3}, {1,2} the panic is necessarily false - just report the mismatched pair or triple.

plotutil/xyer.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	return len(xyz.X)
+}
+
+// XYZ returns the x,y,z values at index idx.
```
kortschak: Same as for the XY method.

plotutil/xyer.go
Kunde21: Type `VecXY`/`XYSlices`/`TypeNameHere` can be used by any package in the go universe.  Prepending "plotter:" does not fit that reality.

kortschak: Sure, but it is a `plotter` functionality. This is the convention.

Kunde21: So, is this supposed to be `plotter`-specific functionality (which was best implemented by the original, non-exported type) or a general utility type (which shouldn't have a plotter-specific error)?  It really can't be both.

kortschak: I don't understand the dichotomy. It is provided by `plotter` and it is general, though it is designed specifically for things that take `plotter.XYer` or `plotter.XYZer` depending on type and is more likely to be used there. If you want, you could make them "plotter.XYSlices: ..." and "plotter.XYZSlices: ...".

eaburns:
Now I regret deleting my rant!

The whole design of Plutonium was exactly in anticipation that folks will disagree with the plotter package, and that all could still be fine. The plotter package has no special privileges. I wasn't convinced that I'd get it right or that everyone would be happy with it, so I made sure that it has no special access to the underlying plot engine.

Anyone can implement their own plotters with their interfaces and their own styles. And I hope that they do, the same way that folks have their own configs and their own themes for their favorite text editor. That way, we can reasonably disagree on the exact interface, and everyone still gets their way. As @btracy points out, they've been using their own implementation of XYSlices for years. It works and everything has been fine.

I agree, there is an issue with discoverability. Folks write code, it's good, and they want others to benefit from it. I understand. I think that the solution isn't to put it all in one monolithic package that is bottlenecked on the opinions of this one particular group of reviewers. What we need is folks writing lots of packages, doing lots of cool things and a central place where we can advertise them: a prominent wiki page on gonum/plot where everyone can link their ideas. Perhaps the zip function can be in the first such package.

plotutil/xyer.go
btracey: Rather, `plotter: slice length`

eaburns:
What's the decision? It seems like most folks want `XYSlices` and no `Zip` function.

decibelcooper:
I agree with adding `XYSlices` and no `Zip` function.  My thought is `XYZip` is a better name for the type, since its the receiver that implements the XYer interface that does the "zipping".  However, this is of course a minor concern and it doesn't really matter.

sbinet:
closing of old age.
(feel free to reopen if you feel it isn't warranted)

PR #317, opened 2016-10-25
plot,plotter: fix time tick location handling

### LINKED ISSUE #276: plotter: UnixTimeTicks is not time.Location agnostic

kortschak:
Fixes #276.

This draws from #278, including comments from @kostya-sh. It is a simpler fix than #278.

@eaburns Please take a look.


eaburns:
LGTM

PR #319, opened 2016-10-30
Export CheckPlot and move to internal

ctessum:
...so that it can be used in places other than plotter.


ctessum:
Ping? #321 depends on this.

internal/cmpimg/checkplot.go
```diff
@@ -0,0 +1,73 @@
...
+			golden := goldenPath(path)
+			_ = os.Remove(golden)
+			if err := os.Rename(path, golden); err != nil {
+				t.Fatal(err)
+			}
+		}
+		return
+	}
+
+	// Overwrite the Golden Images.
```
sbinet: is that comment still true?

ctessum: No it is not.

PR #320, opened 2016-10-30
vg,plotter/testdata: update golden images for changes in draw2d

kortschak:
Updates golden images to fix skew with github.com/llgcode/draw2d.

@ctessum Please take a look.


ctessum:
LGTM

PR #324, opened 2016-11-13
Add link to Community Plotters page in README.md

eaburns:


sbinet:
LGTM

eaburns:
Travis is sad, but it's certainly not because of this PR.

sbinet:
yeah... maybe another draw2d commit?

eaburns:
That's quite annoying. I'm wondering if we should vendor the dependencies, but the recommendation seems to be that vendoring should happen only for binary packages. Do we ignore that advice? Can anyone think of a better solution?

eaburns:
Maybe we can configure travis to pull particular commits for each dependency. That way, binary packages pulling gonum/plot don't get blocked by our internal vendoring, but we can still test to know upstream versions.

PR #322, opened 2016-10-30
Added ColorMap types and legends

ctessum:
This is related to issues #304, #133, and #17

PR #332, opened 2016-12-17
Added ColorMap types and legends

ctessum:
This is a continuation #322 which got closed after I merged the branch it was based on into master.


kortschak:
Please fix the import cycle in the tests.

palette/palette.go
```diff
@@ -32,54 +39,54 @@ type DivergingPalette interface {
...
+type ColorMapInt interface {
+	// At returns the color associated with category cat.
+	At(cat int) (color.Color, error)
+}
 
-	// SetMin sets the minimum value of the ColorMap.
-	SetMin(float64)
+// IntMap fulfils the ColorMapInt interface, mapping integer
+// categories to colors.
+type IntMap struct {
```
ctessum: This can't be a map because it needs to be ordered. It could be a slice I guess, but I'm not sure if that would be an improvement.

palette/palette.go
```diff
@@ -32,54 +39,54 @@ type DivergingPalette interface {
...
+}
 
-	// Alpha returns the opacity value of the ColorMap.
-	Alpha() float64
+// At fulfils the ColorMapInt interface.
+func (im *IntMap) At(cat int) (color.Color, error) {
+	if len(im.Categories) != len(im.Colors) {
+		panic(fmt.Errorf("palette: number of categories (%d) != number of colors (%d)", len(im.Categories), len(im.Colors)))
+	}
+	if i := sort.SearchInts(im.Categories, cat); i < len(im.Categories) && im.Categories[i] == cat {
```
ctessum: It's conceivable that people would use a large number of categories, so I've left this as binary search, but that also wouldn't be a best practice so it might be better to use the non-binary search.

ctessum:
PTAL

ctessum:
Ping?

palette/benchmarkgo17_test.go
```diff
@@ -0,0 +1,74 @@
...
+package palette
+
+import (
+	"fmt"
+	"image/color"
+	"math/rand"
+	"testing"
+)
+
+func BenchmarkColorMap_At(b *testing.B) {
```
kortschak: The benchmarks were informative, but do we need to keep them?

palette/legend.go
```diff
@@ -0,0 +1,118 @@
+// Copyright ©2016 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package palette
```
kortschak: I don't understand why this is in palette. It seems more like something for plotter - and then would be a separate PR.

palette/legend.go
```diff
@@ -0,0 +1,118 @@
...
+	"image"
+
+	"github.com/gonum/plot"
+	"github.com/gonum/plot/vg"
+	"github.com/gonum/plot/vg/draw"
+)
+
+// ColorMapLegend is a plot.Plotter that draws a color bar legend for a ColorMap.
+type ColorMapLegend struct {
+	// Vertical determines wether the legend will be
```
kortschak: whether; we don't keep sheep.

palette/legendcategorical.go
```diff
@@ -0,0 +1,203 @@
+// Copyright ©2016 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package palette
```
kortschak: Again, wrong package IMO.

palette/palette.go
```diff
@@ -32,54 +38,76 @@ type DivergingPalette interface {
...
+	}
+	return nil, fmt.Errorf("palette: category '%d' not found", cat)
+}
 
-	// SetMin sets the minimum value of the ColorMap.
-	SetMin(float64)
+// searchInts returns the index of ints that matches i, or -1 if there
+// is no match.
+func searchInts(ints []int, i int) int {
+	for ii, iii := range ints {
```
kortschak: Please don't speak Go with a Fortran accent.

palette/benchmarkgo17_test.go
ctessum: Well, the obvious answer is if they are informative we should keep them, right? If by "informative" you meant "not informative", then we probably shouldn't keep them, however.

My initial thinking was that this method could theoretically end up in inner loops, so it might be useful to have a benchmark.

kortschak: I meant "informative", though border line; I would have been extremely surprised if they had fallen the other way.

I don't see why plotting is going to end up in an inner loop, at least not under the normal understanding of that phrase; there are too many things in plot that take too long (with justification) for that to be a sensible engineering solution.

sbinet:
ping? I have this 2d-histogram plot that would greatly benefit from having a nice color legend:
https://github.com/go-hep/hep/tree/master/hplot#2d-histogram
h2d_plot

ctessum:
I'm thinking about splitting this into several more focused pull requests:
1. Creating a thumbnailer for palettes (for categorical data or discrete data bins), which could then be used in the normal plot.Legend.
2. Allowing plot.Legend to be rendered outside of the data canvas.
3. Creating a Plotter that renders legends for ColorMaps (for continuous data)

One outstanding question is where these things should live (other than 2, which is obvious). I would tend to think people would be looking for them in the palette package, but Dan has suggested the plotter package. Any other thoughts?

kortschak:
The palette package is not appropriate as that holds palette information and does not concern rendering in any way.

sbinet:
for me, `palettes` is related to colors and stuff. (ie: defining the set of colors)
I would be inclined to put thumbnailer for palettes and legends for ColorMaps inside `plotter` as well.

splitting the PR into 2/3 PRs is fine by me.

ctessum:
Closing this because all of the useful parts have been merged as other pull requests.

PR #333, opened 2016-12-18
Add draw.Split* functions

ctessum:
These are convenience functions that would be useful for instance in splitting up a canvas to put a plot in one part and a color bar in the other part. They are not strictly necessary but I feel like it's worthwhile to have them there because Crop can be pretty confusing to use.

### DIFF
--- a/vg/draw/split_horiz_example_test.go
+++ b/vg/draw/split_horiz_example_test.go
@@ -1,0 +1,25 @@
+// Copyright ©2017 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package draw_test
+
+import (
+	"fmt"
+
+	"github.com/gonum/plot/vg"
+	"github.com/gonum/plot/vg/draw"
+)
+
+// SplitHorizontal returns the left and right portions of c after splitting it
+// along a vertical line distance x from the left of c.
+func SplitHorizontal(c draw.Canvas, x vg.Length) (left, right draw.Canvas) {
+	return draw.Crop(c, 0, c.Min.X-c.Max.X+x, 0, 0), draw.Crop(c, x, 0, 0, 0)
+}
+
+func ExampleCrop_splitHorizontal() {
+	var c draw.Canvas
+	// Split c along a vertical line centered on the canvas.
+	left, right := SplitHorizontal(c, c.Size().X/2)
+	fmt.Println(left.Rectangle.Size(), right.Rectangle.Size())
+}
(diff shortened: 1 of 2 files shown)


kortschak:
I think this would be better as an example in the comment for Crop.

ctessum:
PTAL

vg/draw/draw_test.go
```diff
@@ -57,32 +58,40 @@ func TestCrop(t *testing.T) {
...
-	c := NewCanvas(&r, 6, 3)
-	lower, upper := SplitVertical(c, 1)
-	lowerRect := vg.Rectangle{Min: vg.Point{X: 0, Y: 0}, Max: vg.Point{X: 6, Y: 1}}
-	upperRect := vg.Rectangle{Min: vg.Point{X: 0, Y: 1}, Max: vg.Point{X: 6, Y: 3}}
-	if lowerRect != lower.Rectangle {
-		t.Errorf("lower: have %+v, want %+v", lower.Rectangle, lowerRect)
-	}
-	if upperRect != upper.Rectangle {
-		t.Errorf("upper: have %+v, want %+v", upper.Rectangle, upperRect)
+// This example splits a Canvas along a vertical line.
```
kortschak: If you place these each in a separate ..._example_test.go you can reference a package level func rather than using a func var. These don't need to have output comments, just have the code.

So you'll end up with something like, "split_horiz_example_test.go":

```
// Copyright ©2017 The gonum Authors. All rights reserved.

package draw_test

import (
	"fmt"

	"github.com/gonum/plot/vg"
	"github.com/gonum/plot/vg/draw"
)

// SplitHorizontal returns the left and right portions of c after splitting it
// along a vertical line distance x from the left of c.
func SplitHorizontal(c draw.Canvas, x vg.Length) (left, right draw.Canvas) {
	return draw.Crop(c, 0, c.Min.X-c.Max.X+x, 0, 0), draw.Crop(c, x, 0, 0, 0)
}

func ExampleCrop_splitHorizontal() {
	var c draw.Canvas
	// Split c along a vertical line centered on the canvas.
	left, right := SplitHorizontal(c, c.Size().X/2)
	fmt.Println(left.Rectangle.Size(), right.Rectangle.Size())
}
```

and the equivalent file for split vertical.

ctessum:
Done. This makes me nervous, though, because if at some point these examples become wrong it won't be obvious that that has happened.

kortschak:
I don't think you added the example files.

vg/draw/canvas.go
```diff
@@ -440,7 +454,7 @@ func (c *Canvas) StrokeLines(sty LineStyle, lines ...[]vg.Point) {
 // StrokeLine2 draws a line between two points in the given
 // Canvas.
 func (c *Canvas) StrokeLine2(sty LineStyle, x0, y0, x1, y1 vg.Length) {
-	c.StrokeLines(sty, []vg.Point{{x0, y0}, {x1, y1}})
+	c.StrokeLines(sty, []vg.Point{{X: x0, Y: y0}, {X: x1, Y: y1}})
```
kortschak: This is incidental, please revert it.

ctessum:
Apologies. PTAL.

PR #335, opened 2017-01-17
Moved vg syscall to separate file.

ctessum:
...so that it can be excluded in compilation environments that don't allow syscalls.

### DIFF
--- a/vg/font.go
+++ b/vg/font.go
@@ -12,9 +12,7 @@ package vg
 
 import (
 	"errors"
-	"go/build"
 	"io/ioutil"
-	"os"
 	"path/filepath"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


sbinet:
I am curious: how do you use `gonum/plot` w/ gopherjs? appgenine?

ctessum:
I'm currently trying to create a html ui for a model that I'm working on. It could also be interesting to create a gonum/plot "playground" where people could edit and run the plotter examples. The vgimg backend when transpiled to javascript is unbelievably slow. The svg backend works ok though.

sbinet:
could you squash the 2 commits into one? (thanks)

ctessum:
Sorry I got distracted.

PR #337, opened 2017-02-12
Changed plot axis padding to correctly handle rotated tick labels

ctessum:


eaburns:
The golden plot looks really nice. Thanks!

LGTM

Issue #343, opened 2017-03-21
Add Thumbnail function to bubbles.go


Bastiantheone:
You can't create legends for bubbles. This is the case because Bubbles.go doesn't have a thumbnail function.

Bastiantheone:
#344

PR #344, opened 2017-03-21
Add Thumbnail to bubbles.go. Solves #343

Bastiantheone:
I added a thumbnail function to bubbles.go. It shows a rectangle with the color of the bubble

### DIFF
--- a/plotter/bubbles.go
+++ b/plotter/bubbles.go
@@ -116,3 +116,15 @@ func (bs *Bubbles) GlyphBoxes(plt *plot.Plot) []plot.GlyphBox {
 	}
 	return boxes
 }
+
+// implementing the plot.Thumbnailer interface.
+func (bs *Bubbles) Thumbnail(c *draw.Canvas) {
+	pts := []vg.Point{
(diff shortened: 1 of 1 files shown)


ctessum:
One issue here is that to be the most useful, a legend would include information about what size bubbles correspond to what z-values. My guess is that difficulty was why the thumbnail wasn't implemented originally.

One option would be to do something like [this](https://github.com/gonum/plot/blob/master/plotter/sankey.go#L420), to have a function that returned mutliple thumbnailers for several different size bubbles, as well as labels specifying what z-value each bubble size represents.

Bastiantheone:
I created the issue, because I need a legend that explains which group each color bubble belongs. Would it be possible to add this function anyway?

eaburns:
It seems fine to me to add it. Better some info than none.

kortschak:
>One option would be to do something like this, to have a function that returned mutliple thumbnailers for several different size bubbles, as well as labels specifying what z-value each bubble size represents.

I like this.

Issue #208, opened 2015-06-07
Please change font which is Multilingual support.


axsann:
I'm Japanese.
NimbusRomNo9L-Regu.ttf is only support english.
Change the font please.

benoitmasson:
Hi,

I can't display asian characters either… what can we do about it, is it possible to enforce another font file?

A basic plot with title "aa 漢字 bb" gives me the following result, in Helvetica and Times-Roman fonts:
![chart_helv](https://cloud.githubusercontent.com/assets/10956459/24969048/b8fd9baa-1faf-11e7-96fa-c42a5dda73a0.png)
![chart_roman](https://cloud.githubusercontent.com/assets/10956459/24969049/b90cfa3c-1faf-11e7-8aea-a61bd289f6b9.png)

This is a serious issue I think… I'd be happy to help, but don't know where to start… do Liberation fonts contain these characters or is it an issue with support of these fonts in Go?

eaburns:
Fonts have been a constant annoyance since the beginning. See my comment on 29 Mar 2015 in #190 for a description of the problem.

That said, if you don't care about pdf or esp (which are decreasing in commonality the days — does anyone even know what eps is any more? did they ever?), then you can use whatever font you'd like. Though it's been years and I don't quite remember how. Maybe vg.AddFont? Hopefully someone else here can help.

benoitmasson:
Thanks for the answer, indeed I see the where the trouble is…

I'm only exporting to PNG for now (and yes I used a lot EPS, a long time ago :-) Supporting it may be questionable nowadays indeed, but I think PDF export has to be maintained… unless SVG is considered sufficient for vector graphics), so `vg.AddFont()` sounded like a good option. It did not work at once though, since exporting to PNG uses `vgimg.FillString()`, which in turn uses an internal font map…

So I hacked the `vgimg.go` file to add a `vgimg.AddFont()` function, which in the end calls the `vg.AddFont()` one (the other way around is not possible since `vgimg` already depends on `vg`, this would lead to an import cycle).

This is not very smart by if anybody is interested, I can submit a pull request for this patch, so that in the end `vgimg.AddFont()` does the job of adding a new font for texts in charts which are to be exported as images.

sbinet:
@benoitmasson I don't think that's the "correct" way to go.
`vgimg` should work with `vg` (and its fontmap), making sure the `vgimg` internal fontmap is synchronized with the one from `vg`.

[...]

PR #346, opened 2017-04-25
vg/vgimg: drop unneeded internal fontMap cache

sbinet:
draw2dimg doesn't make use of the informations provided by
draw2d.FontData fields (except for the Name field.)
At least it doesn't make use of them for the `FillString` part.

Fixes gonum/plot#208.


vg/vgimg/vgimg.go
```diff
@@ -17,12 +17,10 @@ import (
 	"image/png"
 	"io"
 
-	"golang.org/x/image/tiff"
```
kortschak: Please revert this aggregation of imports.

PR #348, opened 2017-04-27
vg: restore FontDirs initialization

sbinet:
PR gonum/plot#335 moved the vg-syscall into a different file to support
gopherjs.
But it dropped the initialization of the FontDirs global variable.
Fix this and also move the global const 'importString' closer to its
usage, in vg/font_syscall.go.

Add a test to make sure we don't regress.

### DIFF
--- a/vg/font_syscall.go
+++ b/vg/font_syscall.go
@@ -12,8 +12,15 @@ import (
 	"path/filepath"
 )
 
+const (
+	// importString is the import string expected for
+	// this package.  It is used to find the font
+	// directory included with the package source.
+	importString = "github.com/gonum/plot/vg"
+)
+
 func init() {
-	initFontDirs()
+	FontDirs = initFontDirs()
 }
 
 // InitFontDirs returns the initial value for the FontDirectories variable.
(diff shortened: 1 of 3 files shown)


vg/font_syscall.go
```diff
@@ -12,8 +12,15 @@ import (
...
 
+const (
+	// importString is the import string expected for
+	// this package.  It is used to find the font
+	// directory included with the package source.
+	importString = "github.com/gonum/plot/vg"
+)
+
 func init() {
-	initFontDirs()
```
kortschak: Oops.

vg/font_syscall_test.go
```diff
@@ -0,0 +1,20 @@
...
+package vg_test
+
+import (
+	"testing"
+
+	"github.com/gonum/plot/vg"
+)
+
+func TestVGFONTPATH(t *testing.T) {
+	if got := len(vg.FontDirs); got <= 0 {
```
kortschak: `got` cannot be negative; `len(vg.FontDirs) == 0`

and then `t.Fatal("zero length FontDirs")`

sbinet: done.

vg/font_syscall.go
sbinet: yep. I am bad at reviewing PRs :/
it completely passed me at the time.

PR #349, opened 2017-04-28
Travis add golang 1.8.1

ctessum:


sbinet:
perhaps we should just put `1.8.x` as per:
https://docs.travis-ci.com/user/languages/go/#Specifying-a-Go-version-to-use

```
You can use any tagged version of Go, a version with x in place of the minor or patch level to use the latest for a given major or minor version
```

Kunde21:
We should also add `master` and allow that one to fail.  That lets us test against tip to find any regressions without flagging our build.

kortschak:
@Kunde21 That is an excellent idea. I didn't know you could do that.

ctessum:
Done. PTAL

kortschak:
go1.5 can be dropped; we only support three releases.

ctessum:
Done. Is there some sort of documentation that says these things?

kortschak:
No, I don't think there is. We should add that.

PR #350, opened 2017-04-28
Added FromPalette() and New() to palette package

ctessum:
These functions allow converting from Palettes to ColorMaps and creating palettes from a list of colors, respectively.

### DIFF
--- a/palette/colormap.go
+++ b/palette/colormap.go
@@ -1,0 +1,150 @@
+// Copyright ©2017 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package palette
+
+import (
+	"fmt"
+	"image/color"
+	"math"
+
+	"github.com/gonum/floats"
+)
+
+// FromPalette creates a ColorMap that maps values to colors based
+// on RGB interpolation among the colors in p.
+func FromPalette(p Palette) ColorMap {
+	scalars := make([]float64, len(p.Colors()))
+	return &colorMap{
+		colors:  p.Colors(),
+		scalars: floats.Span(scalars, 0, 1),
+		alpha:   1,
+	}
+}
+
+type colorMap struct {
+	// colors are the colors in the underlying Palette.
+	colors []color.Color
+
+	// scalars are len(colors) equally spaced points between
+	// 0 and 1, used for interpolation.
+	scalars []float64
+
+	// alpha represents the opacity of the returned
+	// colors in the range (0,1). It is set to 1 by default.
+	alpha float64
+
+	// min and max are the minimum and maximum values of the range of scalars
+	// that can be mapped to colors using this ColorMap.
+	min, max float64
+}
+
+// At implements the ColorMap interface for a colorMap value.
+func (cm *colorMap) At(v float64) (color.Color, error) {
+	if err := checkRange(cm.min, cm.max, v); err != nil {
+		return nil, err
+	}
+	scalar := (v - cm.min) / cm.max
+	i := searchFloat64s(cm.scalars, scalar)
+	if i == 0 {
+		return cm.colors[i], nil
+	}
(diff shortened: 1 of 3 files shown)


ctessum:
PTAL

kortschak:
I'm uneasy about this, but I need to be in a meeting now. I will respond further when I have time.

ctessum:
They would both be necessary to avoid circular imports when testing legends if the legends were in the palette package, but since the legends are going in the plotter package these functions may not be as important. I'm not super attached to them, but it does seem like they could be useful in some cases.

palette/colormap.go
```diff
@@ -0,0 +1,150 @@
...
+	"fmt"
+	"image/color"
+	"math"
+
+	"github.com/gonum/floats"
+)
+
+// FromPalette creates a ColorMap that maps values to colors based
+// on RGB interpolation among the colors in p.
+func FromPalette(p Palette) ColorMap {
```
kortschak: This name is not good. It will be called as `palette.FromPalette(p)` which is misleading: `NewColorMap`. It's also odd that you take a `Palette` and then immediately extract the colours. If that's what you want, make the signature `NewColorMap(c []color.Color) ColorMap`.

palette/colormap.go
```diff
@@ -0,0 +1,150 @@
...
+	scalars := make([]float64, len(p.Colors()))
+	return &colorMap{
+		colors:  p.Colors(),
+		scalars: floats.Span(scalars, 0, 1),
+		alpha:   1,
+	}
+}
+
+type colorMap struct {
```
kortschak: Is there a reason this should not be exported?

palette/colormap.go
```diff
@@ -0,0 +1,150 @@
...
+
+// At implements the ColorMap interface for a colorMap value.
+func (cm *colorMap) At(v float64) (color.Color, error) {
+	if err := checkRange(cm.min, cm.max, v); err != nil {
+		return nil, err
+	}
+	scalar := (v - cm.min) / cm.max
+	i := searchFloat64s(cm.scalars, scalar)
+	if i == 0 {
+		return cm.colors[i], nil
```
kortschak: `return cm.colors[0], nil`

palette/palette.go
```diff
@@ -20,6 +20,11 @@ type Palette interface {
 	Colors() []color.Color
 }
 
+// New returns a new palette from the specified colors
```
kortschak: This is not necessary for this PR.

palette/colormap_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"image/color"
+	"math"
+	"strings"
+	"testing"
+)
+
+func TestColorMap_Range(t *testing.T) {
+	p := FromPalette(New(color.White, color.Black))
```
kortschak: `m := NewColorMap([]color.Color{color.White, color.Black})`

s/p/m/g

palette/colormap.go
ctessum: To minimize the "weight of things". With the current ColorMap interface it doesn't contain anything the user needs to interact with directly. Under proposal #352 it would need to be exported though.

ctessum:
One other thing to discuss is whether this is actually a good idea. It would allow Heat, Rainbow, and Radial to be used as color maps, but in my mind it would be more appropriate directly reimplement those functions as color maps instead of palettes (because their main purpose is for use with continuous data), or just delete them. 

It could be useful to be able to convert the Brewer DivergingPalettes and SequentialPalettes to color maps, but I'm not sure it would be useful enough to justify including this (or it make make more sense to put it in the brewer package).

palette/colormap.go
kortschak: OK

kortschak:
I like the idea of Heat, RainBow and Radial being ColorMaps.

ctessum:
Closing because it is not clear that this is useful.

PR #351, opened 2017-04-29
Add thumbnailer for palette.Palette

ctessum:


ctessum:
PTAL. The legend in the example image doesn't look very good but I think that might be a task for a separate pull request.

plotter/heat_test.go
```diff
@@ -58,6 +60,16 @@ func ExampleHeatMap() {
...
+	// Create a legend.
+	rng := h.Max - h.Min
+	thumbs := PaletteThumbnailers(plte)
+	for i := len(thumbs) - 1; i >= 0; i-- {
+		t := thumbs[i]
+		fmin := float64(i) / float64(len(thumbs))
+		fmax := float64(i+1) / float64(len(thumbs))
+		p.Legend.Add(fmt.Sprintf("%.2g - %.2g", h.Min+fmin*rng, h.Min+fmax*rng), t)
+	}
+
```
Kunde21: This for loop seems like a good candidate for incorporation into the API as `BuildThumbnail` and (if so, additionally) incorporated into an `AddThumbnail` method.

ctessum: Because this example seems to be representing a continuous rather  than categorical variable, I feel like a better long term solution here is to use a color map instead of a palette. That would also fix the Too Many Labels problem.

plotter/heat_test.go
```diff
@@ -48,7 +49,8 @@ func ExampleHeatMap() {
 			5, 6, 7, 8,
 			9, 10, 11, 12,
 		})}
-	h := NewHeatMap(m, palette.Heat(12, 1))
+	plte := palette.Heat(12, 1)
```
kortschak: s/plte/pal/g

plotter/palettethumbnailer.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"image/color"
+
+	"github.com/gonum/plot"
+	"github.com/gonum/plot/palette"
+	"github.com/gonum/plot/vg"
+	"github.com/gonum/plot/vg/draw"
+)
+
+// PaletteThumbnailers creates a group of objects that can be used to
```
kortschak: s/group of objects/slice of plot.Thumbnailer/

plotter/palettethumbnailer.go
```diff
@@ -0,0 +1,67 @@
...
+	return thumbnailers
+}
+
+// paletteThumbnailer implements the Thumbnailer interface
+// for color palettes.
+type paletteThumbnailer struct {
+	color color.Color
+}
+
+// Thumbnail fulfills the plot.Thumbnailer interface.
```
kortschak: s/fulfills/satisfies/

plotter/palettethumbnailer.go
```diff
@@ -0,0 +1,67 @@
...
+	pts := []vg.Point{
+		{X: c.Min.X, Y: c.Min.Y},
+		{X: c.Min.X, Y: c.Max.Y},
+		{X: c.Max.X, Y: c.Max.Y},
+		{X: c.Max.X, Y: c.Min.Y},
+	}
+	poly := c.ClipPolygonY(pts)
+	c.FillPolygon(t.color, poly)
+
+	// Here we draw the upper border.
```
kortschak: Leave out the borders please.

plotter/palettethumbnailer.go
```diff
@@ -0,0 +1,67 @@
...
+// Thumbnail fulfills the plot.Thumbnailer interface.
+func (t paletteThumbnailer) Thumbnail(c *draw.Canvas) {
+	lineStyle := draw.LineStyle{
+		Color:    color.Black,
+		Width:    vg.Points(1),
+		Dashes:   []vg.Length{},
+		DashOffs: 0,
+	}
+
+	// Here we draw the fill.
```
kortschak: Delete this comment.

plotter/heat_test.go
```diff
@@ -58,6 +60,16 @@ func ExampleHeatMap() {
...
 	p.Add(h)
 
+	// Create a legend.
+	rng := h.Max - h.Min
+	thumbs := PaletteThumbnailers(plte)
+	for i := len(thumbs) - 1; i >= 0; i-- {
+		t := thumbs[i]
+		fmin := float64(i) / float64(len(thumbs))
+		fmax := float64(i+1) / float64(len(thumbs))
+		p.Legend.Add(fmt.Sprintf("%.2g - %.2g", h.Min+fmin*rng, h.Min+fmax*rng), t)
```
kortschak: If you have the following, you get a less cluttered legend and the issue of borders becomes cleaner.

```
		t := thumbs[i]
		if i == 0 || i == len(thumbs)-1 {
			p.Legend.Add("", t)
			continue
		}
		fmin := float64(i) / float64(len(thumbs))
		fmax := float64(i+1) / float64(len(thumbs))
		p.Legend.Add(fmt.Sprintf("%.2g - %.2g", h.Min+fmin*rng, h.Min+fmax*rng), t)
```

ctessum: It looks like the only change here is to not label the first and last colors. How does that help?

kortschak: Sorry, that was a terrible misspeak.

```
		t := thumbs[i]
		if i != 0 && i != len(thumbs)-1 {
			p.Legend.Add("", t)
			continue
		}
		switch i {
		case 0:
			val = h.Max
		case len(thumbs)-1:
			val = h.Min
		}
		p.Legend.Add(fmt.Sprintf("%.2g", val), t)
```

kortschak:
What would it involve to get the legend to not overlay the heat map?

ctessum:
PTAL. I figured out a way to get the legend to not overlap. It's pretty hacky but it works for now.

PR #353, opened 2017-05-01
plotter: Add ColorBar legend for palette.ColorMap

ctessum:


ctessum:
PTAL

plotter/colorbar.go
```diff
@@ -0,0 +1,107 @@
...
+
+	// Colors specifies the number of colors to be
+	// shown in the legend. The default is 255.
+	Colors int
+
+	cm palette.ColorMap
+}
+
+// NewColorBar creates a new ColorBar plotter.
+func NewColorBar(cm palette.ColorMap) *ColorBar {
```
kortschak: I think we could do without a New function if we made `ColorBar` usable with a non-zero `palette.ColorMap`. This would entail defining the behaviour for `Colors == 0`, probably to be one color per point across the varying axis.

plotter/colorbar.go
```diff
@@ -0,0 +1,107 @@
...
+		img = image.NewNRGBA64(image.Rectangle{
+			Min: image.Point{X: 0, Y: 0},
+			Max: image.Point{X: 1, Y: l.Colors},
+		})
+		for i := 0; i < l.Colors; i++ {
+			color, err := l.cm.At(float64(i) / float64(l.Colors-1))
+			if err != nil {
+				panic(err)
+			}
+			if l.Vertical {
```
kortschak: I don't understand this. We know `l.Vertical` is true here. The conditional code is a reversal, not a horizontal v vertical rendering.

plotter/colorbar.go
```diff
@@ -0,0 +1,107 @@
...
+	if l.Vertical {
+		return 0, 1, l.cm.Min(), l.cm.Max()
+	}
+	return l.cm.Min(), l.cm.Max(), 0, 1
+}
+
+// GlyphBoxes implements the GlyphBoxes method
+// of the plot.GlyphBoxer interface.
+func (l *ColorBar) GlyphBoxes(plt *plot.Plot) []plot.GlyphBox {
+	return nil
```
kortschak: Why?

plotter/colorbar.go
```diff
@@ -0,0 +1,107 @@
...
+	return &ColorBar{
+		cm:     cm,
+		Colors: 255,
+	}
+}
+
+// Plot implements the Plot method of the plot.Plotter interface.
+func (l *ColorBar) Plot(c draw.Canvas, p *plot.Plot) {
+	if l.cm.Max() == l.cm.Min() {
+		panic("palette: ColorMap Max==Min")
```
sbinet: this should read `plotter: ...`

sbinet: perhaps check this invariant inside `NewColorBar` instead of inside each and every method of `ColorBar`.

plotter/colorbar.go
```diff
@@ -0,0 +1,107 @@
...
+		Min: vg.Point{X: xmin, Y: ymin},
+		Max: vg.Point{X: xmax, Y: ymax},
+	}
+	c.DrawImage(rect, img)
+}
+
+// DataRange implements the DataRange method
+// of the plot.DataRanger interface.
+func (l *ColorBar) DataRange() (xmin, xmax, ymin, ymax float64) {
+	if l.cm.Max() == l.cm.Min() {
```
sbinet: perhaps check this invariant inside `NewColorBar` instead of inside each and every method of `ColorBar`.

plotter/colorbar.go
ctessum: This is mutually exclusive with getting rid of NewColorBar.

plotter/colorbar.go
ctessum: This is mutually exclusive with getting rid of NewColorBar.

ctessum:
PTAL

plotter/colorbar.go
```diff
@@ -0,0 +1,111 @@
...
+	// shown in the legend. If Colors is not specified,
+	// a default will be used.
+	Colors int
+}
+
+// colors returns the number of colors to be shown
+// in the legend, substituting invalid values
+// with the default of one color per vg.Point.
+func (l *ColorBar) colors(c draw.Canvas) int {
+	if l.Colors <= 0 {
```
kortschak: Reduce indentation by saying 

```
if l.Colors > 0 {
    return l.colors
}
if l.Vertical {
    return int((c.Max.Y - c.Min.Y).Points())
}
return int((c.Max.X - c.Min.X).Points())
```

Issue #239, opened 2016-01-21
plot: ability to display 2 (or n) connected plots


sbinet:
it should be possible to display multiple plots on the same canvas (or gather multiple canvases on the same image?).
it should also be possible to "connect" the axes of these plots so they are properly aligned:

http://matplotlib.org/examples/axes_grid/simple_axesgrid.html
http://matplotlib.org/examples/axes_grid/simple_axesgrid2.html
http://matplotlib.org/examples/pylab_examples/axes_zoom_effect.html
http://matplotlib.org/examples/pylab_examples/finance_work2.html
http://matplotlib.org/examples/pylab_examples/ganged_plots.html
http://matplotlib.org/examples/pylab_examples/scatter_hist.html
http://matplotlib.org/examples/pylab_examples/subplots_demo.html

eaburns:
Also, see #151.

sbinet:
ah, sorry. my queries on the gonum-plot issue tracker failed me.

sbinet:
@ebusato

ctessum:
Closed via #354.

Issue #151, opened 2015-03-15
Alignment of axes between multiple plots


eaburns:
Original [issue 124](https://code.google.com/p/plotinum/issues/detail?id=124) created by eaburns on 2013-05-01T13:26:55.000Z:

I have two plots with different y axes but with the same x axis. I want to display one plot above the other. The problem is that the spacing differs slightly so the axes aren't perfectly aligned.

I don't see any way to solve this with the existing interface.

One way to solve it is to add an option to set the size of an axis to a fixed value. Basically, add a variable like this to the Axis struct:

```
// If nonzero, the axis will occupy a fixed width regardless of the
// content. This is useful if you need the axes to align between
// multiple plots.
FixedSize vg.Length
```

Then you can set the same FixedSize value for each of your plots and they will always align.

I've attached a patch which does this and solves the issue for me.

brunoqc:
@eaburns by the way you can use "```"  around your code to have nice monospace formatting and syntax highlighting if you specify the "go" keyword.

    ```go
    <your code>
    ```

will render as:

```go
/*
Build a bunch of plots and align their x axes.
*/
var plots []plot.Plot
// build plots
for _, p := range plots {
    // AlignX aligns p's x axis with the other plots x axes.
    p.AlignX(plots...)
}
```

eaburns:
Yeah, sorry, I often reply from my cell phone and don't think about that.

ctessum:
Closed via #354

PR #354, opened 2017-05-02
plot: add Align function

ctessum:
Responds to #239 and #151


align.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/gonum/plot/vg"
+	"github.com/gonum/plot/vg/draw"
+)
+
+// Align returns a two-dimensional row-major array of Canvases which will
+// produce tiled plots with DataCanvases that are evenly sized and spaced.
+// The arguments to the function are a two-dimensional row-major array
+// of plots, a tile configuration, and the canvas to which the tiled
+// plots are to be drawn.
+func Align(plots [][]*Plot, t draw.Tiles, dc draw.Canvas) [][]draw.Canvas {
```
kortschak: s/Align/Facet/g?

ctessum: I know "facet" is the word ggplot uses to describe these, so it could be useful to maintain continuity with that, but on the other hand "facet" is really referring to the data that's being plotted, whereas what this function is actually doing is "align"ing. My slight preference is "align" but I would be fine with either one.

kortschak: Yeah, fair enough.

align.go
```diff
@@ -0,0 +1,115 @@
...
+	o := make([][]draw.Canvas, len(plots))
+
+	// Create the initial tiles.
+	for j := 0; j < t.Rows; j++ {
+		o[j] = make([]draw.Canvas, len(plots[j]))
```
kortschak: Check `plots` matchs the dimensionality of `t`.

align.go
```diff
@@ -0,0 +1,115 @@
...
+		p, n float64
+	}
+	xSpacing := make([]posNeg, t.Cols)
+	ySpacing := make([]posNeg, t.Rows)
+
+	// Calculate the maximum spacing between data canvases
+	// for each row and column.
+	for j, row := range plots {
+		for i, p := range row {
+			c := o[j][i]
```
kortschak: I think it's reasonable to have a `nil` `plot.Plot` in here. Check this and skip if `nil`.

align.go
```diff
@@ -0,0 +1,115 @@
...
+			ySpacing[j].p = math.Max(float64(c.Max.Y-dataC.Max.Y), ySpacing[j].p)
+		}
+	}
+
+	// Adjust the horizontal and vertical spacing between
+	// canvases to match the maximum for each column and row,
+	// respectively.
+	for j, row := range plots {
+		for i, p := range row {
+			c := o[j][i]
```
kortschak: Skip if `p == nil`.

align.go
```diff
@@ -0,0 +1,115 @@
...
+
+	avgWidth := vg.Length((float64(dc.Max.X-dc.Min.X) - xTotalSpace) / float64(t.Cols))
+	avgHeight := vg.Length((float64(dc.Max.Y-dc.Min.Y) - yTotalSpace) / float64(t.Rows))
+
+	// Adjust canvases so the width of each DataCanvas is the
+	// same.
+	for j := 0; j < t.Rows; j++ {
+		var moveHorizontal vg.Length
+		for i := 0; i < t.Cols; i++ {
+			c := o[j][i]
```
kortschak: Skip if `plots[j][i] == nil`.

align.go
```diff
@@ -0,0 +1,115 @@
...
+	for _, s := range ySpacing {
+		yTotalSpace += s.n + s.p
+	}
+
+	avgWidth := vg.Length((float64(dc.Max.X-dc.Min.X) - xTotalSpace) / float64(t.Cols))
+	avgHeight := vg.Length((float64(dc.Max.Y-dc.Min.Y) - yTotalSpace) / float64(t.Rows))
+
+	// Adjust canvases so the width of each DataCanvas is the
+	// same.
+	for j := 0; j < t.Rows; j++ {
```
kortschak: Why not range over `plots`?

Can you merge these two passes into one?

align.go
```diff
@@ -0,0 +1,115 @@
...
+				0,
+				0,
+			)
+			moveHorizontal += avgWidth - width
+		}
+	}
+
+	// Adjust canvases so the height of each DataCanvas
+	// is the same.
+	for i := 0; i < t.Cols; i++ {
```
kortschak: Range over `plots`.

align.go
sbinet: what about returning a, say, `Canvases` type instead of this slice of slices?
this way, we could attach methods to it (`At(i,j int)` comes to mind) and implement e.g. `Thumbnailer`, `Plotter`, ...

one could then create these type of plots more easily:
- http://www.cookbook-r.com/Graphs/Scatterplot/
- http://www.cookbook-r.com/Graphs/Scatterplot/figure/unnamed-chunk-4-2.png

that's more or less what I did here:
- https://github.com/go-hep/hep/tree/master/hplot#tiles-of-1d-histograms
- https://github.com/go-hep/hep/blob/master/hplot/tiledplot.go#L22

ctessum: If I understand right, the main reason for doing this would be to avoid doing the final nested loop to draw the plots to the canvases, and also probably an additional nested loop to create the original matrix of plots. I'm guessing that @kortschak would make a "weight of things" argument that the added convenience would not be worth the added API surface. 

Another option could be to take a 1-d slice of plots and return a 1-d array of canvases, which would mean that the user wouldn't have to go through the nested loops, but they would have to do the 1-d to 2-d array index conversion on their own. I'm not sure whether that would make things easier overall or not. My general feeling is that array index conversions tend to be more confusing than nested loops.

I do like the idea of the idea of the TiledPlot, though, or maybe something like a "MultiPlotter" interface that could allow the automatic creation of things like the covariance matrices you linked to. My feeling is that that would be something that would be built on top of this PR rather than built into it, but let me know what you think.

ctessum:
PTAL. I'm kind of torn between the current function format and something like @sbinet suggested:

``` go
type Tiles struct {
  Plots []*Plot
  Tiles draw.Tiles

  canvases []draw.Canvas // not exported
}

func (t *Tiles) At(c draw.Canvas, i, j int) draw.Canvas {
 ...
}

func (t *Tiles) Draw(c draw.Canvas) {
 ...
}

func (t *Tiles) Save(w, h vg.Length, file string) (err error) {
 ...
}
```

Let me know if there are any thoughts.

eaburns:
sorry for jumping in late with little context, but does this align padding
across subplots? in my opinion, the hardest part is aligning padding such
that the axes line up even if glyphs near the edge of one plot push the
axes around.

ctessum:
It aligns the DataCanvases within rows and columns, and makes sure all DataCanvases are the same size. The axes won't necessarily line up in the case mentioned by @eaburns, although the DataCanvases will.

In my mind the main reasons for doing this is so that the axes don't need to be labeled for all subplots, and so that data can be compared among subplots within a row or column. I think this accomplishes both those things. 

I agree it would be better to also align the glyph padding, but it seems like it would be difficult to do that without adding a substantial amount of complexity to the existing Plot code.

eaburns:
But the problem is that if you don't align the glyph padding then the axes
won't align correctly and not drawing all axes would be misleading.

ctessum:
If I understand the issue correctly, I don't think that's true. I added an example to the test image where one subplot has different axis padding than the rest (column 2, row 0). As you can see, the Y axis of that subplot aligns with the Y axes of the other plots in the same row, and the X axis of that subplot aligns with the X axes of the other plots in the same column, which I think is what's important.

eaburns:
I'm taking about the passing added for glyphboxes. If you are referring to the golden plot, I don't see any glyphs, so I'm not sure how one of the plots could include this padding. I don't remember the code well, so I could very well, be wrong, but if you didn't handle glyph boxes, then there should be a bug.

ctessum:
Since [glyphboxes are accounted for when calculating the DataCanvas](https://github.com/gonum/plot/blob/master/plot.go#L193), I don't think there is a problem. It would be cumbersome to check because I can't add anything from plot/plotter to the test case without causing an import loop.

What I was talking about in the golden image is that I removed the default padding from the axes in one of the plots, causing the axes to be misaligned but not in a way that would affect any conclusions being made about the data.

ctessum:
Ping?

kortschak:
Why was this close? @sbinet has approved it.

ctessum:
I merged it first, at least I thought I did. It looks like it's on there: https://godoc.org/github.com/gonum/plot#Align. I'm not sure why it doesn't show up here.

kortschak:
There are no changes in this PR. At some point you will have rebased onto master or something. We will never know.

PR #360, opened 2017-05-16
Add IntTicks

isa9660:
Nice to meet you. Thank you for a wonderful library.
In order to easily represent integers, I created 'IntTicks'.
By the way, if you change 'DefaultTicks' in 'Custom tick marks' of example to 'IntTicks', it will have the same result as the example.
('1,.5e,+04'->'15,000','1e,+04'->'10,000)

### DIFF
--- a/axis.go
+++ b/axis.go
@@ -5,6 +5,7 @@
 package plot
 
 import (
+	"fmt"
 	"image/color"
 	"math"
 	"strconv"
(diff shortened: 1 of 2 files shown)


DavidGamba:
Take a look at https://github.com/gonum/plot/pull/362.
With it, you should be able to do:

````
    type myCustomFormatTicks struct{}

    var _ plot.Ticker = myCustomFormatTicks{}

    func (myCustomFormatTicks) Ticks(min, max float64, format func(v float64, prec int) string) (ticks []plot.Tick) {
      return plot.DefaultTicks{}.Ticks(min, max, myFormatFloatTick)
    }

    func myFormatFloatTick(v float64, prec int) string {
           return strconv.FormatFloat(floats.Round(v, prec), 'f', 0, 64)
    }

    p.Y.Tick.Marker = myCustomFormatTicks{}
````
And get the same results.

isa9660:
I had planned from before.
Even now I can do it custom, so I am not in trouble,
I will close it. Thank you very much!

PR #361, opened 2017-05-19
plotter: allow user control of NaN values

### LINKED ISSUE #302: Creating a heatmap with 0 depth (min value == max value) causes panic

kortschak:
Also use this to allow plotting of ambiguously colour-mapping values
arising from when the Z-range is zero.

Fixes #302.

@eaburns @btracey @jasonpfox Please take a look.

### DIFF
--- a/plotter/heat.go
+++ b/plotter/heat.go
@@ -49,6 +49,11 @@ type HeatMap struct {
 	Underflow color.Color
 	Overflow  color.Color
 
+	// NaN is the color used to fill heat map elements
+	// that are NaN or do not map to a unique palette
+	// color.
+	NaN color.Color
+
 	// Min and Max define the dynamic range of the
 	// heat map.
 	Min, Max float64
@@ -57,8 +62,8 @@ type HeatMap struct {
 // NewHeatMap creates as new heat map plotter for the given data,
 // using the provided palette. If g has Min and Max methods that return
 // a float, those returned values are used to set the respective HeatMap
-// fields. If the returned HeatMap is used when Min is greater than or
-// equal to Max, the Plot method will panic.
+// fields. If the returned HeatMap is used when Min is greater than Max,
+// the Plot method will panic.
 func NewHeatMap(g GridXYZ, p palette.Palette) *HeatMap {
 	var min, max float64
 	type minMaxer interface {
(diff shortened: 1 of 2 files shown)


plotter/heat.go
```diff
@@ -93,7 +98,7 @@ func NewHeatMap(g GridXYZ, p palette.Palette) *HeatMap {
 
 // Plot implements the Plot method of the plot.Plotter interface.
 func (h *HeatMap) Plot(c draw.Canvas, plt *plot.Plot) {
-	if h.Min >= h.Max {
+	if h.Min > h.Max {
 		panic("heatmap: non-positive Z range")
```
vladimir-ch: negative Z range

kortschak: Done.

plotter/heat.go
```diff
@@ -108,35 +113,27 @@ func (h *HeatMap) Plot(c draw.Canvas, plt *plot.Plot) {
 	var pa vg.Path
 	cols, rows := h.GridXYZ.Dims()
 	for i := 0; i < cols; i++ {
-
 		var right, left float64
 		switch i {
 		case 0:
-			right = (h.GridXYZ.X(i+1) - h.GridXYZ.X(i)) / 2
+			right = (h.GridXYZ.X(1) - h.GridXYZ.X(0)) / 2
```
sbinet: nit: `x/2` -> `0.5 * x`

kortschak: Performance is not an issue here and `/ 2` is easier to read. I'd prefer to leave it if that's OK.

sbinet: depends on the number of rows and cols, but ok.

jasonpfox:
Looks like a good solution to allow control when desired without a panic.

Am I correct in that if heatmap.NaN is not set by the user after NewHeatmap is called then filling the heatmap is a no color noop?

jasonpfox:
On a related note to the .NaN that I just noticed, should we be setting .Underflow and  .Overflow by default to the min and max palette values vice blank areas?  Typically in other applications and usages a saturated heatmap cell shows the values at the extent of the defined color range for values outside of the defined range and I just noticed that we were not  doing that, which explains some behavior I thought was odd that I have seen from time to time.

kortschak:
Yes, that is correct.

kortschak:
*ping*

plotter/heat.go
vladimir-ch: The panic message should say negative, the condition and the changes in this PR allow empty Z range. I'll then approve.

kortschak: I had actually done that, and then not pushed.

Actually done now.

PR #362, opened 2017-05-25
fixes-336: Extend the Ticker interface to allow passing a format function

DavidGamba:
This allows for easy wrappers around the plot.DefaultTicks struct with
custom format functions:

````
    type myCustomFormatTicks struct{}

    var _ plot.Ticker = myCustomFormatTicks{}

    func (myCustomFormatTicks) Ticks(min, max float64, format func(v float64, prec int) string) (ticks []plot.Tick) {
      return plot.DefaultTicks{}.Ticks(min, max, myFormatFloatTick)
    }

    func myFormatFloatTick(v float64, prec int) string {
           return strconv.FormatFloat(floats.Round(v, prec), 'g', 8, 64)
    }

    p.Y.Tick.Marker = myCustomFormatTicks{}
````

### DIFF
--- a/axis.go
+++ b/axis.go
@@ -21,7 +21,7 @@ const displayPrecision = 4
 // Ticker creates Ticks in a specified range
 type Ticker interface {
 	// Ticks returns Ticks in a specified range
-	Ticks(min, max float64) []Tick
+	Ticks(min, max float64, format func(v float64, prec int) string) []Tick
 }
 
 // Normalizer rescales values from the data coordinate system to the
@@ -200,7 +200,7 @@ func (a *horizontalAxis) size() (h vg.Length) {
 		h -= a.Label.Font.Extents().Descent
 		h += a.Label.Height(a.Label.Text)
 	}
-	if marks := a.Tick.Marker.Ticks(a.Min, a.Max); len(marks) > 0 {
+	if marks := a.Tick.Marker.Ticks(a.Min, a.Max, nil); len(marks) > 0 {
 		if a.drawTicks() {
 			h += a.Tick.Length
 		}
@@ -220,7 +220,7 @@ func (a *horizontalAxis) draw(c draw.Canvas) {
 		y += a.Label.Height(a.Label.Text)
 	}
 
-	marks := a.Tick.Marker.Ticks(a.Min, a.Max)
+	marks := a.Tick.Marker.Ticks(a.Min, a.Max, nil)
 	ticklabelheight := tickLabelHeight(a.Tick.Label, marks)
 	for _, t := range marks {
 		x := c.X(a.Norm(t.Value))
@@ -254,7 +254,7 @@ func (a *horizontalAxis) draw(c draw.Canvas) {
 
 // GlyphBoxes returns the GlyphBoxes for the tick labels.
 func (a *horizontalAxis) GlyphBoxes(*Plot) (boxes []GlyphBox) {
-	for _, t := range a.Tick.Marker.Ticks(a.Min, a.Max) {
+	for _, t := range a.Tick.Marker.Ticks(a.Min, a.Max, nil) {
 		if t.IsMinor() {
 			continue
 		}
@@ -278,7 +278,7 @@ func (a *verticalAxis) size() (w vg.Length) {
 		w -= a.Label.Font.Extents().Descent
 		w += a.Label.Height(a.Label.Text)
 	}
-	if marks := a.Tick.Marker.Ticks(a.Min, a.Max); len(marks) > 0 {
+	if marks := a.Tick.Marker.Ticks(a.Min, a.Max, nil); len(marks) > 0 {
 		if lwidth := tickLabelWidth(a.Tick.Label, marks); lwidth > 0 {
 			w += lwidth
 			w += a.Label.Width(" ")
@@ -302,7 +302,7 @@ func (a *verticalAxis) draw(c draw.Canvas) {
 		c.FillText(sty, vg.Point{X: x, Y: c.Center().Y}, a.Label.Text)
 		x += -a.Label.Font.Extents().Descent
 	}
-	marks := a.Tick.Marker.Ticks(a.Min, a.Max)
+	marks := a.Tick.Marker.Ticks(a.Min, a.Max, nil)
 	if w := tickLabelWidth(a.Tick.Label, marks); len(marks) > 0 && w > 0 {
 		x += w
 	}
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: gob/gob_test.go [hunk 1]
```diff
@@ -129,8 +129,8 @@ func randomPoints(n int, rnd *rand.Rand) plotter.XYs {
 // into the labels for the major tick marks.
 type commaTicks struct{}
 
-func (commaTicks) Ticks(min, max float64) []plot.Tick {
-	tks := plot.DefaultTicks{}.Ticks(min, max)
+func (commaTicks) Ticks(min, max float64, format func(v float64, prec int) string) []plot.Tick {
+	tks := plot.DefaultTicks{}.Ticks(min, max, nil)
```
sbinet: you should probably weave the `format` arg through the call of `DefaultTicks`, instead of dropping it on the floor

sbinet: actually, I don't understand why weaving `format` through *all* the calls to `.Ticks(...)` isn't done.

axis.go
```diff
@@ -21,7 +21,7 @@ const displayPrecision = 4
 // Ticker creates Ticks in a specified range
 type Ticker interface {
 	// Ticks returns Ticks in a specified range
```
sbinet: if the interface is modified, the new argument should also be documented.

axis_test.go
```diff
@@ -42,7 +42,7 @@ func TestAxisSmallTick(t *testing.T) {
 			want: []string{"0.0001", "0.00011", "0.00012", "0.00013", "0.00014"},
 		},
 	} {
-		ticks := d.Ticks(test.min, test.max)
+		ticks := d.Ticks(test.min, test.max, nil)
```
sbinet: you should also add a test with the new format argument

DavidGamba:
@sbinet please take another look.

axis.go
```diff
@@ -200,7 +202,7 @@ func (a *horizontalAxis) size() (h vg.Length) {
 		h -= a.Label.Font.Extents().Descent
 		h += a.Label.Height(a.Label.Text)
 	}
-	if marks := a.Tick.Marker.Ticks(a.Min, a.Max); len(marks) > 0 {
+	if marks := a.Tick.Marker.Ticks(a.Min, a.Max, nil); len(marks) > 0 {
```
sbinet: ok. so how would I use my own format here?

shouldn't `Axis` grow another field that could be passed there?
something like:

```go
type Axis struct {
   // ...
   Tick struct {
      // ...
      // Marker returns the tick marks.  Any tick marks
      // returned by the Marker function that are not in
      // range of the axis are not drawn.
      Marker Ticker

      // <doc for Format>
      Format func(v float64, prec int) string
   }
}
```

and then:

```go
if marks := a.Tick.Marker.Ticks(a.Min, a.Max, a.Tick.Format); len(marks) > 0 {
```

sbinet:
@kortschak @eaburns any feedback about this change?

DavidGamba:
@sbinet please take another look. I apologize for the delay...
With this, to provide a format to the default ticks, all I have to do is:

````golang
...
	p, err := plot.New()
	if err != nil {
		return err
	}
	p.X.Tick.Format = myAxisTickFormat
	p.Y.Tick.Format = myAxisTickFormat
...

func myAxisTickFormat(v float64, prec int) string {
	return strconv.FormatFloat(floats.Round(v, prec), 'f', 0, 64)
}
````

File: axis.go [hunk 5]
```diff
@@ -355,11 +361,14 @@ type DefaultTicks struct{}
 var _ Ticker = DefaultTicks{}
 
 // Ticks returns Ticks in a specified range
-func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
+func (DefaultTicks) Ticks(min, max float64, format func(v float64, prec int) string) (ticks []Tick) {
 	const SuggestedTicks = 3
 	if max < min {
 		panic("illegal range")
 	}
+	if format == nil {
```
sbinet: instead of sprinkling `DefaultTicks` and `LogTicks` with these `if format == nil` checks, perhaps we could instead put this default inside `makeAxis` ?
Like we do for `Axis.Tick.Marker` ?

and expose `formatFloatTick` as `DefaultTickFormatter` ?

DavidGamba: Do you mean:

1. Creating an interface:

````
type TickFormatter interface {
    Format func(v float64, prec int) string
}
````

And having `Axis.Format` be of type `TickFormatter` and have a `DefaultTickFormat` struct implementing the interface and assigning that in `makeAxis`?

or

2. Simply create a `DefaultTickFormat` function that is exposed and gets assigned to `Axis.Format` (of type `func(v float64, prec int) string`) in `makeAxis`?

Comment (sbinet) [reply axis.go]: (apologies for not being very clear)

the latter. this makes this new field (and its handling in the code) much more inline with the others.

**Comment by DavidGamba** [comment]
@sbinet PTAL

**Review by sbinet — DISMISSED** [review]
LGTM.

@kortschak and/or @eaburns should definitely comment, though :)

File: axis.go [hunk 6]
```diff
@@ -74,6 +76,10 @@ type Axis struct {
 		// returned by the Marker function that are not in
 		// range of the axis are not drawn.
 		Marker Ticker
+
+		// Format function used to format the Axis Ticks.
+		// When no format is provided a sane default is used.
```
kortschak: `// When format is nil DefaultTickFormat is used.`

axis.go
```diff
@@ -20,8 +20,10 @@ const displayPrecision = 4
 
 // Ticker creates Ticks in a specified range
 type Ticker interface {
-	// Ticks returns Ticks in a specified range
-	Ticks(min, max float64) []Tick
+	// Ticks returns Ticks in a specified range and formatted according to the
+	// given format function.
+	// When no format is provided (nil) a sane default is used.
```
kortschak: `// When format is nil DefaultTickFormat is used.`

Sanity is subjective. We can claim this in talks, blogs and wiki articles, but I don't think it's appropriate in documentation.

axis_test.go
```diff
@@ -7,42 +7,57 @@ package plot
...
 	"reflect"
+	"strconv"
 	"testing"
 )
 
 func TestAxisSmallTick(t *testing.T) {
 	d := DefaultTicks{}
 	for _, test := range []struct {
 		min, max float64
+		format   func(v float64, prec int) string
```
kortschak: There should be a test where `format` is `nil`.

DavidGamba: This is where my lack of knowledge of the codebase comes in. Initially I added this check:

````
       if format == nil {
               format = formatFloatTick
       }
````

In `DefaultTicks.Ticks` and `LogTicks.Ticks`. However, @sbinet told me that the check wasn't required if I applied the DefaultTickFormat in `makeAxis` ( https://github.com/gonum/plot/pull/362#discussion_r136914101 ).
Right now, if you call `DefaultTicks.Ticks` directly without a format you get a panic in the test. I am assuming this will never happen but I don't have enough insight into it.

@sbinet and @kortschak please let me know what course to take on this one.

Issue #364, opened 2017-06-08
vg/vgimg: consider adding Image() draw.Image to Canvas


sbinet:
it would be great to be able to retrieve the underlying `draw.Image` of a `vgimg.Canvas`.

ie:
```go
func (c *Canvas) Image() draw.Image {
        return c.img
}
```

`vgimg.Canvas` can be created with an already created `draw.Image` (using the `UseImage` functional option) but once created there is no efficient way to get back at it. (one very inefficient way is to `WriteTo` the canvas to an `io.Writer` and `png.Decode`-it back...)

I have been using this to create an animated GIF out of a slice of PNG plots.
Directly creating the animated GIF from the `draw.Image` images was faster than writing all the PNG files, decoding them back, encoding them to GIF and assembling them as an animated GIF. (d'oh!)
see:
 https://github.com/sbinet-lsst/ctrack-plot

kortschak:
That seems pretty reasonable to me.

PR #365, opened 2017-06-09
vg/vgimg: add Image to vgimg.Canvas

sbinet:
Fixes gonum/plot#364.

### DIFF
--- a/vg/vgimg/vgimg.go
+++ b/vg/vgimg/vgimg.go
@@ -164,6 +164,13 @@ func UseImageWithContext(img draw.Image, gc draw2d.GraphicContext) option {
 	}
 }
 
+// Image returns the image the canvas is drawing to.
+//
+// The dimensions of the returned image must not be modified.
+func (c *Canvas) Image() draw.Image {
+	return c.img
+}
+
 func (c *Canvas) Size() (w, h vg.Length) {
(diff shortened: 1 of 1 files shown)


vg/vgimg/vgimg.go
```diff
@@ -164,6 +164,14 @@ func UseImageWithContext(img draw.Image, gc draw2d.GraphicContext) option {
 	}
 }
 
+// Image returns the image the canvas is drawing to.
+//
+// Users are not supposed to directly modify the image once the canvas
```
kortschak:
```
// Users should not directly modify the returned image.
```

Are you saying that no modification is allowed? or just things like size. I think this needs to be fleshed out more.

sbinet: yes, the size (ie: via `draw.Image.Bounds()` and then `image.Rectange{ Min, Max Point }`).

sbinet: fixed.

kortschak: OK. So then probably we should say "The dimensions of the returned image must not be modified."

sbinet: done.

kortschak: Thanks.

PR #367, opened 2017-06-20
all: update imports to gonum.org

btracey:


btracey:
PTAL @sbinet @kortschak

PR #369, opened 2017-07-06
Correction for Luminance.At() when range min < 0

bnoon:
Need to calculate normalized value over whole range.


ctessum:
looks good to me.

kortschak:
@bnoon Can you send a PR to github.com/gonum/gonum adding yourself to A+C and reference this PR in the request.

bnoon:
Submitted.  Thanks!

PR #371, opened 2017-08-07
plotter: fix ColorBar range bug

ctessum:
The color bars previously panicked or rendered incorrectly when the range was not [0,1].


plotter/colorbar.go
```diff
@@ -58,35 +58,36 @@ func (l *ColorBar) Plot(c draw.Canvas, p *plot.Plot) {
 	colors := l.colors(c)
 	var img *image.NRGBA64
 	var xmin, xmax, ymin, ymax vg.Length
+	rnge := l.ColorMap.Max() - l.ColorMap.Min()
```
sbinet: what about:

```go
delta := (l.ColorMap.Max() - l.ColorMap.Min()) / float64(colors-1)
```

instead ?

(and modify accordingly below)

ctessum: Done.

PR #373, opened 2017-08-08
plotter: Fix floating point rounding error in ColorBar

ctessum:
Occasionally there would be a panic caused by a rounding error. This fixes it but I'm not sure whether this is the best fix, or of a good way to test it.

### DIFF
--- a/plotter/colorbar.go
+++ b/plotter/colorbar.go
@@ -6,6 +6,7 @@ package plotter
 
 import (
 	"image"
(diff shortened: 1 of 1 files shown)


kortschak:
I would like to understand this better before a fix is applied. Please instrument your local code to test index values prior to access and log the input and result.

kortschak:
I believe this was fixed by #411.

PR #385, opened 2017-09-02
palette: Create Reverse ColorMap.

ctessum:
This allows the direction of the colors to be reversed.


palette/reverse.go
```diff
@@ -0,0 +1,30 @@
...
+
+package palette
+
+import (
+	"image/color"
+)
+
+// Reverse is a ColorMap that reverses the direction of the ColorMap it
+// contains.
+type Reverse struct {
```
sbinet: Instead of exposing a type, perhaps do like the stdlib's 'sort.Reverse' that exposes a function and returns an unexposed type that implements the same interface?

sbinet:
ping?

ctessum:
PTAL

Issue #366, opened 2017-06-12
plot: Y-range value display bug


sbinet:
running the following program:

```go
package main

import (
	"flag"
	"log"

	"github.com/gonum/plot"
	"github.com/gonum/plot/plotter"
	"github.com/gonum/plot/vg"
)

func main() {
	v := flag.Float64("v", 1300, "y-value")
	flag.Parse()

	p, err := plot.New()
	if err != nil {
		log.Fatal(err)
	}

	p.Y.Label.Text = "RPMs"

	var data = make(plotter.XYs, 10)
	for i := range data {
		data[i].X = float64(i)
		data[i].Y = *v
	}

	lines, points, err := plotter.NewLinePoints(data)
	if err != nil {
		log.Fatal(err)
	}
	p.Add(points, lines)
	p.Add(plotter.NewGrid())

	err = p.Save(10*vg.Centimeter, 10*vg.Centimeter, "foo.png")
	if err != nil {
		log.Fatal(err)
	}
}
```

gives the following plot:
![bug-1300](https://user-images.githubusercontent.com/1070920/27033100-2116b634-4f78-11e7-8600-c3d033a5d99b.png)

ie: while the data line is at the correct location, the Y-axis labels are wrong (1300 is repeated twice, 1301 and 1299 aren't probably consistent).

this doesn't occur when I run with, say, `1401` or `130` as a value, but this bug also occurs for `1201`, `1400` and `1501`.

I suspect an issue in the rounding+formatting of the axis labels.

sbinet:
fixed by #388.

PR #386, opened 2017-09-10
This commit fixes issue #366

Scorpiokat:
displayPrecision used to be a constant which did not depend on a Tick value. Now it is being calculated based on the number of figures in the integer part + 1 of a Tick Value which is enough for large g-formatted numbers.

### DIFF
--- a/axis.go
+++ b/axis.go
@@ -9,6 +9,7 @@ import (
 	"math"
 	"strconv"
 	"time"
+	"strings"
 
 	"gonum.org/v1/gonum/floats"
 
@@ -16,8 +17,6 @@ import (
 	"gonum.org/v1/plot/vg/draw"
 )
 
-// displayPrecision is a sane level of float precision for a plot.
-const displayPrecision = 4
 
 // Ticker creates Ticks in a specified range
 type Ticker interface {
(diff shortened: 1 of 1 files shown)


sbinet:
@Scorpiokat it's getting there :) (and thanks for sticking with us and working on this issue!)

as you can see, there are still 2 tests failing.

by the way, you can run the tests with:
```
$> cd $GOPATH/src/gonum.org/v1/plot
$> go test ./...
```

the `logscale.png` one is changing `1e+04` into `10000`. I guess we could live with that.

the other one, `timeseries.png` is actually a regression, and one that is related to the issue we are trying to fix with this commit (see attached plot)
timeseries

see how we get 2 y-ticks at `1.4` ?

I am thinking out loud, here, but perhaps we need to change a bit how we figure out the needed number of float-digits for the precision.
instead of trying to figure it out from just one value, we'd need to figure it out from a range of values (or just the min and max, probably also taking care of the case where min < 0 and max > 0).
we'd compute the delta over the range and see that we'd need x digits to correctly represent that varying range.

sbinet:
also, as we are trying to fix `#366`, it would be good to add a test (under `plotter`) that reproduces that issue.

here is an example:

```go
// Copyright ©2017 The gonum Authors. All rights reserved.

package plotter_test

import (
        "testing"

        "gonum.org/v1/plot"
        "gonum.org/v1/plot/internal/cmpimg"
        "gonum.org/v1/plot/plotter"
)

func TestFloatPrecision(t *testing.T) {

        const fname = "precision.png"

        cmpimg.CheckPlot(func() {
                p, err := plot.New()
                if err != nil {
                        t.Fatal(err)
                }

                p.X.Label.Text = "x"
                p.Y.Label.Text = "y"

                const y = 1300

                var data = make(plotter.XYs, 10)
                for i := range data {
                        data[i].X = float64(i)
                        data[i].Y = y
                }

                lines, points, err := plotter.NewLinePoints(data)
                if err != nil {
                        t.Fatal(err)
                }
                p.Add(points, lines)
                p.Add(plotter.NewGrid())

                err = p.Save(200, 200, "testdata/"+fname)
                if err != nil {
                        t.Fatal(err)
                }
        }, t, fname)

}
```

sbinet:
actually: a simple(r) algorithm would be to take all the ticks values to be displayed, format them, and check that all strings are different.
if not, then add a precision digit until they are all different.

the algorithm complexity is `O(N^2)` but the number `N` should be quite small, right ?

Scorpiokat:
#388 is the newer version of this PR. Closing this one.

PR #388, opened 2017-09-18
Y-axis rounding and display bug fixes.

Scorpiokat:
Hi again:) 

This is my third attempt to fixing the issue  #366. I hope the most successful one. Thank you @sbinet for your advice and support.

Here are the changes made:
1.) I have changed the way of making Ticks. 

Now it is more "float64-based" rather than the "strings" one (before). 

Method (briefly): DisplayPrecision should not be a constant, as we have figured out earlier. So, to find out which one will be the perfect one, first of all, the array of y-values is created. Then the precision of one is taken, applied to all the values in the array and checked if all of them are different. If they are - the right level of precision is found and Ticks can be already made, if not - precision level is increased by one and the checking process starts again, until all the ticks in the array are different.

For implementing this method - two functions have been added -  RemoveDuplicates (removes duplicates from an array and returns one with just unique numbers) and FindRightLevelOfPrecision (find the right level of precision and returns an array with unique numbers truncated to that precision level) In the second function 2 methods were used (for small and very large numbers). floats.Round was not used here as unluckily it did not work for large numbers.

2.) logscale_golden.png has been replaced under plotter/testdata, as I do really think that the same level of precision should be applied to all the Ticks. Now it does.

3.) New test precision_test.go under plotter has been created to test the precision issue (like the one we are now trying to fix). precision_golden.png has been added to the testdata folder which contains all the plots to be compared to.

4.) Some more test cases were added to axis_test.go, so that the test covers almost 100% of lines in func (DefaultTicks)Ticks. Just one test for max<=min is left uncovered though.

5.) math.Nextafter have been removed twice from the code, as these cases can take place if and only if majorDelta = 0, which can be true only when min = max (and this case has been caught by 'panic (illegal range)'.


axis.go
```diff
@@ -355,12 +350,22 @@ type DefaultTicks struct{}
 
 var _ Ticker = DefaultTicks{}
 
+var labels []float64
+
+var new_labels []float64
```
sbinet: why do you make this a global variable ?

AFAICT, it's only accessed inside `DefaultTicks.Ticks(min,max)`.
just declare it there?

axis.go
```diff
@@ -355,12 +350,22 @@ type DefaultTicks struct{}
...
+
+var displayPrecision int
+
 // Ticks returns Ticks in a specified range
 func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
 	const SuggestedTicks = 3
-	if max < min {
+	labels = nil
+	ticks = nil
```
sbinet: `ticks` is already `nil` at this point.
just drop that line.

axis.go
```diff
@@ -355,12 +350,22 @@ type DefaultTicks struct{}
 
 var _ Ticker = DefaultTicks{}
 
+var labels []float64
```
sbinet: why do you make this a global variable ?

AFAICT, it's only accessed inside DefaultTicks.Ticks(min,max).
just declare it there?

axis.go
```diff
@@ -355,12 +350,22 @@ type DefaultTicks struct{}
...
+var displayPrecision int
+
 // Ticks returns Ticks in a specified range
 func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
 	const SuggestedTicks = 3
-	if max < min {
+	labels = nil
+	ticks = nil
+
+	if max <= min {
```
sbinet: I am not sure I understand why you change this test from `max < min` to `max <= min` ?
(it's perhaps correct, but I just don't understand why)

axis.go
```diff
@@ -407,14 +418,63 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
...
 	}
 	return
 }
 
+func removeDuplicates(elements []float64) []float64 {
+	// Uses a map to record duplicates as we find them.
+	encountered := map[float64]bool{}
+	result := []float64{}
+
+	for v := range elements {
```
sbinet: I'd use the key,value range instead:

```go
for i,v := range elements {
    if encountered[v] {
        continue
    }
    encountered[v] = true
    results = append(results, v)
}
return result
```

also, more Go-idiomatic names would be:
```go
seen := make(map[float64]bool)
out := make([]float64, 0, len(elements))
```

axis.go
```diff
@@ -407,14 +418,63 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
...
+	}
+	// Returns the new slice.
+	return result
+}
+
+func findRightLevelOfPrecision(elements []float64) []float64 {
+
+	var result []float64
+
+	for i := 1; i < 308; { //308 is the precision of the MaxFloat64 which we take as the maximum one here
```
sbinet: I'd add the `i++` part of the for-loop here and drop the:
```go
} else {
    i = i + 1
}
```
from the end.

axis.go
```diff
@@ -407,14 +418,63 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
...
+			} else {
+				new_element, _ = strconv.ParseFloat(strconv.FormatFloat(v, 'g', i, 64), 64)
+			}
+
+			result = append(result, new_element)
+		}
+		//checks whether the result array has duplicates. If not, the right level of precision is found, if it does, one more digit to the precision level is added and the check runs again.
+		if len(result) == len(removeDuplicates(result)) {
+			displayPrecision = i
+			break
```
sbinet: instead of `break`-ing, just `return result`.

axis.go
```diff
@@ -407,14 +418,63 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
...
+			result = append(result, elements[v])
+		} // Otherwise, no elements are added as they are duplicates.
+	}
+	// Returns the new slice.
+	return result
+}
+
+func findRightLevelOfPrecision(elements []float64) []float64 {
+
+	var result []float64
```
sbinet: instead of declaring `result` here and reinitializing it in the outer for-loop, just declare it inside that for-loop (and `return nil` at the end of the function.)

axis.go
```diff
@@ -407,14 +418,63 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
...
+	for i := 1; i < 308; { //308 is the precision of the MaxFloat64 which we take as the maximum one here
+
+		var new_element float64
+
+		var v_mult float64
+
+		result = nil
+
+		for _, v := range elements {
+			v_mult = v * math.Pow10(i)
```
sbinet: you should probably compute `math.Pow10(i)` only once outside of the inner for-loop.
and reuse it there.

also, `v_mult` doesn't look too Go-idiomatic.
`w` ? or `vpow`, or `vp` or `v10`.

axis.go
```diff
@@ -407,14 +418,63 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
...
+	return result
+}
+
+func findRightLevelOfPrecision(elements []float64) []float64 {
+
+	var result []float64
+
+	for i := 1; i < 308; { //308 is the precision of the MaxFloat64 which we take as the maximum one here
+
+		var new_element float64
```
sbinet: just declare `new_element` and `v_mult` inside the inner for-loop.

axis.go
```diff
@@ -407,14 +418,63 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
...
+
+		var v_mult float64
+
+		result = nil
+
+		for _, v := range elements {
+			v_mult = v * math.Pow10(i)
+			//As within 'if' statement v_mult is turned to integer, it cannot exceed the MaxInt64. If it does, another rounding technique is used.
+			if v_mult < (math.MaxInt64) {
```
sbinet: what happens if `v` is negative ?
(also, drop the parens around `math.MaxInt64`)

axis.go
```diff
@@ -407,14 +418,63 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
...
+				new_element = float64(int(v_mult)) / math.Pow10(i)
```
sbinet: you should probably cached the inverse of `math.Pow10(i)` outside of the inner for-loop and multiply by it there.

axis.go
```diff
@@ -407,14 +418,63 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
...
+			} else {
+				new_element, _ = strconv.ParseFloat(strconv.FormatFloat(v, 'g', i, 64), 64)
```
sbinet: you should probably still check if the `error` is not `nil` and `panic` otherwise.
just to be sure.

axis.go
```diff
@@ -571,13 +631,13 @@ func log(x float64) float64 {
 	return math.Log(x)
 }
 
-// formatFloatTick returns a g-formated string representation of v
+//formatFloatTick returns a g-formated string representation of v
```
sbinet: please keep the space between `//` and `formatFloatTick`.

axis.go
```diff
@@ -571,13 +631,13 @@ func log(x float64) float64 {
...
 // to the specified precision.
 func formatFloatTick(v float64, prec int) string {
-	return strconv.FormatFloat(floats.Round(v, prec), 'g', displayPrecision, 64)
+	return strconv.FormatFloat(v, 'g', displayPrecision, 64)
```
sbinet: I am a bit unconfortable with `formatFloatTick` depending on a global variable (`displayPrecision`) that can/could be modified from various places.
couldn't we either:
- pass it as an additional argument
- or make `formatFloatTick` a method of some new `int`-like type
?

(same comment for `precisionOf`)

axis.go
Scorpiokat: Changed that. Thanks!

axis.go
Scorpiokat: Changed that. Thanks!

axis.go
Scorpiokat: Changed that. Thanks!

axis.go
Scorpiokat: As you might have noticed, I have removed lines
if math.Nextafter(val, val+majorDelta) == val {
-			break
-		}
Same for the minorDelta. The only one case when it can have happened is when majorDelta = 0, right? And this is true when min = max. So, I moved that condition to min<=max which will cause panic, not break, Would you agree&

axis.go
Scorpiokat: Changed that. Thanks!

axis.go
Scorpiokat: Changed that. Thanks!

axis.go
Scorpiokat: I was actually thinking where would be the best place to put it. Made the wrong choice. Changed it. Thanks!

axis.go
Scorpiokat: Changed that. Thanks!

axis.go
Scorpiokat: Changed that. Thanks!

axis.go
Scorpiokat: Have added a check for the smaller v-s, as for larger ones it does not make any difference. does it?

axis.go
Scorpiokat: Changed it. Thanks!

axis.go
Scorpiokat: Thanks! Added that, but in this case result has to be returned from under 'if', and 'else'.

axis.go
Scorpiokat: Changed that. Thanks!

axis.go
Scorpiokat: Changed that. Sorry for removing the space. Was not done on purpose.

axis.go
Scorpiokat: This one is the most tricky one. I have changed displayPrecision to prec in the fuction returned. Making it int-type might not be a great idea, as in this case all our 'looking for the right precision' work will be useless, right?

Scorpiokat: As for precisionOf, here adding a new argument might work. Will try to fix that.

Scorpiokat:
Sebastian, first of all, thank you for the review of changes. I have made some changes and comments to them. Just one more issue is left.

axis.go
```diff
@@ -355,12 +350,21 @@ type DefaultTicks struct{}
 
 var _ Ticker = DefaultTicks{}
 
+var displayPrecision int
+
 // Ticks returns Ticks in a specified range
 func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
 	const SuggestedTicks = 3
```
kortschak: While you are here, please make this `suggestedTicks`.

axis.go
```diff
@@ -377,17 +381,23 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
...
-		if math.Nextafter(val, val+majorDelta) == val {
-			break
+			//Makes a list of values.
+			labels = append(labels, val)
 		}
 		val += majorDelta
 	}
 
+	//Makes a list of labels with a level of precision where all the labels are different.
+	new_labels = findRightLevelOfPrecision(labels)
```
kortschak: s/findRightLevelOfPrecision/adjustPrecision/g

axis.go
```diff
@@ -407,14 +417,74 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
...
+	// Uses a map to record duplicates as we find them.
+	seen := make(map[float64]bool)
+	out := make([]float64,0,len(elements))
+
+	for _,v := range elements {
+		if seen[v] {
+			//No elements are added as they are duplicates.
+			continue
+		}
+			// Records this element as a seen element.
```
kortschak: Please run gofmt.

axis.go
```diff
@@ -407,14 +417,74 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
...
 }
 
+func removeDuplicates(elements []float64) []float64 {
+	// Uses a map to record duplicates as we find them.
+	seen := make(map[float64]bool)
+	out := make([]float64,0,len(elements))
+
+	for _,v := range elements {
+		if seen[v] {
+			//No elements are added as they are duplicates.
```
kortschak: The comments in this function from here on are not necessary.

axis.go
```diff
@@ -407,14 +417,74 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
...
+			seen[v] = true
+			// Appends it to the result slice.
+			out = append(out, v)
+	}
+	// Returns the new slice.
+	return out
+}
+
+func findRightLevelOfPrecision(elements []float64) []float64 {
+
```
kortschak: Please don't lead with a blank line.

axis.go
```diff
@@ -407,14 +417,74 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
...
+	var result []float64
+
+	for i := 1; i < 308; i++ { //308 is the precision of the MaxFloat64 which we take as the maximum one here
+
```
kortschak: Please no blank line.

axis.go
```diff
@@ -407,14 +417,74 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
...
+	}
+	// Returns the new slice.
+	return out
+}
+
+func findRightLevelOfPrecision(elements []float64) []float64 {
+
+	var result []float64
+
+	for i := 1; i < 308; i++ { //308 is the precision of the MaxFloat64 which we take as the maximum one here
```
kortschak: Rather than using a comment, make a `const maxExp = 308` and use that.

axis.go
```diff
@@ -407,14 +417,74 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
...
+
+		var result []float64
+		var i10 float64
+
+		i10 = math.Pow10(i)
+
+		for _, v := range elements {
+
```
kortschak: Please no leading blank line.

axis.go
```diff
@@ -407,14 +417,74 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
...
+			var new_element float64
+			var vpow float64
+
+			vpow = v*i10
+
+			//As within 'if' statement vpow is turned to integer, it cannot exceed the MaxInt64. If it does, another rounding technique is used.
+				if vpow < math.MaxInt64 {
```
kortschak: gofmt

axis.go
```diff
@@ -407,14 +417,74 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
...
+					if v>=0 {
+						new_element = float64(int(vpow)) / i10
+					} else {
+						new_element = 0-float64(int(math.Abs(v)*i10)) / i10
```
kortschak: Is the 0 necessary?

axis.go
```diff
@@ -407,14 +417,74 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
...
+}
+
+func findRightLevelOfPrecision(elements []float64) []float64 {
+
+	var result []float64
+
+	for i := 1; i < 308; i++ { //308 is the precision of the MaxFloat64 which we take as the maximum one here
+
+		var result []float64
+		var i10 float64
```
kortschak: Delete this and replace the following line with the gopher operator.

```
i10 := math.Pow10(i)
```

and remove the blank line following it.

axis.go
```diff
@@ -407,14 +417,74 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
...
+
+		i10 = math.Pow10(i)
+
+		for _, v := range elements {
+
+			var new_element float64
+			var vpow float64
+
+			vpow = v*i10
```
kortschak: `vpow := v*i10`

axis.go
```diff
@@ -407,14 +417,74 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
...
+
+			//As within 'if' statement vpow is turned to integer, it cannot exceed the MaxInt64. If it does, another rounding technique is used.
```
kortschak: We have (not quite correct - but good enough for this) rounding code in floats. Can you just copy that here as an unexported function and use that please?

axis.go
```diff
@@ -407,14 +417,74 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
...
 		}
-		if math.Nextafter(val, val+minorDelta) == val {
-			break
-		}
 		val += minorDelta
 	}
 	return
 }
 
+func removeDuplicates(elements []float64) []float64 {
```
kortschak: Can you move this below where it is called please.

axis.go
```diff
@@ -407,14 +417,74 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
...
+	// Uses a map to record duplicates as we find them.
+	seen := make(map[float64]bool)
```
kortschak: Note that elements are in an order, so you can just iterate over elements and retain elements that are not the same as the previous one.

axis.go
```diff
@@ -355,12 +350,21 @@ type DefaultTicks struct{}
 
 var _ Ticker = DefaultTicks{}
 
+var displayPrecision int
```
kortschak: This cannot be used as a global. If I have more than one plot being generated at the same time by different goroutines, now we have a race.

axis.go
```diff
@@ -407,14 +417,74 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
...
+					new_element, err := strconv.ParseFloat(strconv.FormatFloat(v, 'g', i, 64), 64)
+					if err != nil {
+						panic(err)
+					}
+					result = append(result, new_element)
+				}
+		}
+		//checks whether the result array has duplicates. If not, the right level of precision is found, if it does, one more digit to the precision level is added and the check runs again.
+		if len(result) == len(removeDuplicates(result)) {
+			displayPrecision = i
```
kortschak: Potential race here.

axis.go
```diff
@@ -428,18 +498,17 @@ func (LogTicks) Ticks(min, max float64) []Tick {
 	if min <= 0 {
 		panic("Values must be greater than 0 for a log scale.")
 	}
-	prec := precisionOf(max)
 	for val < max*10 {
 		for i := 1; i < 10; i++ {
 			tick := Tick{Value: val * float64(i)}
 			if i == 1 {
-				tick.Label = formatFloatTick(val*float64(i), prec)
+				tick.Label = formatFloatTick(val*float64(i), displayPrecision)
```
kortschak: ... and here ...

axis.go
```diff
@@ -428,18 +498,17 @@ func (LogTicks) Ticks(min, max float64) []Tick {
...
 			}
 			ticks = append(ticks, tick)
 		}
 		val *= 10
 	}
-	tick := Tick{Value: val, Label: formatFloatTick(val, prec)}
+	tick := Tick{Value: val, Label: formatFloatTick(val, displayPrecision)}
```
kortschak: ... and here.

plotter/precision_test.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"testing"
+	"gonum.org/v1/plot/internal/cmpimg"
+	"log"
+	"gonum.org/v1/plot"
+	"gonum.org/v1/plot/plotter"
+)
+
+func TestFloatPrecision(t *testing.T) {
+
```
kortschak: Please don't lead with a blank line.

plotter/precision_test.go
```diff
@@ -0,0 +1,49 @@
...
+		p.Add(plotter.NewGrid())
+
+
+		err = p.Save(200, 200, fname)
+		if err != nil {
+			log.Fatal(err)
+		}
+	}, t, fname)
+
+}
```
kortschak: gofmt.

axis.go
Scorpiokat: Changed that. I hope now it will not cause race.

axis.go
Scorpiokat: Much better name, thanks!

axis.go
Scorpiokat: Done. Thanks!

axis.go
Scorpiokat: Apparently, yes. Works for v = 0 (I had to change the if condition to v>0, otherwise int(vpow) did not work).

Scorpiokat:
And I cannot see how I can replace the logscale_golden,png plot again, as the precision method was changed. Could you please help me with that?
logscale_golden

axis.go
Scorpiokat: Sorry, but it looks like new_element = floats.Round(v,i) does not work here. I am exploring why.

plotter/precision_test.go
Scorpiokat: Dan, thanks a lot for your help and comments!

kortschak:
Did you mean to close this? #386 is still open, but appears to be older.

Scorpiokat:
I am sorry, pressed it by mistake.

axis.go
Scorpiokat: I tried making that 
`for _, v := range elements {
		for _, w := range elements[1:] {
			if v == w {
				continue
			}
		}
		out = append(out, v)
	}
	return out
}`
but it does not work. I do realise that it takes much less memory than the method offered by me, but here the arrays are relatively small, should not be a problem, right? Anyway, I will think why the level of precision get distorted.

kortschak:
>And I cannot see how I can replace the logscale_golden,png plot again, as the precision method was changed. Could you please help me with that?

Pass `-regen` to `go test` and it will regenerate the Golden images. Then commit the changed files after checking them visually.

axis.go
Scorpiokat: And, just a thought, I have changed the way of identification precision of LogScale, but here we can make the level of precision constant, so that really large numbers do not look ugly and overload the plot with zeros. My opinion is that scientific notation on all the Y-axis will look more elegant. And as the ticks here cannot be equal to each other (due to the way we make them), even the precision level of 1 should work, would you agree?

axis.go
kortschak: You only need to check whether the previous matches the current:
```
out := make([]float64, 1, len(elements))
out[0] = elements[0]
for i, v := range elements[1:] {
		if v == elements[i] {
			continue
		}
		out = append(out, v)
}
return out
```

axis.go
kortschak: What happens?

axis.go
kortschak: That is extremely troubling. What is the effect?

Also `float64(int(f))` where `f` is `float64` is better spelled as `math.Trunc(f)`.

axis.go
kortschak: That's an aesthetic decision. I think that needs discussion. Please make an issue to deal with that after this is in.

axis.go
```diff
@@ -424,26 +429,88 @@ var _ Ticker = LogTicks{}
...
+		if len(result) == len(removeDuplicates(result)) {
+			//return the resulting array
+			return result
+		}
+	}
+	return nil
+}
+
+func removeDuplicates(elements []float64) []float64 {
+	// Uses a map to record duplicates as we find them.
```
kortschak: This will change with the code that I posted replacing this with `hasDuplicates`, but for future reference none of the comments in this function are necessary since the code speaks clearly enough for itself.

axis.go
```diff
@@ -424,26 +429,88 @@ var _ Ticker = LogTicks{}
...
+
+	for j := 0; j < len(labels); j++ {
+		//Adds big ticks to the list of small ones
+		ticks = append(ticks, Tick{Value: labels[j], Label: formatFloatTick(new_labels[j], -1)})
+	}
 	return ticks
 }
 
+func adjustPrecision(elements []float64) []float64 {
+	const maxExp = 308 //precision of the MaxFloat64 which we take as the maximum one here
```
kortschak: At the most, this comment should be `// maxExp is the maximum float64 exponent.`

axis.go
```diff
@@ -424,26 +429,88 @@ var _ Ticker = LogTicks{}
...
+	for i := 1; i < maxExp; i++ {
+		var result []float64
+		i10 := math.Pow10(i)
+
+		for _, v := range elements {
+			var new_element float64
+			vpow := v * i10
+			//As within 'if' statement vpow is turned to integer, it cannot exceed the MaxInt64. If it does, another rounding technique is used.
```
kortschak: This comment is not necesary, and it turns out also incorrect for some platforms; in the case that this code is written compiled on a 32 but arch, the guard you have will fail to prevent a truncation since `sizeof(int)` < `sizeof(int64`. This is (probably) obviated by changing to use `math.Trunc`, but I'd like to understand/know what the problem is with using the code from `floats.Round`.

axis.go
```diff
@@ -424,26 +429,88 @@ var _ Ticker = LogTicks{}
...
+				result = append(result, new_element)
+			} else {
+				new_element, err := strconv.ParseFloat(formatFloatTick(v, i), 64)
+				if err != nil {
+					panic(err)
+				}
+				result = append(result, new_element)
+			}
+		}
+		//checks whether the result array has duplicates. If not, the right level of precision is found, if it does, one more digit to the precision level is added and the check runs again.
```
kortschak: This comment is not necessary since the code speaks for itself. If you are troubled, a nicer way to make the explanation clearer would be to write

```
    if !hasDuplicates(result) {
        return result
    }
```

then (since `removeDuplicates`'s return value is not used for anything else beyond testing for whether an element has been removed) `removeDuplicates` can be rewritten as `hasDuplicates`:

```
// hasDuplicates returns whether the sorted slice f has a duplicate element.
// Standard IEEE 754 NaN equality is used since tick values are not expected
// to contain NaN values.
func hasDuplicates(f []float64) bool {
    for i, v := range f[1:] {
        if v == f[i] {
            return true
        }
    }
    return false
}
```

This improves the code, the performance, and the clarity.

axis.go
Scorpiokat: So, if I change the function to this:
`func adjustPrecision(elements []float64) []float64 {`
`        const maxExp = 308 //maxExp is the maximum float64 exponent`
`        for i := 1; i < maxExp; i++ {`
`        var result []float64`
`            for _, v := range elements {`
`                       newElement := floats.Round(v,i)`
`                       result = append(result, newElement)`
`                        }`
`                if !hasDuplicates(result) {`
`                 return result`
`                 }`
`         }`
`      return nil`
`}`
the code fails the TestAxisSmallTick (axis_test.go).

Here are the test results:
	axis_test.go:63: tick labels mismatch:
		got: ["-2e-15" "-1e-15" "0"]
		want:["-1.6e-15" "-8e-16" "0"]
	axis_test.go:63: tick labels mismatch:
		got: ["4.799999999999997e+307" "5.199999999999997e+307" "5.599999999999996e+307"]
		want:["4.8e+307" "5.2e+307" "5.6e+307"]

TestLegendAlignment fails too (cannot catch the difference visually).

My new test TestFloatPrecision fails too, but here it is fine - ticks are just a bit different (1299-1299.6-1300.2-1300.8).

kortschak: I think those failures are OK. The first one is an aesthetic choice and the second one is unreasonable to expect correctness for.

Note that while you can test with `floats.Round` you cannot import `floats` and will have to copy-paste the implementation into an unexported function here - we don't import any gonum code into plot to avoid having people who just want plotting to have to bring all the rest of the suite.

Also, you may want to look at how to use \`\`\` and \` marks for \<pre\> blocks. See [here](https://guides.github.com/features/mastering-markdown/).

Scorpiokat: Actually `floats` have been imported to the code when I started to make changes. Never mind. I have replaced the function made by myself with `Round`. Under the same name. 

Thanks a lot for your help!

axis.go
```diff
@@ -424,26 +427,96 @@ var _ Ticker = LogTicks{}
...
+	return nil
+}
+
+// Round returns the half away from zero rounded value of x with prec precision.
+//
+// Special cases are:
+// 	Round(±0) = +0
+// 	Round(±Inf) = ±Inf
+// 	Round(NaN) = NaN
+func Round(x float64, prec int) float64 {
```
kortschak: Make this unexported please.

axis.go
```diff
@@ -424,26 +427,96 @@ var _ Ticker = LogTicks{}
...
+
+	for j := 0; j < len(labels); j++ {
+		//Adds big ticks to the list of small ones
+		ticks = append(ticks, Tick{Value: labels[j], Label: formatFloatTick(newLabels[j], -1)})
+	}
 	return ticks
 }
 
+func adjustPrecision(elements []float64) []float64 {
+	const maxExp = 308 //maxExp is the maximum float64 exponent
```
kortschak: Space after `//` and full stop at end of sentence.

axis.go
```diff
@@ -424,26 +427,96 @@ var _ Ticker = LogTicks{}
...
 		}
 		val *= 10
 	}
-	tick := Tick{Value: val, Label: formatFloatTick(val, prec)}
-	ticks = append(ticks, tick)
+
+	newLabels = adjustPrecision(labels)
+
+	for j := 0; j < len(labels); j++ {
+		//Adds big ticks to the list of small ones
```
kortschak: Space after `//` and full stop at end of sentence and move above the for loop.

axis.go
```diff
@@ -424,26 +427,96 @@ var _ Ticker = LogTicks{}
...
-	prec := precisionOf(max)
+
 	for val < max*10 {
 		for i := 1; i < 10; i++ {
 			tick := Tick{Value: val * float64(i)}
 			if i == 1 {
-				tick.Label = formatFloatTick(val*float64(i), prec)
+				labels = append(labels, val*float64(i))
 			}
+			//Makes a list of small ticks
```
kortschak: Space after `//` and full stop at end of sentence.

axis.go
```diff
@@ -424,26 +427,96 @@ var _ Ticker = LogTicks{}
 // Ticks returns Ticks in a specified range
 func (LogTicks) Ticks(min, max float64) []Tick {
 	var ticks []Tick
+	var labels, newLabels []float64
+	labels = nil
```
kortschak: Delete this line; `labels` is already `nil`.

axis.go
```diff
@@ -377,17 +378,22 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
...
+			labels = append(labels, val)
 		}
 		val += majorDelta
 	}
 
+	//Makes a list of labels with a level of precision where all the labels are different.
+	newLabels = adjustPrecision(labels)
+
+	for j := 0; j < len(labels); j++ {
+		//Makes a list of big ticks.
```
kortschak: Space after `//` and move above the for loop.

axis.go
```diff
@@ -377,17 +378,22 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
...
+			//Makes a list of values.
+			labels = append(labels, val)
 		}
 		val += majorDelta
 	}
 
+	//Makes a list of labels with a level of precision where all the labels are different.
+	newLabels = adjustPrecision(labels)
+
+	for j := 0; j < len(labels); j++ {
```
kortschak: `for j, v := range labels {`

axis.go
```diff
@@ -377,17 +378,22 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
...
+		//Makes a list of big ticks.
+		ticks = append(ticks, Tick{Value: labels[j], Label: formatFloatTick(newLabels[j], -1)})
```
kortschak: `ticks = append(ticks, Tick{Value: v, Label: formatFloatTick(newLabels[j], -1)})`

Also, I suspect that if you get `adjustPrecision` to return the precision that was decided on as a second return parameter, if you pass that to `formatFloatTick` you would get rid of the problem with the very high exponent test failure (also below).

axis.go
```diff
@@ -377,17 +378,22 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
...
-		}
-		if math.Nextafter(val, val+majorDelta) == val {
-			break
+			//Makes a list of values.
+			labels = append(labels, val)
 		}
 		val += majorDelta
 	}
 
+	//Makes a list of labels with a level of precision where all the labels are different.
```
kortschak: `// Make a slice of distinct tick labels.`

axis.go
```diff
@@ -377,17 +378,22 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
 	}
 	majorDelta := float64(majorMult) * tens
 	val := math.Floor(min/majorDelta) * majorDelta
-	prec := precisionOf(majorDelta)
 	for val <= max {
 		if val >= min && val <= max {
```
kortschak: The second expression here is always true.

axis.go
```diff
@@ -424,26 +427,96 @@ var _ Ticker = LogTicks{}
...
 		val *= 10
 	}
-	tick := Tick{Value: val, Label: formatFloatTick(val, prec)}
-	ticks = append(ticks, tick)
+
+	newLabels = adjustPrecision(labels)
+
+	for j := 0; j < len(labels); j++ {
+		//Adds big ticks to the list of small ones
+		ticks = append(ticks, Tick{Value: labels[j], Label: formatFloatTick(newLabels[j], -1)})
```
kortschak: `ticks = append(ticks, Tick{Value: v, Label: formatFloatTick(newLabels[j], -1)})`

axis.go
```diff
@@ -424,26 +427,96 @@ var _ Ticker = LogTicks{}
...
 			ticks = append(ticks, tick)
 		}
 		val *= 10
 	}
-	tick := Tick{Value: val, Label: formatFloatTick(val, prec)}
-	ticks = append(ticks, tick)
+
+	newLabels = adjustPrecision(labels)
+
+	for j := 0; j < len(labels); j++ {
```
kortschak: `for j, v := range labels {`

axis.go
```diff
@@ -424,26 +427,96 @@ var _ Ticker = LogTicks{}
...
+		//Adds big ticks to the list of small ones
+		ticks = append(ticks, Tick{Value: labels[j], Label: formatFloatTick(newLabels[j], -1)})
+	}
 	return ticks
 }
 
+func adjustPrecision(elements []float64) []float64 {
+	const maxExp = 308 //maxExp is the maximum float64 exponent
+	for i := 1; i < maxExp; i++ {
```
kortschak: I don't think this is the correct range. `Round` can take negative values too. These become valuable for high exponent values.

axis.go
Scorpiokat: Removed second `val <= max`, the rest of the condition is valuable.

axis.go
Scorpiokat: It did not work, as formatFloatTick's precision is different, as it works with strings.

axis.go
Scorpiokat: THere is no problem with `round` being negative here, but precision(i) cannot be negative, can it? And it should be limited with something, where our maxExp is coming.

axis.go
kortschak: Yes, thanks, that's true. This highlights something that was at the back of my mind. When you and Seb were discussing your approach previously I had thought that you were going to do rendering and assess whether the rendered labels differed. I had forgotten this. It looks like there is an impedance mismatch in the approach that is being used that that approach would not have. We can forge ahead with this approach, since the basic structure is the same, but I think this particular problem goes away with that approach, the n is smaller (though the constants are larger).

axis.go
kortschak: You are right here. See comment above.

kortschak:
Fixes #390.

kortschak:
Sorry, I've just run the tests locally (I don't understand why they are not running here) and the small ticks tests fails. What I intended above was that you comment out those failing tests. You can add `// TODO(kortschak): Fix precision to make these tests pass.` when you do that.

Scorpiokat:
@kortschak Dan, I have mentioned this fail above when replaced my way of rounding numbers with the 'round' function. You said that was okay, so I left is as it was. Should I have changed the test itself?

Here are the results of the test failing:
axis_test.go:63: tick labels mismatch:
got: ["-2e-15" "-1e-15" "0"]
want:["-1.6e-15" "-8e-16" "0"]
axis_test.go:63: tick labels mismatch:
got: ["4.799999999999997e+307" "5.199999999999997e+307" "5.599999999999996e+307"]
want:["4.8e+307" "5.2e+307" "5.6e+307"]

kortschak:
Yeah, sorry. What I meant was that it was OK for the in principle failure to exist, but the tests should be made to pass by excluding those tests, but leaving a comment as above.

Scorpiokat:
Sorry for not understanding. Have commented them out.

kortschak:
Sorry for not being more clear.

kortschak:
That can come when a rendered-value-based choice is used to replace the float64-value-based approach that is here.

Scorpiokat:
@kortschak Sorry for asking, but could you please explain what do you mean by rendered-value-based choice?

kortschak:
The goal of the code here is to generate a series of tick label strings that are unique over the desired data range. The approach used at the moment is to generate a set of minimally procise `float64` values that are unique over that range and then render them to a set of strings. The problem with that (as demonstrated by the failing tests) is that sometime the rendering of a floating point value may be counterintuitive because base 10 non-integer values can not always be represented (e.g. 0.1), so when we use the `%g` verb, we get things like ["4.8e+307", "5.200000000000001e+307", "5.600000000000001e+307"] instead of ["4.8e+307", "5.2e+307", "5.6e+307"].

If we did what we are actually wanting, to get a set of unique strings, rather than going through what the `float64` representation is, i.e. we use the unrounded `float64` value and iterate over `%.*g` increasing the prec value, then we have 0 to < ~17 rounds of testing (instead of 0 to < 307) and we find the first prec that gives us a unique set of strings. We also get rid of the use of round, which we know to be frought. By using `strconv.AppendFloat` rather than `fmt.Sprintf`, we can do that with only a single allocation per loop (potentially on the stack, since we can work into an `[n]byte`).

PR #392, opened 2017-10-02
plot: fix axis test build

sbinet:
Commit e7f526d disabled a few tests.
But they were the only ones that needed "math".

Remove (temporarily) the import of "math".


sbinet:
no worries :)

Issue #195, opened 2015-04-16
plotter: Add capability for colored scatter plots


btracey:
Feature request for having a plotter where data points are {x,y,z} values and the z values are plotted according to color. Scatter plots are useful when the data is not evenly spread throughout the space (so generating a grid for a heat-map is difficult), or when there is noise in the data, so there isn't really a 'surface' to plot. 

All of the parts for colored-scatter exist. We can already plot 3-D scattered data where the glyph varies with the third dimension (plotter.Bubbles), and we can plot 3-D data where the color varies with the third dimension (HeatMap). Thus, in my opinion, the benefit to providing colored scatter greatly outweighs the small marginal maintenance cost of the extra plotter.

drgrib:
Sad to see this was posted 2 years ago. But, I agree as I posted in #345. The `HeatMap` requires us to interpolate values and associate them with a grid. A palette colored scatter plot allows us to freely assign x / y values in continuous space colored by a z value.

kortschak:
This is a straightforward plotter to write; take the codewalk shown for [`Bubbles`](https://github.com/gonum/plot/wiki/Creating-Custom-Plotters:-A-tutorial-on-creating-custom-Plotters) and instead of setting the radius, set the colour based on a colour interpolation as done in `HeatMap`.

PR #394, opened 2017-10-06
Add capability for colored scatter plots #195

Scorpiokat:
As recommended by @kortschak, I took the codewalk shown for Bubbles and instead of setting the radius, set the colour based on a colour interpolation as done in HeatMap.


ctessum:
A couple of initial comments:

* This pull request contains a lot of changes that are not related to colored scatter plots. We will want to review a pull request that only includes colored scatter plot-related changes.
* An alternative strategy for implementing this would be to do something similar to what is in the [sankey diagram plotter](https://github.com/gonum/plot/blob/master/plotter/sankey.go#L63), where there is an optional field that can be used to specify a function for specifying styles.

This could be accomplished by changing the ```Scatter``` type to look like this:

``` go 
// Scatter implements the Plotter interface, drawing
// a glyph for each of a set of points.
type Scatter struct {
	// XYs is a copy of the points for this scatter.
	XYs

       // GlyphStyleFunc, if not nil, specifies GlyphStyles
      // for individual points.
       GlyphStyleFunc  func(int) draw.GlyphStyle

	// GlyphStyle is the style of the glyphs drawn
	// at each point.
	draw.GlyphStyle
}
```

Then changing the plotting logic to use GlyphStyleFunc---if it is specified---instead of the GlyphStyle field, to specify the style of individual points. This would allow both colored scatter plots and bubble plots to be implemented using the existing ```Scatter``` type, and if we wanted we could delete the ```Bubbles``` type as it would be redundant and just add examples to show how to use ```Scatter``` to create colored scatter plots and bubble plots. For example, creating a colored scatter plot could roughly look like this:

``` go
	s, err := NewScatter(scatterData)
	if err != nil {
		log.Panic(err)
	}
        cm := moreland.BlackBody() // Initialize a color map.
        cm.Max = max
        cm.Min = min
        s.GlyphStyleFunc = func(i int) draw.GlyphStyle {
              c, err := cm.At(scatterData.Z(i))
             if err!=nil {log.Panic(err)}      
             return draw.GlyphStyle{Color:c, Radius: vg.Points(3)}
       }
       etc...
```

kortschak:
@Scorpiokat can you please rebase your changes here onto gonum/plot master (preferably in a non-master branch in your repo) so that there will not be commits from the previous PR.

[This](https://gist.github.com/Chaser324/ce0505fbed06b947d962) is a pretty good explanation of how to go about things. In short, please work from our master (keep your repo up to date) and do your work in a branch that you delete after the PR has been merged.

PR #395, opened 2017-10-06
axis: Added alternate Ticker "PreciseTicks"

### LINKED ISSUE #393: axis: Incorrect positioning of tick marks new scheme
Starting with commit e7f526d68a91a8b21c225cdbbbd8662d53c541f4, there is a problem with positioning and/or labeling of tick marks.  The new scheme rounds tick labels, but the tick marks do not move to reflect this.  The code _could_ be changed to move the tick marks to reflect the labels, but this would ruin the regular tick mark positioning.  I believe it fundamentally does not make sense to round the tick marks and/or labels.

The problem is exemplified by this plot generated before and after the commit in question (using the [go-hep.org hplot package](https://github.com/go-hep/hep/tree/master/hplot) which in-turn uses gonum plot), where 
the peak on the y axis is ideally 3.096:
## Before
before
## After
after

One can see that in the After case, the vertical tick labels misrepresent the position of the peak.

decibelcooper:
This PR resolves #393 

This alternate Ticker is heavily based on the default with two changes:
1. There is a different condition for setting major tick interval to a
lower order of magnitude
2. There is a different mechanism for limiting label precision.  Instead
of creating a new list that is rounded until all labels at just unique,
precision is inferred by the orders of magnitude for the largest label
and the major tick interval.  This precision is passed to the float
format function.

### DIFF
--- a/axis.go
+++ b/axis.go
@@ -356,8 +356,6 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
 
 	var labels, newLabels []float64
 
-	labels = nil
-
 	if max <= min {
 		panic("illegal range")
 	}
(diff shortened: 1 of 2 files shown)


axis.go
```diff
@@ -415,6 +415,76 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
...
+
+var _ Ticker = AltTicks{}
+
+// Ticks returns Ticks in a specified range
+func (AltTicks) Ticks(min, max float64) (ticks []Tick) {
+	const suggestedTicks = 3
+
+	var labels []float64
+
+	labels = nil
```
sbinet: that isn't needed. `var labels []float64` already initializes `labels` to `nil`.

axis.go
```diff
@@ -415,6 +413,76 @@ func (DefaultTicks) Ticks(min, max float64) (ticks []Tick) {
...
+			if t.Value == val {
+				found = true
+			}
+		}
+		if val >= min && val <= max && !found {
+			ticks = append(ticks, Tick{Value: val})
+		}
+		val += minorDelta
+	}
+	return
```
kortschak: I know that `DefaultTicks` does this (it shouldn't), but can you remove the named return and make this an explicit return value.

decibelcooper: Sure

decibelcooper:
# Simple reproducer
I have stripped down my code and put in random points in place of real data.  The follow code creates the plots below with the PreciseTicks and DefaultTicks, respectively.
```go
package main

import (
    "log"

    "go-hep.org/x/hep/hbook"
    "go-hep.org/x/hep/hplot"
    "gonum.org/v1/gonum/stat/distuv"
    "gonum.org/v1/plot"
    "gonum.org/v1/plot/vg"
)

const (
    m_jpsi  = 3.096916
    nEvents = 10000
)

func main() {
    hist2D := hbook.NewH2D(12, -6, 6, 100, m_jpsi-0.125, m_jpsi+0.125)

    massDist := distuv.Normal{Mu: m_jpsi, Sigma: 0.01}
    etaDist := distuv.Normal{Mu: -2, Sigma: 1}
    for i := 0; i < nEvents; i++ {
        hist2D.Fill(etaDist.Rand(), massDist.Rand(), 1)
    }

    p := hplot.New()
    p.X.Label.Text = "Eta"
    p.X.Tick.Marker = plot.DefaultTicks{}
    p.Y.Label.Text = "J/Psi Reconst. Mass {GeV}"
    p.Y.Tick.Marker = plot.DefaultTicks{}

    p.Add(hplot.NewH2D(hist2D, nil))
    err := p.Save(6*vg.Inch, 6*vg.Inch, "jpsiReconFromTracks2D.png")
    if err != nil {
        log.Print(err)
    }
}
```
## PreciseTicks
jpsireconfromtracks2d_precise
## DefaultTicks
jpsireconfromtracks2d_default

decibelcooper:
Ping

kortschak:
Thanks. Sorry for the delay.

PR #396, opened 2017-10-07
Adds capability for colored scatter plots (issue #195)

Scorpiokat:
I have closed the previous pull request, and created this one according to your comments @kortschak and @ctessum. Thanks!

@ctessum I decided to send this PR anyway (and it's up to you, obviously, to accept it or not right now), so that at least the plot has this capability and it will fix the #195 issue. Meanwhile, I will start working on the alternative solution which you have proposed. It should definitely look much more elegant and be an improvement of the existing way of making scatter, scatter colour and bubble plots.


Scorpiokat:
@ctessum Chris, I am a bit confused. If we decide not to merge this PR, and follow the directions from your comments [here](https://github.com/gonum/plot/pull/394), may I ask 2 things that I do not clearly understand:
1) It looks like we are trying to avoid using Z as a third parameter,  but then here `c, err := cm.At(scatterData.Z(i))` it suddenly appears. If we add `Z` to `scatter.go` this will cause a need to add it in a lot of other places like `line.go` too. If that is fine, I will do so.
2) What is the role of `i` here `s.GlyphStyleFunc = func(i int) draw.GlyphStyle` and where it comes from? We will use it to plot scatter color or bubble (or any other) plots?

ctessum:
Hi @Scorpiokat,

In Go, [functions can be variables](https://golang.org/doc/codewalk/functions/). 

How the concept I proposed would work is that we would add a field to the `Scatter` struct that is a function (`GlyphStyleFunc`). This is different than adding a method to `Scatter`, because method functions are defined at compile time-but field functions can be specified at run-time by the user.

Because it is a field and not a method, by default `GlyphStyleFunc` is nil. However, a user could at their discretion set `GlyphStyleFunc` to a function of their choosing, which would override the default style (that does not vary among scatter points) with the style specified by `GlyphStyleFunc` (which can vary among scatter points). 

There is already something similar implemented in the `Sankey` plotter. In this [example](https://godoc.org/gonum.org/v1/plot/plotter#ex-Sankey--Grouped) you can see how we can specify custom style functions (`FlowStyle` and `StockStyle`). In the `Sankey` plotter logic, you can see that [here](https://github.com/gonum/plot/blob/master/plotter/sankey.go#L194) we specify default `StockStyle` and `FlowStyle` fields (which can be changed by the user before plotting), and then [here](https://github.com/gonum/plot/blob/master/plotter/sankey.go#L258) we call the function variable to determine the style of each item.

So, to answer your specific questions:
1. Z would not be used in the code of the plotter itself, but could necessary in some user-specified `GlyphStyleFunc`s. In those cases, it would be up to the user to provide a values for Z or whatever other variable name they choose to call it.
2. `i` refers to the point index number, so for the first point `i=0`, for the second point `i=1`, etc. It is necessary to have `i` as a function argument so that users can create functions that give different styles for different points (e.g., a user could create a function that colors the first point red, the second point blue, etc.)

Let me know whether that makes sense.

kortschak:
You may also like to see how [bíogo rings](https://godoc.org/github.com/biogo/graphics/rings) does this kind of thing where data series are reflected on to see if they want to do styling. Note that, from memory, @eaburns  has indicated that he does not really want to see things like this in the plotters here.

Scorpiokat:
@ctessum thanks you for your comments. 

I got the idea, but still not completely clear on the Z (or whichever variable that is), and how all this `c, err := cm.At(scatterData.Z(i))` works, sorry. `scatterData.Z(i)` should be a function that depends on `i` and results in `float64`? That is a bit confusing.

And will `cm.At` give the same nice 'colour intensity growing' effect like palette.Heat? Or it will depend on the `float64` in the brackets?

ctessum:
What I didn't include in the original explanation was that I was asssuming that `scatterData` would be some type that had a method `Z(i int) float64` that returned the `z` value at index `i`. It wouldn't need to be like that, for example you could also have var `z []float64`, and then it would be `c, err := cm.At(z[i])`. Regardless, the plotter code itself wouldn't need to know anything about z values, it would just be in the example showing how to make colored scatter plots.

ctessum:
An independent question is whether there is consensus regarding whether this is a good idea. This may not be the correct location for that discussion, however.

Scorpiokat:
@ctessum I have sent a new commit with the changes you have proposed. As this is a 2-in-1 PR, it became a bit messy. I should probably delete `scatterColor.go` and `scatterColor_test.go`. And add a legend to the `scatter_test.go`, should I?

Scorpiokat:
ping

plotter/scatter_test.go
```diff
@@ -10,10 +10,14 @@ import (
 	"math/rand"
 	"testing"
 
+	"fmt"
```
sbinet: please put `"fmt"` and `"os"` as part of the stdlib import block

plotter/scatter_test.go
```diff
@@ -73,15 +79,66 @@ func ExampleScatter() {
...
+	const legendWidth = 3 * vg.Centimeter
+	// Slide the legend over so it doesn't overlap the ScatterPlot.
+	p.Legend.XOffs = legendWidth
 
-	err = p.Save(200, 200, "testdata/scatter.png")
+	img := vgimg.New(350, 290)
+	dc := draw.New(img)
+	dc = draw.Crop(dc, 0, -legendWidth, 0, 0) // Make space for the legend.
+	p.Draw(dc)
+	w, err := os.Create("testdata/scatter.png")
```
sbinet: we are missing a `defer w.Close()` here as well as an explicit `err = w.Close()` (with the according error check) after the call to `png.WriteTo(w)`.

plotter/scattercolor_test.go
```diff
@@ -0,0 +1,101 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package plotter
+
+import (
+	"gonum.org/v1/plot/palette"
```
sbinet: this `gonum.org/v1/plot/palette` import should be part of the `gonum.org` import block below.

plotter/scattercolor_test.go
```diff
@@ -0,0 +1,101 @@
...
+	"log"
+	"math/rand"
+	"testing"
+
+	"fmt"
```
sbinet: and `fmt` and `os` should be part of the stdlib import block above.

plotter/scattercolor_test.go
```diff
@@ -0,0 +1,101 @@
...
+	// This is the width of the legend, experimentally determined.
+	const legendWidth = 1.25 * vg.Centimeter
+	// Slide the legend over so it doesn't overlap the ScatterColorPlot.
+	p.Legend.XOffs = legendWidth
+
+	img := vgimg.New(300, 200)
+	dc := draw.New(img)
+	dc = draw.Crop(dc, 0, -legendWidth, 0, 0) // Make space for the legend.
+	p.Draw(dc)
+	w, err := os.Create("testdata/scattercolor.png")
```
sbinet: same comments than above about `defer w.Close()` and the explicit `w.Close()`.

plotter/scattercolor_test.go
```diff
@@ -0,0 +1,101 @@
...
+	"fmt"
+	"gonum.org/v1/plot"
+	"gonum.org/v1/plot/internal/cmpimg"
+	"gonum.org/v1/plot/vg"
+	"gonum.org/v1/plot/vg/draw"
+	"gonum.org/v1/plot/vg/vgimg"
+	"os"
+)
+
+// ExampleScatter draws some scatter coloured points.
```
sbinet: `s/ExampleScatter/ExampleScatter_color/`

plotter/scattercolor_test.go
```diff
@@ -0,0 +1,101 @@
...
+func ExampleScatterColor() {
```
sbinet: `s/ExampleScatter/ExampleScatter_color/`

plotter/scattercolor_test.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	p.Title.Text = "Scatter Colour Plot"
+	p.X.Label.Text = "X"
+	p.Y.Label.Text = "Y"
+	p.Add(NewGrid())
+
+	pal := palette.Heat(12, 1)
+
+	sc, err := NewScatterColor(scatterColorData, pal)
```
sbinet: use `NewScatter` instead.

plotter/scattercolor.go
```diff
@@ -0,0 +1,117 @@
...
+	"gonum.org/v1/plot/vg/draw"
+	"image/color"
+)
+
+// ScatterColor implements the Plotter interface, drawing
+// a scatter plot of x, y, z triples where the z value
+// determines the colour of a scatter plot point.
+// Scatter implements the Plotter interface, drawing
+// a glyph for each of a set of points.
+type ScatterColor struct {
```
sbinet: remove this type. it's not needed anymore.

Scorpiokat:
Thank you @sbinet . I decided to delete scattercolor.go and scattercolor_test.go completely, as they are excessive to this implementation. 

I am not very happy about the look of the legend on the right though, but have no idea on a better way to locate it.

sbinet:
I personnally prefered keeping the original scatter example untouched and add a new one exercizing the new scatter-color plot.

(but I'll let others chime in)

plotter/scatter.go
```diff
@@ -24,22 +28,32 @@ type Scatter struct {
...
 	}
-	return &Scatter{
-		XYs:        data,
-		GlyphStyle: DefaultGlyphStyle,
-	}, err
+	s.XYs = data
+
+	s.GlyphStyle = DefaultGlyphStyle
+
+	s.GlyphStyleFunc = func(int) draw.GlyphStyle {
```
sbinet: we could keep it `nil`, couldn't we? (as we handle the `nil` case in the `Scatter.Plot` method)

actually, I think we could leave the whole `NewScatter` function untouched.

plotter/scatter.go
```diff
@@ -24,22 +28,32 @@ type Scatter struct {
...
 }
 
 // Plot draws the Scatter, implementing the plot.Plotter
 // interface.
 func (pts *Scatter) Plot(c draw.Canvas, plt *plot.Plot) {
 	trX, trY := plt.Transforms(&c)
-	for _, p := range pts.XYs {
-		c.DrawGlyph(pts.GlyphStyle, vg.Point{X: trX(p.X), Y: trY(p.Y)})
+	for i, p := range pts.XYs {
+		if pts.GlyphStyleFunc != nil {
```
sbinet: instead of testing over and over whether `GlyphStyleFunc` is nil or not, couldn't we create a little closure (before the for-loop) that will return the correct `GlyphStyle` value?

something like:
```go
glyph := func(i int) draw.GlyphStyle { return pts.GlyphStyle }
if pts.GlyphStyleFunc != nil {
    glyph = pts.GlyphStyleFunc
}
for i, p := range pts.XYs {
    c.DrawGlyph(glyph(i), vg.Point{X: trX(p.X), Y: trY(p.Y)})
}
```

Scorpiokat: I do absolutely agree with you, have made the changes.

As for a separate color scatter plot, I am happy to bring it back, as it looks much better on its own, than together with 3 others.

Scorpiokat:
As for a separate color scatter plot, I am happy to bring it back, as it looks much better on its own, than together with 3 others.

sbinet:
yes, I think having a separate example (and thus a separate test) would be better.

Scorpiokat:
Looks better now?

sbinet:
yes.

plotter/scatter.go
```diff
@@ -16,6 +16,10 @@ type Scatter struct {
 	// XYs is a copy of the points for this scatter.
 	XYs
 
+	//GlyphStyleFunc, if not nil, specifies GlyphStyles
```
sbinet: please leave a space between `//` and the start of the comment.

plotter/scatter.go
```diff
@@ -16,6 +16,10 @@ type Scatter struct {
...
+	//for individual points
```
sbinet: here as well.

plotter/scatter.go
sbinet: my comment above is still relevant, I believe.

plotter/scatterColor_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+	"gonum.org/v1/plot"
+	"gonum.org/v1/plot/internal/cmpimg"
+	"gonum.org/v1/plot/palette/moreland"
+	"gonum.org/v1/plot/vg"
+	"gonum.org/v1/plot/vg/draw"
+	"gonum.org/v1/plot/vg/vgimg"
+)
+
+// ExampleScatterColor draws some scatter points, a line,
```
sbinet: please rename this as: `ExampleScatter_color` so it looks well on godoc (and is part of the examples for `Scatter`.)

plotter/scatterColor_test.go
```diff
@@ -0,0 +1,119 @@
...
+// and a line with points.
+func ExampleScatterColor() {
```
sbinet: please rename this as: `ExampleScatter_color` so it looks well on godoc (and is part of the examples for `Scatter`.)

plotter/scatterColor_test.go
```diff
@@ -0,0 +1,119 @@
...
+	sc.GlyphStyleFunc = func(i int) draw.GlyphStyle {
+		c, err := colors.At(z[i])
+		if err != nil {
+			log.Panic(err)
+		}
+		return draw.GlyphStyle{Color: c, Radius: vg.Points(3), Shape: draw.CircleGlyph{}}
+	}
+
+	p.Add(sc)
+	//p.Legend.Add("",sc)
```
sbinet: that comment could be removed, I believe.

plotter/scatter.go
Scorpiokat: When I make it `nil`, I got a warning about a unability to use it in this case.

And a lot of tests (timeseries, for instance) use `DeafultGlyphStyle`, so a lot of tests fail in this case.

sbinet: I am talking about `GlyphStyleFunc` being `nil`. not `GlyphStyle` (which is indeed set by default to `DefaultGlyphStyle`.)

I believe that, if you leave the whole `NewScatter` unmodified with regard to what it was before, then everything should work well.

ie: `NewScatter` should look like:
```go
// NewScatter returns a Scatter that uses the
// default glyph style.
func NewScatter(xys XYer) (*Scatter, error) {
	data, err := CopyXYs(xys)
	if err != nil {
		return nil, err
	}
	return &Scatter{
		XYs:        data,
		GlyphStyle: DefaultGlyphStyle,
	}, err
}
```

Scorpiokat: Ooops, sorry. Did not understand that. Now it works with the original `NewScatter`.

plotter/scatterColor_test.go
sbinet: still there :)
also, the comment doesn't match what the example actually does :)

perhaps something along these lines would do ?
```go
// ExampleScatter_color draws some scatter points.
// Each point will be plotted with a different color depending on some external criteria.
func ExampleScatter_color() {
```

(it could be improved, though.)

plotter/scatterColor_test.go
```diff
@@ -0,0 +1,118 @@
...
+)
+
+// ExampleScatterColor draws some scatter points, a line,
+// and a line with points.
+func ExampleScatter_color() {
+	rnd := rand.New(rand.NewSource(1))
+
+	// randomPoints returns some random x, y points
+	// with some interesting kind of trend.
+	randomPoints := func(n int) XYs {
```
ctessum: Would it make sense for this function to return `XYZs` instead of `XYs`? Then we wouldn't need to separately create z values below.

plotter/scatterColor_test.go
```diff
@@ -0,0 +1,118 @@
...
+		log.Panic(err)
+	}
+
+	colors := moreland.Kindlmann() // Initialize a color map.
+	colors.SetMax(255)
+	colors.SetMin(0)
+
+	z := []float64{31, 41, 51, 61, 71, 81, 91, 101, 111, 121, 131, 141, 151, 161, 171, 181}
+
+	sc.GlyphStyleFunc = func(i int) draw.GlyphStyle {
```
ctessum: Maybe add a comment describing what this is doing?

plotter/scatterColor_test.go
```diff
@@ -0,0 +1,118 @@
...
+		c, err := colors.At(z[i])
+		if err != nil {
+			log.Panic(err)
+		}
+		return draw.GlyphStyle{Color: c, Radius: vg.Points(3), Shape: draw.CircleGlyph{}}
+	}
+
+	p.Add(sc)
+
+	// Create a legend.
```
ctessum: I think [this](https://godoc.org/gonum.org/v1/plot/plotter#ex-ColorBar--Vertical) would be a better way to create the legend.

ctessum: So for example:

``` go
pLegend, err := plot.New()
if err != nil {
    log.Panic(err)
}
l := &ColorBar{ColorMap: colors}
l.Vertical = true
pLegend.Add(l)
pLegend.HideX()
pLegend.Y.Padding = 0
pLegend.Title.Text = "Title"
```

Then, below:

``` go

dcPlot := draw.Crop(dc, 0,-legendWidth, 0, 0)
dcLegend := draw.Crop(dc, 300-legendWidth, 0, 0, 0)
pLegend.Draw(dcLegend)
p.Draw(dcPlot)
```

ctessum:
One additional outstanding issue is that since the `GlyphStyle` also determines the size of the glyph, that needs to be accounted for in the `GlyphBoxes` method. You can see [how this is done in the Bubbles plotter](https://github.com/gonum/plot/blob/master/plotter/bubbles.go#L106).

One way to test that this is implemented correctly would be to add a `ExampleScatter_bubbles` example which re-creates the `ExampleBubbles` using the `GlyphStyleFunc`. Then, if others are comfortable with it, we can delete the whole `Bubbles` plotter to reduce the size of the API.

plotter/scatterColor_test.go
Scorpiokat: This solution is a bit confusing, as it adds the legend (and the values which our `z[i]` is taking) on the Y axis (or X axis in case of the `l.Horizontal` option). So,I kept the previous solution for now.

plotter/scatterColor_test.go
Scorpiokat: In this case we will have to include `Z` to the type `Scatter`, won't we?

Scorpiokat:
Addition of `ExampleScatter_bubbles` will cause us to add `Z` to the type `Scatter` anyway, right? At least, we will need it when implementing `GlyphBoxes` method.

Scorpiokat:
I have added `bubble_new_test` (I know that name is not perfect, I will rename it if we decide to keep it), so that we can use `GlyphStyleFunc` logic. And replace the previous one. Please have a look.

Actually I do not know now how to fix the `GlyphBoxes` issue in this case. Do you have any suggestions?

ctessum:
I think this might be a way to do the GlyphBoxes:

``` go
func (pts *Scatter) GlyphBoxes(plt *plot.Plot) []plot.GlyphBox {
        glyph := func(i int) draw.GlyphStyle { return pts.GlyphStyle }
        if pts.GlyphStyleFunc != nil {
           glyph = pts.GlyphStyleFunc
        }
	bs := make([]plot.GlyphBox, len(pts.XYs))
	for i, p := range pts.XYs {
		bs[i].X = plt.X.Norm(p.X)
		bs[i].Y = plt.Y.Norm(p.Y)
		r := glyph(i).Radius
	        bs[i].Rectangle = vg.Rectangle{
			Min: vg.Point{X: -r, Y: -r},
			Max: vg.Point{X: +r, Y: +r},
		}
		bs[i].Rectangle = pts.GlyphStyle.Rectangle()
	}
	return bs
}
```

ctessum:
Ideally we would want to do this without changing the bubbles test image at all. To do so we would need to add the [bubble radius calculator](https://github.com/gonum/plot/blob/master/plotter/bubbles.go#L89) as part of the `GlyphStyleFunc` in the example.

Scorpiokat:
The current logic of calculating radius depends on i which can be a function.
And [the one specified by you](https://github.com/gonum/plot/blob/master/plotter/bubbles.go#L89) needs the maximum and minimum radii, and so on. I might not get what you were trying to say though.

plotter/bubbles_new_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+	sc, err := NewScatter(scatterDataNew)
+	if err != nil {
+		log.Panic(err)
+	}
+
+	//Specify style for individual points.
+	sc.GlyphStyleFunc = func(i int) draw.GlyphStyle {
+		c := color.RGBA{R: 196, B: 128, A: 255}
+		r := float64(i)
```
ctessum: This can calculate the radius
``` go
        var minRadius, maxRadius = vg.Points(1), vg.Points(20)
	rng := maxRadius - minRadius
 	d := (scatterData.Z(i) - minZ) / (maxZ - minZ)
 	r:= vg.Length(d)*rng + minRadius
 ```

File: plotter/bubbles_new_test.go [hunk 21]
```diff
@@ -0,0 +1,85 @@
...
+
+// ExampleScatter_bubbles draws some scatter points.
+// Each point is plotted with a different radius size depending on
+// some external criteria.
+func ExampleScatter_bubbles() {
+	rnd := rand.New(rand.NewSource(1))
+
+	// randomPoints returns some random x, y points
+	// with some interesting kind of trend.
+	randomPoints := func(n int) XYs {
```
ctessum: This function can be replaced with the one currently in `ExampleBubbles`, which returns `XYZs` instead of `XYs`.

plotter/bubbles_new_test.go
```diff
@@ -0,0 +1,85 @@
...
+				pts[i].X = pts[i-1].X + rnd.Float64()
+			}
+			pts[i].Y = pts[i].X + 10*rnd.Float64()
+		}
+		return pts
+	}
+
+	n := 15
+	scatterDataNew := randomPoints(n)
+
```
ctessum:
``` go
// Calculate the range of Z values. 
 minZ, maxZ := math.Inf(1), math.Inf(-1)
for _, xyz := range XYZs {
  if xyz.Z > maxZ { maxZ = xyz.Z }
  if xyz.Z < minZ { minZ = xyz.Z
}
```

plotter/scatter.go
```diff
@@ -38,11 +42,26 @@ func NewScatter(xys XYer) (*Scatter, error) {
...
+	glyph := func(i int) draw.GlyphStyle { return pts.GlyphStyle }
+	if pts.GlyphStyleFunc != nil {
+		glyph = pts.GlyphStyleFunc
+	}
+	for i, p := range pts.XYs {
+		c.DrawGlyph(glyph(i), vg.Point{X: trX(p.X), Y: trY(p.Y)})
 	}
 }
 
+/*
```
ctessum: This can be deleted

plotter/bubbles_new_test.go
Scorpiokat: Then `sc, err := NewScatter(scatterData)` will cause change in `Scatter` type.

ctessum: Can we add an `XY()` method to `XYZs` so that it implements the `XYer` interface in addition to the `XYZer` interface?

plotter/plotter.go
```diff
@@ -196,6 +199,11 @@ func (xyz XYZs) XYZ(i int) (float64, float64, float64) {
 	return xyz[i].X, xyz[i].Y, xyz[i].Z
 }
 
+// XY implements the XY method of the XYZer interface.
```
ctessum: XYer interface

plotter/plotter.go
```diff
@@ -181,6 +181,9 @@ type XYZer interface {
 
 	// XYZ returns an x, y, z triple.
 	XYZ(int) (float64, float64, float64)
+
+	// XY returns a pair of x and y.
```
ctessum: Maybe "XY returns an x, y pair"?

plotter/bubbles_test.go
```diff
@@ -66,29 +42,61 @@ func ExampleBubbles() {
...
+			minZ = xyz.Z
+		}
+	}
 
 	p, err := plot.New()
 	if err != nil {
 		log.Panic(err)
 	}
-	p.Title.Text = "Bubbles"
+	p.Title.Text = "Bubble Plot"
```
ctessum: If this is changed to "Bubbles", then the reference image should hopefully not change

plotter/bubbles_test.go
```diff
@@ -66,29 +42,61 @@ func ExampleBubbles() {
...
-	bs.Color = color.RGBA{R: 196, B: 128, A: 255}
-	p.Add(bs)
 
-	err = p.Save(200, 200, "testdata/bubbles.png")
+	// Specify style for individual points.
+	sc.GlyphStyleFunc = func(i int) draw.GlyphStyle {
+		c := color.RGBA{R: 196, B: 128, A: 255}
+		var minRadius, maxRadius = vg.Points(1), vg.Points(20)
+		rng := maxRadius - minRadius
+		d := (float64(i)*0.8 - minZ) / (maxZ - minZ) // float64(i)*0.8 is just a random function (i)
```
ctessum: Why?

plotter/bubbles_test.go
```diff
@@ -66,29 +42,61 @@ func ExampleBubbles() {
...
+		r := vg.Length(d)*rng + minRadius
+		//r := float64(i)
```
ctessum: Delete

plotter/bubbles_test.go
ctessum: why not:

``` go 
_, _, z := scatterData.XYZ(i)
d := (z-minZ) / (maxZ - minZ)
```

plotter/bubbles_test.go
```diff
@@ -76,19 +63,40 @@ func ExampleBubbles() {
...
+		var minRadius, maxRadius = vg.Points(1), vg.Points(20)
+		rng := maxRadius - minRadius
+		_, _, z := scatterData.XYZ(i)
+		d := (z-minZ) / (maxZ - minZ)
+		r := vg.Length(d)*rng + minRadius
+		return draw.GlyphStyle{Color: c, Radius: r, Shape: draw.CircleGlyph{}}
+	}
+	p.Add(sc)
+
+	img := vgimg.New(200, 200)
```
ctessum: This doesn't need to be changed from `err = p.Save(200, 200, "testdata/bubbles.png")` as far as I can tell

plotter/scatterColor_test.go
ctessum: I think we can change it to be the same `randomTriples` function that is in the bubbles example.

plotter/scatterColor_test.go
```diff
@@ -0,0 +1,121 @@
...
+			} else {
+				pts[i].X = pts[i-1].X + rnd.Float64()
+			}
+			pts[i].Y = pts[i].X + 10*rnd.Float64()
+		}
+		return pts
+	}
+
+	n := 15
+	scatterDataNew := randomPoints(n)
```
ctessum: Maybe `scatterData` instead of `scatterDataNew`?

plotter/scatterColor_test.go
```diff
@@ -0,0 +1,121 @@
...
+		c, err := colors.At(z[i])
+		if err != nil {
+			log.Panic(err)
+		}
+		return draw.GlyphStyle{Color: c, Radius: vg.Points(3), Shape: draw.CircleGlyph{}}
+	}
+	p.Add(sc)
+
+	// Create a legend.
+	thumbs := PaletteThumbnailers(colors.Palette(n))
```
ctessum: I think this would work better with a palette.ColorMap instead of a palette.Palette. I think I put some example code for that somewhere in this thread.

Scorpiokat: I have actually replied about this issue - here is the problem I am facing - colors are taking the Y axis values which is wrong, I am sure. 
scattercolor

plotter/bubbles_test.go
```diff
@@ -7,46 +7,20 @@ package plotter
...
-		if r := b.radius(test.z); r != test.r {
-			t.Errorf("Got incorrect radius (%g) on %v", r, test)
-		}
-	}
-}
-
-func ExampleBubbles() {
+// ExampleScatter_bubbles draws some scatter points.
+// Each point is plotted with a different radius size depending on
+// some external criteria.
```
ctessum: delete "some"

kortschak:
I'll have a look over the next few days.

plotter/scatterColor_test.go
```diff
@@ -6,27 +6,24 @@ package plotter
...
 	"gonum.org/v1/plot"
 	"gonum.org/v1/plot/internal/cmpimg"
 	"gonum.org/v1/plot/palette/moreland"
 	"gonum.org/v1/plot/vg"
 	"gonum.org/v1/plot/vg/draw"
-	//"gonum.org/v1/plot/vg/vgimg"
 )
 
-// ExampleScatter_color draws some scatter points.
+// ExampleScatter_color draws a coloured scatter plot.
```
ctessum: color/colour consistency

ctessum:
Actually, the tests are still failing so that needs to be fixed too. I think the problem is in the `gob` subpackage.

plotter/scatterColor_test.go
kortschak: All public facing natural language should be written in American Standard English unless there is a good reason otherwise; `s/coloured/colored/`. Internal comments should be written in English.

kortschak:
[This line](https://github.com/gonum/plot/blob/24e75db66036fd0c4301651f8cf1c318492d1f6a/gob/gob.go#L32) needs to go away.

Scorpiokat:
Tests are still failing, for some reason.

kortschak:
You need to regenerate the golden image for the scatter plotter tests.

Scorpiokat:
I got that and it's got regenerated. Never mind. Will try that once again.

Scorpiokat:
In my system the files match, so I cannot find a problem. Could you @kortschak and @ctessum and @sbinet please have a look at the stat information?

kortschak:
That test does not fail for me locally.

Scorpiokat:
Is that the way of how the particular system creates the file? Can we improve this test?

ctessum:
Probably, or different versions of the `draw2d` upstream library.

We've gone through problems like this before and ended up with a system that works, e.g.:

``` go
func TestScatter_color(t *testing.T) {
     cmpimg.CheckPlot(ExampleScatter_color, t, "scatterColor.png")
}
```

Is there a reason that you implemented a different solution here?

Scorpiokat:
@ctessum this is the same solution, I have just added some stats to it in the last commit to show that the system builds a file that differs in size from the golden one. However, on the systems I have tested the code, these 2 files are identical, so the test does not fail. I will remove these lines once we find out why that happens.

ctessum:
If you do go get -u ./... on your system to update all the dependencies,
does the test still pass?

Scorpiokat:
@ctessum yes. On this machine, and on a new freshly made one.

kortschak:
I get perfect a perfect match on my machine (test passes and sha1 matches). A suggestion I have is to render the png to base64 (we have a helper for this). Then at least we can see what the image looks like. Additionally, you can make changed in cmpimg to see in detail why it's failing the check.

Scorpiokat:
@kortschak and what sha1 are you getting on your machine? Could you please send that to me? And that would be really helpful if I could have got the image that the system is generating, otherwise I cannot test it against my result. just by making commits.

kortschak:
The sha1 I get is f0217b5efc20f13c2c43a4e5a760b23eef9d1eec.

What you can do is use encoding/base64 to output a text representation of the file to the travis log. Then you can reconvert that to a binary after scraping it from the log.

Scorpiokat:
@kortschak I have done that. The base64 representations are different. However, when reconverted to binary - they look the same, while sha1sum says otherwise. Looks like cmpimg needs to be inspected.

kortschak:
I have compared the two images and they are nearly the same, but there is a very slight difference in the colour spectrum at the bottom of the plot (not visible with a straight subtraction of images, but very clear when scaled by ~25x).

scattercolor_fail

scattercolor_diff

What we can see from this is that the failure is entirely from the colourbar (this is easy to check by removing that code and seeing if the failure persists). What I suspect is the underlying cause is floating point differences between the arch we are running and the arch travis is using. We've seen things like this before in Gonum proper (lapack being a good example). I don't know why it's affecting the bar and not the scatter dots, but I suspect it is the bar image scaling. If this is the problem, the obvious solution is to remove the colorbar.

Scorpiokat:
@kortschak thanks a lot for checking the images! I haven't noticed that difference in colour. Changing ColorBar back to PaletteThambnailers definitely worked for the best now. Please have a look.

ctessum:
LGTM, but remember to squash all the commits into one, or perhaps two commits: one for everything except for the last commit here, and then a second one where the colorbar is changed to the palette thumbnails.

kortschak:
It's also worth rebasing onto master (there are conflicts that will make this difficult though) since default ticks behaviour has changed.

I'd rebase -i into the structure that you want, then attempt the rebase onto master and fix any conflicts that arise.

kortschak:
There's still a lot going on in these commits that is not relevant to the actual change (left over from the debugging). To avoid a lot of messing around, I think we should just merge this into one commit. However, it still needs to be rebased onto master; the failure you see is due to a disagreement between the old and the new default ticks.

So, can you try to rebase onto the gonum plot master and then regen the golden image and push it back here.

Scorpiokat:
Dan @kortschak I have regenerated files, but the one that got regenerated is just one which now is conflicting. ScatterColor_golden has passed the tests on my machine, i.e. was not regenerated. I guess, the Travis check will be failing again once the existing conflict is resolved.

kortschak:
Travis is correct here, the current gonum/plot master is not in the history of this PR. To get this, you need to rebase onto gonum/plot's master; your master needs to be made to match the gonum/plot master.

Unfortunately you have squashed the two commits together prior to thie merge of your master (I must have been unclear - we squash when we merge). I've gone through the reflog to find that.

```
git checkout master
git fetch git@github.com:gonum/plot.git master
git reset --hard FETCH_HEAD
git checkout newfeature
git reset --hard 0e1f7c64cf5940fd8f59ef0254e3913c9ee66b12 # Your previous state.
git rebase master
```

check everything is OK

```
cd plotter
go test -regen
git add testdata/scatterColor_golden.png
git commit -m "fix scatterColor_golden.png"
git push -f origin newfeature
```

Scorpiokat:
Thank you for such a detailed reply. However, not everything is still OK:
```
git fetch git@github.com:gonum/plot.git master
Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

kortschak:
Oh, sorry, that requires, auth use `git fetch https://github.com/gonum/plot.git master` instead.

Scorpiokat:
We decided to leave the ColorBar issue for later, right? Or...?

kortschak:
You mean the float errors?

kortschak:
We have 25 commits back, sorry. I am just going to leave this and squash them in the merge.

Scorpiokat:
Yes, I meant that.

kortschak:
Without knowing what is going on on their hardware, there is not much we can do.

Scorpiokat:
Sure.
Thank you. I can delete the `newfeature` branch, I guess.

kortschak:
I should have checked. That was not rebased onto the gonum/plot master and was passing because the old default ticks algorithm was still being used. I have sent a PR fixing the breakage.

Scorpiokat:
Sorry and thanks a lot.

Issue #397, opened 2017-10-25
Package import with `go get gonum.org/v1/plot/...` is broken — 404 returned.


denisbrodbeck:
I am unable to import this project with `go get` or `dep`, which never failed me before.

Call to `Parsing meta tags from https://gonum.org/v1/plot?go-get=1 (status code 404)` fails.
The page [https://www.gonum.org/v1/plot?go-get=1](https://www.gonum.org/v1/plot?go-get=1) returns `Page not found`.

Output:
```bash
$ go get -v gonum.org/v1/plot/...

Fetching https://gonum.org/v1/plot?go-get=1
Parsing meta tags from https://gonum.org/v1/plot?go-get=1 (status code 404)
get "gonum.org/v1/plot": found meta tag get.metaImport{Prefix:"gonum.org/v1/plot", VCS:"git", RepoRoot:"https://github.com/gonum/plot"} at https://gonum.org/v1/plot?go-get=1
gonum.org/v1/plot (download)
Fetching https://gonum.org/v1/plot/vg?go-get=1
Parsing meta tags from https://gonum.org/v1/plot/vg?go-get=1 (status code 404)
get "gonum.org/v1/plot/vg": found meta tag get.metaImport{Prefix:"gonum.org/v1/plot", VCS:"git", RepoRoot:"https://github.com/gonum/plot"} at https://gonum.org/v1/plot/vg?go-get=1
get "gonum.org/v1/plot/vg": verifying non-authoritative meta tag
Fetching https://gonum.org/v1/plot?go-get=1
Parsing meta tags from https://gonum.org/v1/plot?go-get=1 (status code 404)
github.com/golang/freetype (download)
Fetching https://golang.org/x/image/math/fixed?go-get=1
Parsing meta tags from https://golang.org/x/image/math/fixed?go-get=1 (status code 200)
get "golang.org/x/image/math/fixed": found meta tag get.metaImport{Prefix:"golang.org/x/image", VCS:"git", RepoRoot:"https://go.googlesource.com/image"} at https://golang.org/x/image/math/fixed?go-get=1
get "golang.org/x/image/math/fixed": verifying non-authoritative meta tag
Fetching https://golang.org/x/image?go-get=1
Parsing meta tags from https://golang.org/x/image?go-get=1 (status code 200)
golang.org/x/image (download)
Fetching https://golang.org/x/image/font?go-get=1
Parsing meta tags from https://golang.org/x/image/font?go-get=1 (status code 200)
[...]

PR #398, opened 2017-10-25
Add doc for dependency on `hg/mercurial`.

denisbrodbeck:
see #397


kortschak:
I'd rather not add this. It won't be true in the future and in general, git and mercurial are sort of expected for the Go ecosystem (bzr also to a much lesser degree)

/cc @eaburns

denisbrodbeck:
Sure, no problem.

Issue #389, opened 2017-09-20
vg/vgpdf: switch to a backend that supports font embedding


sbinet:
we should (probably) consider switching away from `bitbucket.org/zombiezen/gopdf/pdf` that isn't maintained much.

and probably switch to `github.com/jung-kurt/gofpdf`.

4 reasons:

- it is maintained
- it is stable
- it is on github (so we don't require a mercurial client)
- it supports fonts embedding

the last one is a rather nice feature and allows for better PDF compatibility (w/o relying on fonts metrics being the exact same than the `Courier`, `Helvetica` and `Times` ones.)

sbinet:
@eaburns @kortschak do you agree ?

eaburns:
Yes, this sounds like a great change.

PR #401, opened 2017-10-27
vg/fonts: update base URL of Liberation fonts

sbinet:
The fedorahosted.org site has been retired (since March 2017.)
This CL now downloads fonts from the new location, pagure.org.

Fixes gonum/plot#400.

PR #402, opened 2017-10-27
vg/vgpdf: migrate to jung-kurt/gofpdf 

sbinet:
This CL drops bitbucket/zombiezen/gopdf/pdf in favour of
github.com/jung-kurt/gofpdf.

Fixes gonum/plot#389.


sbinet:
needs gonum/plot#401.

sbinet:
also 2 noteworthy items:

- the way `gofpdf` handles and registers fonts is a bit cringy b/c the API is file-oriented. this results in the `liberation_fonts_generated.go` file being waaay too big. before merging this in, we should fixup the `gofpdf` API to directly deal with the `[]byte` (or `io.Reader`) of the `ttf` font.

- embedding fonts results in bigger PDF files e.g.: 
```
  1.5K Oct 27 16:39 plotter/testdata/polygon_holes_golden.pdf (no embedding)
210.0K Oct 27 16:32 plotter/testdata/polygon_holes_golden.pdf (w/ embedding)
```
perhaps we should give the user the choice to do embedding or not (with a field on the `vg/vgpdf.Canvas` value)

ctessum:
The panic statements aren't ideal but I don't see a great way to improve that. Would those panics occur if a user tried to use the wrong font?

I think one way to deal with file size is to only embed the characters that are used in the document. Is that what `gofpdf` does? I feel like an option to choose whether fonts are embedded or not may add more complexity than usefulness.

sbinet:
Please don't merge just yet :)

kortschak:
>the way gofpdf handles and registers fonts is a bit cringy b/c the API is file-oriented. this results in the liberation_fonts_generated.go file being waaay too big. before merging this in, we should fixup the gofpdf API to directly deal with the []byte (or io.Reader) of the ttf font.

What do you propose to do about this?

vg/vgpdf/vgpdf.go
```diff
@@ -25,123 +29,230 @@ const DPI = 72
...
+		// not defined (standard font?), use average of 81%
+		top = 0.81 * h
+	} else {
+		top = -float64(d.Ascent) * h / float64(d.Ascent-d.Descent)
+	}
+	return 0, top, c.pdf.GetStringWidth(txt), top + h
+}
+
+// whitespace determines if a rune is whitespace
+func whitespace(r rune) bool {
```
kortschak: These are not the only whitespace characters. Is there not a whitespace check function in "utf8"?

vg/vgpdf/vgpdf.go
```diff
@@ -25,123 +29,230 @@ const DPI = 72
...
+
+// font registers a font and a size with the PDF canvas.
+func (c *Canvas) font(fnt vg.Font, pt vg.Point) {
+	if _, ok := c.fonts[fnt.Name()]; ok {
+		return
+	}
+	if n, ok := vg.FontMap[fnt.Name()]; ok {
+		json, err := fonts.Asset(n + ".json")
+		if err != nil {
+			log.Panicf("vgpdf: could not load JSON description for TTF font %q: %v", n+".json", err)
```
kortschak: I'm really not very happy with this.

vg/vgpdf/vgpdf.go
```diff
@@ -25,123 +29,230 @@ const DPI = 72
...
-	c.page.DrawImage(img, r)
+	log.Panicf("vgpdf: not implemented")
 }
 
-// pdfPath returns a pdf.Path from a vg.Path.
-func pdfPath(c *Canvas, path vg.Path) *pdf.Path {
-	p := new(pdf.Path)
+// pdfPath processes a vg.Path and applies it to the canvas.
+func (c *Canvas) pdfPath(path vg.Path, style string) {
+	var (
```
kortschak: `var xp, yp float64`

vg/vgpdf/vgpdf.go
```diff
@@ -229,12 +340,26 @@ func (w *writerCounter) Write(p []byte) (int, error) {
...
 	b := bufio.NewWriter(&wc)
-	if err := c.doc.Encode(b); err != nil {
+	if err := c.pdf.Output(b); err != nil {
 		return wc.n, err
 	}
 	err := b.Flush()
 	return wc.n, err
 }
+
+const (
```
kortschak: `const c255 = 255.0 / 65535.0`

sbinet:
Upon further consideration, I think I will split this CL in two.
I'll extract the part about the migration to the new PDF package and submit that.
But leave the part about the font embedding for later, when the PDF library API won't be so 'file oriented'.

Issue #403, opened 2017-10-30
plot: fix locations of default ticks


kortschak:
This is a continuation of #393.

I'd like to look into [this implementation](vis.stanford.edu/files/2010-TickLabels-InfoVis.pdf), but in the mean time, change the location of the tick marks.

sbinet:
adding the linked file here as well: I was only able to look at it / download it from my phone but it 404-ed from chromium on my desktop...

[2010-TickLabels-InfoVis.pdf](https://github.com/gonum/plot/files/1426527/2010-TickLabels-InfoVis.pdf)

kortschak:
[Here](https://play.golang.org/p/Ozohq2QdCP) is a Go implementation (from the paper and the R code in the labelling package - MIT + file LICENSE | Unlimited according to the R documentation) with modifications to allow capturing the magnitude of the values and demonstrating how this is used.

Please have a play with values of dMin, dMax and m.

I think with this and a log-space equivalent we can resolve the tick labelling situation into a single type again.

PR #404, opened 2017-10-30
plot: place ticks where we say we do

kortschak:
Please take a look.

/cc @decibelcooper

Updates #403.


decibelcooper:
This would certainly be an improvement.  However, I also think that it is a bandaid.  There are two problems that I see with the code that this covers up:
1. The use of `suggestedTicks` instead of `(suggestedTicks - 1)` on lines 365 and 370 unnecessarily drops the tick intervals to a lower order of magnitude in certain circumstances.
2. In these circumstances, the call to `adjustPrecision()` then aggressively rounds, counteracting the decreased tick intervals.

In my opinion, `adjustPrecision()` should not be used, and one should consider replacing `DefaultTicks` with `PreciseTicks`.

ctessum:
I agree that this change does what it says it does, and that it is an improvement over the previous behavior. I don't think any comment text needs to be added to explain the change from the previous version because the previous version was wrong.

I'm not able to provide a useful opinion regarding whether this is better or worse than `PreciseTicks`, but my feeling is that they are similar enough that in the long term we should only keep one of them. This is especially true after this change because the default ticks will also be "precise", if by "precise" we mean "not wrong".

kortschak:
@sbinet The change is pretty straightforward; instead of being a case of `b = fn(a); use(b, a)` it is `a = fn(a); use(a, a)`.

@decibelcooper Please read the underlying issue. This is a minor part of that - to fix an error. The full fix is to change the algorithm completely. I don't think that changing from the `DefaultTicks` implementation to the `PreciseTicks` implementation is the correct change for reasons that I outlined in the original issue; we should not put ourselves in the position that floating point rounding effects impact on the plot beyond what is reasonable. Not rounding guarantees that that happens for fairly commonly used floating point tick values.

decibelcooper:
@kortschak I have read the issue.  What I'm trying to say is that there are more errors that this only covers up.  You clearly have not looked at the code for PreciseTicks in detail.  It does round.

decibelcooper:
https://github.com/gonum/plot/blob/16d6d3ecc1a6eacb83102b67b0159c561812d9d0/axis.go#L455-L461

kortschak:
This PR updates the underlying issue; it's not marked as a fix. The approached used in `DefaultTicks` and `PreciseTicks` are based on fundamentally difference design intentions. The rationale behind `DefaultTicks` is to provide a reasonable set of labelled tick marks with uniform label precision that minimises the visual clutter. It's pretty clear that it's not correct, but `PreciseTicks` will not satisfy these criteria as, AFAIU, you don't want that.

Yes, `PreciseTicks` rounds the values, but not the printed representation (the `-1` for `formatFloatTick`'s `prec` parameter). This leads to a discordance between the number of printed places and the float representation for some sequences of values.

decibelcooper:
> PreciseTicks will not satisfy these criteria as, AFAIU, you don't want that.

Incorrect

> Yes, PreciseTicks rounds the values, but not the printed representation (the -1 for formatFloatTick's prec parameter). This leads to a discordance between the number of printed places and the float representation for some sequences of values.

If this leads to "a discordance between the number of printed places and the float representation for some sequences of values", then exactly the same is true for DefaultTicks: https://github.com/gonum/plot/blob/16d6d3ecc1a6eacb83102b67b0159c561812d9d0/axis.go#L387-L391
I use `-1` in `formatFloatTicks()` because if I specify the precision that I actually want, it has very different behavior as far as deciding when to switch to an exponent notation, and tends to increase the number of characters needed to describe numbers with low precision.  Regardless, this is identical to `DefaultTicks`.  If you believe using `-1` for precision in `formatFloatTicks()` is wrong, this belief is independent of the choice between the current `DefaultTicks` and `PreciseTicks`, and is therefore irrelevant.

kortschak:
>> PreciseTicks will not satisfy these criteria as, AFAIU, you don't want that.

>Incorrect

That's unhelpful; there are two claims in the quoted sentence. Which is incorrect?

> Regardless, this is identical to `DefaultTicks`.  If you believe using `-1` for precision in `formatFloatTicks()` is wrong, this belief is independent of the choice between the current `DefaultTicks` and `PreciseTicks`, and is therefore irrelevant.

Yes-ish. Finding the correct print precision is not done correctly in `DefaultTicks` - it was seen as too difficult at the time (I'll point again to the original comment in this PR that this is an update, not a fix). The significant difference between `DefaultTicks` and `PreciseTicks` is that the former makes a uniform precision cut based on what will be the *minimum* precision required to make unique tick labels. That was the design motivation.

I don't know if this is just my interpretation, but your comments here are to my view pretty abrasive. I /cc'd you because I know you have an interest in this and I wanted to get your constructive input. If my comments have prompted this, my apologies.

decibelcooper:
@kortschak `PreciseTicks` was always intended as a replacement for `DefaultTicks`, but as suggested I put the code in as an alternative ticker so that I could provide myself and others that have similar trouble with `DefaultTicks` an alternative.  `PreciseTicks` is also intended to minimize clutter, it just does it more correctly.  It can replace `DefaultTicks` directly as a _fix_.

In hindsight, putting the code in as an alternative and not as a fix was a mistake: it has convoluted the intentions behind it, and made it difficult to see what changes were actually made.  I genuinely don't want my comments to come off as abrasive, but I must admit some frustration with how you are handling this issue.  I will now drop this discussion, since it's clearly going nowhere.  `PreciseTicks` makes gonum/plot usable again, so I will continue using it rather than `DefaultTicks`.

PR #405, opened 2017-10-30
plotter: check for negative Z-range

### LINKED ISSUE #359: plotter: Contour does not check Z range
This should match the behaviour of HeatMap.

kortschak:
Please take a look.

Fixes #359.

### DIFF
--- a/plotter/contour.go
+++ b/plotter/contour.go
@@ -50,6 +50,8 @@ type Contour struct {
 // the 0.01, 0.05, 0.25, 0.5, 0.75, 0.95 and 0.99 quantiles.
 // If g has Min and Max methods that return a float, those returned
 // values are used to set the respective Contour fields.
+// If the returned Contour is used when Min is greater than Max, the
+// Plot method will panic.
 func NewContour(g GridXYZ, levels []float64, p palette.Palette) *Contour {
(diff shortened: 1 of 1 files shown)


plotter/contour.go
```diff
@@ -122,6 +122,10 @@ const naive = false
 
 // Plot implements the Plot method of the plot.Plotter interface.
 func (h *Contour) Plot(c draw.Canvas, plt *plot.Plot) {
+	if h.Min > h.Max {
```
sbinet: perhaps we should document this in the `Controur` struct or the `NewContour` function ?

kortschak: Thanks. Done.

kortschak:
PTAL

plotter/contour.go
```diff
@@ -122,6 +124,10 @@ const naive = false
...
+		panic("contour: negative Z range")
```
sbinet: instead of `negative Z range` we could perhaps say `invalid Z range` (I could certainly imagine a contour which represents, say, a negative altitude for a seafloor map) ?

kortschak: The string reflects the code used in `HeatMap`. We can change it, but I think in another PR.

PR #407, opened 2017-11-02
plot: fix axis labelling

kortschak:
This goes a long way towards fixing #403, though there are problems that I believe are a result of aggressive cropping of the plot space after the axes have been decided (the axes that show up here with only one major tick mark *do* have the labels created, but the plot does not end up rendering them).

@eaburns Can you suggest anything here?

/cc @decibelcooper


sbinet:
I have just looked at the pretty pictures :)

indeed, there is this issue of cropped axes...
also, the axes seem to be a tad more crowded in terms of the number of minor ticks (there are more of them than previously.)

one other thing to look at would be the additional CPU this new method requires.
well, just to have a rough handle on that quantity and whether this might be an issue in the future or not.

(but of course, first order of business is to make it correct.)

ctessum:
I think the aggressive cropping was a design choice, right? I personally think it's theoretically better to determine the axis range based on the range of the data (like this package does) rather than the range of the labels (like other packages do).

Would it be feasible to have it always put labels at the endpoints of the data range in cases where the endpoint labels wouldn't overlap or crowd the other labels?

kortschak:
@sbinet Yes, the minor ticks are denser, but that is easily modified in the step switch. The CPU cost is unlikely to be significantly different to the current implementation and minor in comparison to the work done by the vg back ends.

@ctessum Yes, it is a design choice, but it seems to me that the max axis extension should happen after the choice has been made, rather than cropping off things that have been used to define presentation. I think I need to work through the process of calling `Plot` to get a better idea of the call graph.

kortschak:
I've had a look at removing the checks for clipping and I think that the omitted ticks and labels should be printed, but I don't understand how the plot area size is determined. @sbinet, you've done a fair bit with that, can you point me in the right direction?

kortschak:
I've just added code changes that prevent axis drawing from cropping. I think overall it's an improvement with a couple of comments:

1. The bar chart test example gets completely broken, but TBH I'm not sad about that since I think it is sort of broken by design; it has two data series on a single axis, where it would be better to split over a couple of tiles. I think this example should be changed (probably anyway even if this change to cropping does not).

2. The reverse palette test does weird things that I've not yet managed to figure out due to scaling.

kortschak:
*ping*

In particular I am unable to make head or tail of how BarChart is failing.

axis.go
```diff
@@ -318,15 +326,14 @@ func (a *verticalAxis) draw(c draw.Canvas) {
...
-				continue
-			}
 			start := t.lengthOffset(len)
 			c.StrokeLine2(a.Tick.LineStyle, x+start, y, x+len, y)
 		}
 		x += len
 	}
 	c.StrokeLine2(a.LineStyle, x, c.Min.Y, x, c.Max.Y)
+
+	return a.Min, a.Max
```
ctessum: Does this mean that user-specified axes ranges are overwritten? That seems problematic.

kortschak: I'm trying to figure out ways to pass the information around that does not do this, but the engineering of the system as it stands is very tight. IMV there are a number of values that get set/passed that mean more than one thing. I'd like to try to sort that out without expanding the API too much.

plotter/functions.go
```diff
@@ -13,8 +13,9 @@ import (
 // Function implements the Plotter interface,
 // drawing a line for the given function.
 type Function struct {
-	F       func(float64) float64
-	Samples int
+	F        func(float64) float64
+	Min, Max float64
```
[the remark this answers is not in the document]
kortschak: Fixed

axis.go
kortschak: >Does this mean that user-specified axes ranges are overwritten? That seems problematic.

Only in the case that a `Ticker` returns marks that are outside the min and max given. If people don't want that, then they can write their own `Ticker` and not use `DefaultTicks`.

kortschak: I have made changes that restore the original user-specified values after using the calculated values. I think the code here can do with a clean-up, but that can happen later.

kortschak:
OK, this is ready for a proper review.

PTAL

plotter/functions.go
```diff
@@ -13,8 +13,14 @@ import (
 // Function implements the Plotter interface,
 // drawing a line for the given function.
 type Function struct {
-	F       func(float64) float64
+	F func(float64) float64
```
ctessum: maybe `func(x float64) (y float64)`?

kortschak:
>What is the current behavior regarding user-specified axis ranges? For example, I would say that palette/testdata/reverse_golden.png is still broken if we're not able to at least manually set the axis range to be 0.5-4.5

The behaviour of `DefaultTicks` is as stated on the can in either state ("it returns a *resonable* default set of tick marks"), but I think overall the new behaviour is an improvement. In cases like the reverse colorbar test, I think a specifically defined `Ticker` is appropriate.

plotter/functions.go
kortschak: Done

ctessum:
> The behaviour of DefaultTicks is as stated...

I guess I usually think of the axis range and the tick locations as somewhat independent (e.g., MS Excel creates defaults for both in a way that's probably similar to this PR but then allows the "bounds" and the "units" to be overridden independently of each other), and they were independent of each other up until this PR.

Would it be possible to maintain (what I understand to be) the old behavior in this regard, which is to use calculated axis ranges (before: based on the data range, now: based on tick locations) unless the user has manually specified the range, in which case the user-specified value would be used instead?

ctessum:
Otherwise, if the user is not allowed to set the range, Max and Min should not be exported fields, correct?

kortschak:
>I guess I usually think of the axis range and the tick locations as somewhat independent (e.g., MS Excel creates defaults for both in a way that's probably similar to this PR but then allows the "bounds" and the "units" to be overridden independently of each other)

The system allows people to do that already, again remember that `DefautlTicks` is just that, a default.

If you want the cropping version, you can get the ticks from `DefaultTicks`, drop the marks you don't want and hand them as a `ConstantTicks`,

>, and they were independent of each other up until this PR.

They still are, just not with `DefaultTicks`.

>Would it be possible to maintain (what I understand to be) the old behavior in this regard, which is to use calculated axis ranges (before: based on the data range, now: based on tick locations) unless the user has manually specified the range, in which case the user-specified value would be used instead?

This complicates the API in ways that aren't necessary since the behaviour can already be achieved.

>Otherwise, if the user is not allowed to set the range, Max and Min should not be exported fields, correct?

No, this only affects `DefaultTicks`. At the moment, we plot all data points that are handed to the `Plot`, this is a consistent and clear behaviour.

kortschak:
This helper type get's you the behaviour you want:

```
type TruncatedTicker struct {
        Min, Max float64
        Ticker
}

func (t TruncatedTicker) Ticks() []Ticks {
        ticks := t.Ticker.Ticks()
        sort.Sort(byValue(ticks))
        var low int
        for i, tick := range ticks {
                if tick.Value < Min {
                        low = i + 1
                }
                if tick.Value > Max {
                        return ticks[low:i]
                }
        }
        return ticks[low:]
}

type byValue []Ticks

func (t byValue) Len() int           { return len(t) }
func (t byValue) Less(i, j int) bool { return t.Value[i] < t.Value[j] }
func (t byValue) Swap(i, j int)      { t[i], t[j] = t[j], t[i] }
```

kortschak:
As a counter example to the Excel behaviour, ggplot2 extends axes to include specified tick marks. I am inclined to towards ggplot2's aesthetic over Excels for most things.

ctessum:
> As a counter example to the Excel behaviour, ggplot2 extends axes to include specified tick marks.

I'm willing to drop this, but to clarify Excel also extends the axes by default, but allows users to override the default behavior. I imagine ggplot is the same. I mentioned Excel because it is the least-common-denominator for plotting in my mind. As you mention, it is also possible to override the default here, but to me it seems like the way to do so is less obvious. (Perhaps a future change could make it more obvious).

I do agree that overall this is an improvement. Perhaps a compromise is to document somewhere (possibly in the `Axis` type documentation) that user-specified Max and Min axis values may be ignored, which I imagine may lead to some confusion otherwise?

kortschak:
I'll wait for at least one other review. ggplot2 does allow truncation (and extension) of plots, but it does this with a variety of options that make for what would be a complicated and opaque API when translated to Go / Gonum/plot (see [here](http://ggplot2.tidyverse.org/reference/lims.html)). The direct translation would be to add a range and domain pair of fields, each with a min and max and use that for calculating the transforms. The issue is how to know when that has been set explicitly - either you need to make the fields pointer values and ignore them when nil, or add a boolean and ignore them when that is false. Neither of these is particularly appealing. If someone can think of a better approach to either of these I'd be very happy to implement it.

kortschak:
I have come up with an alternative solution; the Talbot, Lin and Hanrahan algorithm can be constrained to ensure that the min and max ticks exist within the given data range - the published algorithm does not do this, but does allow constraint such that the data are guaranteed to be within the label range.

kortschak:
PTAL This now constrains labels within min and max and does not alter the plot's axes' ranges. There is a little issue with the sankey plot, but I think we can try to address that later. I would still like to be able to ensure that an axis in in the plot range, but I guess we can address that laler - at least I now know how to go about this.

ctessum:
LGTM. It might be good (perhaps for a future change if not feasible here) to allow the option to either clip the axis range to the data or extend it to fit the ticks, because some of the plots did look better before this commit.

I think other ways to tell if the min and max have been set by the user would be either have the initializer function set them both to NaN, and if they are still NaN at plotting time then use defaults, or to just use defaults if they both equal zero at plotting time.

kortschak:
Thanks.

Yes, the plan for `talbotLinHanrahan` was to use it as the basis for a more generalised Ticker.

NaN is good, ±Inf would work too. This should be an issue after this has been merged.

I'll still wait for one other, @sbinet ?

kortschak:
*ping* @sbinet

axis_test.go
```diff
@@ -5,88 +5,83 @@
...
-			max:  27800.9875,
-			want: []string{"8000", "16000", "24000"},
-		},
-		{
-			min:  55.6545,
-			max:  1555.9875,
-			want: []string{"500", "1000", "1500"},
-		},
-	} {
+	for _, test := range axisSmallTickTests[:1] {
```
sbinet: I don't understand why we only take the first element of `axisSmallTickTests`.
is it a remnant of a debugging session?

sbinet: ah... there was `//TODO(kortschak) Fix precision to make these tests pass` previously.
perhaps add that comment here as well?

kortschak: Ooops I forgot about that. There is something weird going on, I was trying to debug it and the weekend happened. I'll try to sort out what is going on.

kortschak:
Sorry about that, PTAL.

axis.go
```diff
@@ -341,146 +351,103 @@ func (a *verticalAxis) GlyphBoxes(*Plot) (boxes []GlyphBox) {
...
 	}
 
-	majorMult := int(n / suggestedTicks)
-	switch majorMult {
-	case 7:
-		majorMult = 6
-	case 9:
-		majorMult = 8
+	// Choose a reasonable, but ad
+	// hoc formatting for labels.
```
sbinet: this 2-line comment could perhaps be on just 1 line.

axis.go
```diff
@@ -175,12 +175,12 @@ func (LogScale) Normalize(min, max, x float64) float64 {
 // system, normalized to its distance as a fraction of the
 // range of this axis.  For example, if x is a.Min then the return
 // value is 0, and if x is a.Max then the return value is 1.
-func (a *Axis) Norm(x float64) float64 {
+func (a Axis) Norm(x float64) float64 {
```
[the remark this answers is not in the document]
kortschak: In no place except one it the `Axis` mutated. I'm tightening that.

axis.go
kortschak: Style.

kortschak:
Thanks for the review.

Issue #412, opened 2017-11-09
vg: soften third-party dependencies 


fawick:
At the moment, installing `gonum.org/v1/plot/vg/draw` has a hard requirement on various non-vendored third-party packages. 

I'd like to propose a refactoring of the `vg/{draw,vgeps,vgimg,vgpdf,vgsvg}` so that implementations of `vg.Canvas` need to be registered, which the packages could do in their `func init()`.

This would make all third-party dependencies soft requirements and a user might choose not to fulfill some of them in case the respective output formats are not required (e.g. one might want to use PNG but have no need for PDF).

As a side benefit, this would make it easy to incubate new implementations of the `vg.Canvas` interface, as these no longer need to be hardwired to `vg/draw`.

(Incidentally, this would fix issues like #238, but that's not the objective here.)

kortschak:
It would be good to see a plan for your implementation of this.

fawick:
At the moment, I do not have plan, as I am not very familiar with the codebase. But if you are open to the general idea, I might get in deeper into the source and try to devise one in the next couple of weeks.

kortschak:
I'm open to the idea, but it's worth waiting for input from others here. The issues that you may face is handling how `plot.WriterTo` and `plot.Save` work and are documented, but I don't see anything that would make handling those issues impossible.

fawick:
As `plot.WriterTo` calls `draw.NewFormattedCanvas` which returns an error my naive approach would be to simply let that error be non-nil in case the requested format has not been registered, which would then trickle down the stack to `plot.Save` as well. I'd add to the documentation to `plot.Save` and `plot.WriterTo` that trying any non-registered format would lead to such an error.

I wonder whether this an API-breaking change, as on the one hand any using code would still compile while on the other hand the behaviour changes and care must be taken accordingly to import and register any implementations of the `plot.Canvas`interface. Does the fact that there are no release tags have an impact? I'd be genuinely interested in the thoughts of the gonum contributors on these questions.

kortschak:
It will be breaking, but we aer breaking things at the moment. We had intended to freeze the API at the beginning of this year, but delays...

fawick:
@kortschak So far, nobody else chimed in. Should we ping somebody specific to get input from?

sbinet:
I am a bit of two minds here.
[...]

PR #415, opened 2017-11-18
vg/...: Invert image format imports

fawick:
This addresses #412.

Image formats used by vg/draw need to be registered with
draw.RegisterFormat now. For this it was necessary to invert the imports
so that each implemenation also imports vg/draw.

A new package vgall was added which is used to import all image
formats at once. This can be disabled with the build tag "minimal".


vgall.go
```diff
@@ -0,0 +1,7 @@
+// +build !minimal
+
+package plot // import "gonum.org/v1/plot"
+
+import (
```
kortschak: I though we were just going to do the underscore imports here. Can you please do that please; an additional package just to do this seems overly complex.

vgall.go
```diff
@@ -0,0 +1,7 @@
+// +build !minimal
```
kortschak: Also, add a license header here please.

fawick:
Thanks @kortschak  PTAL.

vgall.go
```diff
@@ -0,0 +1,14 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
```
kortschak: 2017

fawick:
Happy to change to uppercase. 

FWIW I plucked this from `plot.go`, so there may be older files that are still claiming 2015 and "gonum"

vg/draw/canvas.go
```diff
@@ -9,14 +9,33 @@ import (
...
-	"gonum.org/v1/plot/vg/vgsvg"
 )
 
+// formats holds the registered canvas image formats
+var formats struct {
+	sync.Mutex
+	m map[string]func(w, h vg.Length) vg.CanvasWriterTo
+}
+
+// RegisterFormat registers an image format for use by NewformattedCanvas.
```
sbinet: `s/NewformattedCanvas/NewFormattedCanvas/`

vg/draw/canvas.go
```diff
@@ -9,14 +9,33 @@ import (
...
+// Name is the name of the format, like "jpeg" or "png".
+// New is the construction function to call for the format.
+func RegisterFormat(name string, fn func(w, h vg.Length) vg.CanvasWriterTo) {
+	formats.Lock()
+	defer formats.Unlock()
+	if formats.m == nil {
+		formats.m = make(map[string]func(w, h vg.Length) vg.CanvasWriterTo)
```
sbinet: couldn't we deal with the initialization of the map when creating the `formats` value?

vg/draw/canvas.go
```diff
@@ -9,14 +9,33 @@ import (
...
+	}
+	if fn == nil {
```
sbinet: this isn't documented. I think we should document that.
but I must say this feels to me like `RegisterFormat` wants to do too much.
what's the use case, actually? removing a format from the map won't un-import the `vgfoo` package.

vg/draw/canvas.go
```diff
@@ -264,36 +283,22 @@ func New(c vg.CanvasSizer) Canvas {
...
-		return nil, fmt.Errorf("unsupported format: %q", format)
+	formats.Lock()
+	defer formats.Unlock()
+	for name, fn := range formats.m {
+		if format == name {
+			return fn(w, h), nil
+		}
 	}
-	return c, nil
+	return nil, fmt.Errorf("unsupported format: %q", format)
```
sbinet: perhaps also print the known, supported formats?

vg/draw/canvas.go
fawick: Done.

vg/draw/canvas.go
fawick: Done.

vg/draw/canvas.go
fawick: Requested by @kortschak, so I like to pass the question for the use case to him.

vg/draw/canvas.go
fawick: Done.

vg/draw/canvas.go
kortschak: The response is a question, "What is the desired behaviour when someone does `RegisterFormat(name, fn)` and `fn` is accidentally `nil`?" If we're OK panicking with that, then I'm fine with removing this, but replace it with a conditional panic here so that it's seen when it happens. The main motivator for me was easing debugging, but that's not a big deal.

sbinet: I'd be fine with panicking too.
`database/sql.Register` does it as well and documents it:
```
$> go doc database/sql.Register 
func Register(name string, driver driver.Driver)
    Register makes a database driver available by the provided name. If Register
    is called twice with the same name or if driver is nil, it panics.
```

kortschak: OK, I would like to be able to call `RegisterFormat` twice though.

sbinet: because you are worried about multiple different packages registering their favorite `vg` backend in their `init` function, which happens to be `"png"` ?
yeah... I guess that could happen.

fawick: Since we have no control over the order in which packages are initialized in the eventual program, the outcome of "which package will provide `"png"`" is uncertain (at least for the formats registered by the `init()`s). A `panic()` might be cleaner, might it not? 
Hmm, OTOH, for that same reason the author of `main()` might have no immediate control over what packages are initialized further down in the dependency tree, and allowing double registrations for the same formats would at least allow the program to run.
To give another example compared to `database/sql`: [Package `image`](https://golang.org/src/image/format.go#26) uses a slice in which the the first registration would always win, but which would never panic. If a registered function is `nil`, the function parsing the formats returns an error (for us, that would be `NewFormattedCanvas`).

Anyway, no matter which way we decide to have implemented here, I agree that I need to mention it in the  documentation.

sbinet: the final user/dev of `main` has perhaps no control over the order of package initialization (except for the one in the `package main`, of course) but the order is stable, single-threaded and reproducible.
so given a set of packages that depend on each others, the order can be known and (somewhat) standardized:
- https://golang.org/ref/spec#Package_initialization

_ie:_ it's not C++ :)

fawick: There have been no further comments, so I am going to
- implement and document panic on `nil`, but
- keep allowing format registrations under the same name.

fawick:
Do we need a `func draw.Formats() []string` that returns a sorted list of registered formats?

kortschak:
I think that's a nice touch.

sbinet:
+1 from my end as well (I was considering suggesting it but preferred to table it for a later PR  :P)

fawick:
Done.

kortschak:
Please rebase onto master to fix the failure.

fawick:
Ah, thanks for the fix in master. I did an rebase half an hour ago but the tests still failed locally, so I ignored to push that. :) Rebased again now.

sbinet:
hi,

hum... it seems we've "kind of" dropped the ball here. flat apologies are in order.

if you're willing to send a PR against the main [gonum/gonum](https://github.com/gonum/gonum) repo adding yourself to the `A+C` files, I am willing to fix the merge conflicts.

deal?

fawick:
@sbinet: :rofl: 

It's so long ago that I cannot even remember what it was I wanted to achieve with this... Sure, if you are still interested in merging I can add myself to A+C

sbinet:
Yes, please.
It's something that'd be useful for 'my' hplot users as well.

vg/draw/canvas.go
```diff
@@ -8,15 +8,45 @@ import (
...
-	"gonum.org/v1/plot/vg/vgeps"
-	"gonum.org/v1/plot/vg/vgimg"
-	"gonum.org/v1/plot/vg/vgpdf"
-	"gonum.org/v1/plot/vg/vgsvg"
-	"gonum.org/v1/plot/vg/vgtex"
 )
 
+// formats holds the registered canvas image formats
+var formats = struct {
+	sync.Mutex
```
kortschak, line 19: This should probably be a `RWMutex`.

vg/draw/canvas.go
```diff
@@ -256,39 +286,23 @@ func New(c vg.CanvasSizer) Canvas {
...
-
-	default:
-		return nil, fmt.Errorf("unsupported format: %q", format)
+	formats.Lock()
+	defer formats.Unlock()
+	for name, fn := range formats.m {
+		if format == name {
+			return fn(w, h), nil
+		}
 	}
```
kortschak, line 304: Factor this out to a helper that returns the `fn`?

vg/vgimg/vgimg.go
```diff
@@ -20,8 +20,26 @@ import (
...
+	jf := func(w, h vg.Length) vg.CanvasWriterTo {
+		return JpegCanvas{Canvas: New(w, h)}
+	}
+	vgdraw.RegisterFormat("jpg", jf)
+	vgdraw.RegisterFormat("jpeg", jf)
+	tf := func(w, h vg.Length) vg.CanvasWriterTo {
+		return TiffCanvas{Canvas: New(w, h)}
+	}
+	vgdraw.RegisterFormat("tif", tf)
+	vgdraw.RegisterFormat("tiff", tf)
```
kortschak, line 40: Can we split this into meaningful stanzas (one for each `func`).

vg/vgeps/vgeps.go
```diff
@@ -17,8 +17,15 @@ import (
 	"time"
 
 	"gonum.org/v1/plot/vg"
+	"gonum.org/v1/plot/vg/draw"
 )
 
+func init() {
+	draw.RegisterFormat("eps", func(w, h vg.Length) vg.CanvasWriterTo {
+		return New(w, h)
+	})
```
kortschak, line 26: One line? (consistency with vgpdf - or make that 3 lines). Given the simplicity of the values being returned, I think one line is better).

sbinet: in the spirit of keeping the col-count around 80, I changed it to the 3-line form though out.

vg/vgimg/vgimg.go
sbinet: done.

vg/draw/canvas.go
sbinet: it's only needed once.
I'd keep it like as it is (well, I'd err on using the `!=` + `continue` form that keeps the col-count lower and the "indentation arrow of death" at bay, but ok.)

vgall.go
```diff
@@ -0,0 +1,15 @@
+// Copyright ©2017 The Gonum Authors. All rights reserved.
```
sbinet, line 1:
```suggestion
// Copyright ©2021 The Gonum Authors. All rights reserved.
```
? or keep 2017?

kortschak: 2017 was when it was written.

vg/draw/canvas.go
kortschak: OK

vg/vgimg/vgimg.go
kortschak: I was unclear, sorry, but this if fine. I meant jut adding blank lines between specific groups of registrations.

kortschak (approved):

PR #417, opened 2017-11-22
internal/cmpimg: add image diffing output

### LINKED ISSUE #416: cmpimg: add image diff output function for debugging golden image differences
Debugging the failure of #396 was difficult because of possible hardware differences between travis and local machines. To debug the failure I used gimp to get a subtractive image of the golden v output images. We can do this in cmpimg in an easier way; I suggest we add `ImageDiff(a, b image.Image) image.Image` that does a subtraction of `b` from `a` and scales the intensity of difference to maximise the dynamic range if they are different. The resulting image should then be output as a base64 png with the prefix `IMAGE:` to make it possible to extract image failures from travis and use the playground's image rendering feature.

kortschak:
Fixes #416.

See example use [here](https://play.golang.org/p/aiRVekv2z-). The output is the a real vector difference between images, but really just something to highlight where in an image something has gone wrong.

Please take a look.


DavidGamba:
Hi, I wrote this little tool to do the same. The README has some examples on the different approaches to highlight images in case you want to take a look: https://github.com/davidgamba/imgdiff

kortschak:
@DavidGamba I'd be happy to have a more sensible diff if you want to contribute that. The important thing here is that the diff should be intensity scaled to maximise the dynamic range of the signal - in the motivating issue, a simple diff was not good enough.

PR #419, opened 2017-11-24
plotter: fix build

kortschak:
Please take a look.


sbinet:
(I would perhaps give a tad bit more context to the commit message though)

kortschak:
In the context of the git log, it's pretty clear (I don't think anything less than a long paragraph would otherwise explain the issue better).

PR #427, opened 2018-02-22
plotter: Make Image plots and ColorBars handle non-linear axes correctly

### LINKED ISSUE #426: plotter: Image and ColorBar don't adapt to non-linear axes
I would like to plot an Image or Color bar on a plot with a non-linear axis.

### What did you do?

I will submit a pull request with a reproduce

The image should be warped to account for the non-linear axis.

### What actually happened?

The image assume the axis was linear even when it wasn't

### What version of Go and Gonum are you using?

go 1.10, gonum master

yes

ctessum:
Please take a look. Fixes #426.


plotter/image.go
```diff
@@ -71,6 +71,45 @@ func (img *Image) GlyphBoxes(plt *plot.Plot) []plot.GlyphBox {
...
+	if xLinear && yLinear {
+		return img.img
+	}
+	b := img.img.Bounds()
+	o := image.NewNRGBA64(b)
+	for c := 0; c < img.cols; c++ {
+		for r := 0; r < img.rows; r++ {
+			// Find the equivalent image row and column after applying
+			// the axis transforms.
+			cTrans := int(p.X.Norm(img.x(c)) * float64(img.cols))
```
kortschak: The column calculations can be hoisted outside the row loop.

plotter/image.go
```diff
@@ -71,6 +71,45 @@ func (img *Image) GlyphBoxes(plt *plot.Plot) []plot.GlyphBox {
...
+			rTrans := int(p.Y.Norm(img.y(r)) * float64(img.rows))
+			// Find the equivalent row and column of the previous image row
+			// and column after applying the axis transforms.
+			cPrevTrans := int(p.X.Norm(img.x(maxInt(c-1, 0))) * float64(img.cols))
```
kortschak: Here too.

plotter/image.go
```diff
@@ -71,6 +71,45 @@ func (img *Image) GlyphBoxes(plt *plot.Plot) []plot.GlyphBox {
...
+			rPrevTrans := int(p.Y.Norm(img.y(maxInt(r-1, 0))) * float64(img.rows))
+			crColor := img.img.At(c, r)
+			// Set all the pixels in the new image between (cPrevTrans, rPrevTrans)
```
kortschak: Maybe add a TODO to revisit how pixel interpolation is done here.

plotter/image.go
```diff
@@ -71,6 +71,45 @@ func (img *Image) GlyphBoxes(plt *plot.Plot) []plot.GlyphBox {
 	return nil
 }
 
+// transform warps the image to align with non-linear axes.
+func (img *Image) transform(p *plot.Plot) image.Image {
```
kortschak: Can you rename this to `transformFor`? The current name suggests the transformation is being done to `p`.

plotter/image_test.go
```diff
@@ -53,3 +53,46 @@ func ExampleImage() {
...
+// An example of embedding an image in a plot with non-linear axes.
+func ExampleImage_log() {
+	p, err := plot.New()
+	if err != nil {
+		log.Fatalf("error: %v\n", err)
+	}
+	p.Title.Text = "A Logo"
+
+	// load an image
+	f, err := os.Open("testdata/image_plot_input.png")
```
kortschak: Can we use a different input image here. The effect of the transform is not very clear. I would suggest something that is less right-angular.

ctessum:
PTAL

PR #435, opened 2018-04-04
plotter: embed XY interface in XYZ, fix doc comment

anonymouse64:
* Better to embed the interface so changes to `XYer` automatically get moved to `XYZer`
* Fix doc comment to reflect that the `XY` method (for a `XYZs`) is for `XYZer` interface, not `XYer` interface (though it is technically correct as `XY` is embedded from the `XYer`, it's simpler to read this way)

### DIFF
--- a/plotter/plotter.go
+++ b/plotter/plotter.go
@@ -176,14 +176,10 @@ func (ys YValues) Value(i int) float64 {
 
 // XYZer wraps the Len and XYZ methods.
 type XYZer interface {
-	// Len returns the number of x, y, z triples.
-	Len() int
+	XYer
 
 	// XYZ returns an x, y, z triple.
 	XYZ(int) (float64, float64, float64)
-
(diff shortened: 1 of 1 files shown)


sbinet:
I'll wait for others to chime in, but I am not really sold on this code movement.
I'd be willing to instead use this:

```go
var (
	_ XYer = (XYZer)(nil)
)
```

to make sure the 2 interfaces are in line.

anonymouse64:
@sbinet Is there a specific reason you don't like this code movement? To me, it is more readable to see that the `XYZer` interface contains everything from `XYer` using embedded interfaces than see somewhere a compiler check that the interfaces agree. Obviously both things accomplish the same effect at compile-time, so it's just a matter of style I guess.

kortschak:
TBH, I'm no sure why `XYZer` includes `XYer`.

kortschak:
No movement. So closing.

Issue #430, opened 2018-03-14
Make it easier to create a Heat Map with a Legend


btracey:
Is it possible to make a utility function (maybe in plotutil?) for making a simple heat map with a legend? Seems like the easiest way at the moment is to copy and paste the current code in `heat_test.go`, but even that seems to be difficult to get to work right.

kortschak:
Do you want to suggest API for this?

ctessum:
I think a more general-purpose solution (or perhaps a prerequisite to the proposal above) might be to add (or export) the capability to draw a Legend on an arbitrary canvas. Right now it's only allowed to draw it on the same canvas as the plot

ctessum:
It looks like it might be as simple as exporting [Legend.draw](https://github.com/gonum/plot/blob/master/legend.go#L84) and maybe [makeLegend](https://github.com/gonum/plot/blob/master/legend.go#L72)

btracey:
My thought was maybe something with `plotutil` that returns the heat map and the legend, but it sounds like @ctessum has a much clearer idea.

ctessum:
So with #436, it is now possible to create a utility function because we now know how tall and wide a legend is and therefore know how much room to leave for it. I don't have a strong idea of what such a function should look like, though, so I will leave that to someone else.

PR #436, opened 2018-04-10
Add standalone legend capability

ctessum:
Please take a look.

This is a partial fix for #430, in that it allows the width of the legend to be programatically determined rather than through guess-and-check as before.


legend_test.go
```diff
@@ -0,0 +1,102 @@
...
+	outline := c.ClipLinesY(pts)
+	c.StrokeLines(draw.LineStyle{
+		Color: color.Black,
+		Width: vg.Points(1),
+	}, outline...)
+}
+
+// This example creates a some standalone legends with borders around them.
+func ExampleLegend_standalone() {
+
```
kortschak: Delete blank line.

legend_test.go
```diff
@@ -0,0 +1,102 @@
...
+	img := vgimg.New(vg.Points(120), vg.Points(100))
```
sbinet: I would call `img`, a value of type `*vgimg.Canvas`, `c` instead (to keep in line with other examples.)

but it's just nit-picking.

PR #445, opened 2018-05-07
plotter: Fix line clipping bug

ctessum:
Please take a look.

### DIFF
--- a/plotter/polygon.go
+++ b/plotter/polygon.go
@@ -67,6 +67,9 @@ func (pts *Polygon) Plot(c draw.Canvas, plt *plot.Plot) {
 		c.SetColor(pts.Color)
 		var pa vg.Path
 		for _, ring := range ps {
+			if len(ring) == 0 {
+				continue
+			}
 			pa.Move(ring[0])
 			for _, p := range ring {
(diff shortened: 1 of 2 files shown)


ctessum:
Do you mean that you would like me to file an issue?

Without this change, the Polygon plotter panics when it contains rings that are entirely outside of the axis range. In these cases, the clipping function deletes all of points in the ring and then it panics when it tries to move to the first point in the ring.

I changed the golden image so that the test would reproduce the bug. If you run the new test with the old (non-fixed) Polygon code, it will panic.

kortschak:
A description here is good enough, though it might be worth changing the commit message to mention that it prevents a panic ([this](https://groups.google.com/forum/#!msg/golang-dev/6M4dmZWpFaI/SyU5Sl4zZLYJ) is a worthwhile read).

I would prefer if the example test remained the same and you add a `func Test...` that tests this particular issue with a comment explaining what is being done. Otherwise the example looks odd. We should be using examples as examples and ensuring they are tested, but not making all our tests examples. The test for this is just that it doesn't panic, so no image comparison is necessary (although you can if you want).

PR #446, opened 2018-05-14
vg/vgpdf: export Canvas.EmbedFonts

sbinet:


sbinet:
after a good night sleep on this, I am wondering whether we shouldn't unexport the field and instead provide a method to modify it.

then, other canvas backend could also implement that `FontEmbedder` interface: obvious other implementer (I think): the PostScript canvas.

well, I can only think of these 2 canvas (PDF, PS) so perhaps it's not general enough to derive and implement an interface... WDYT?

sbinet:
PTAL: I went with the method instead with exposing the field.

@kortschak let me know if you prefer that way or the previous one.

kortschak:
Either of these is fine - I don't have any strong feelings eitherway - though possibly the method is better.

sbinet:
PTAL @kortschak 
(the test failure is unrelated, but I don't understand why it suddenly appeared... we should perhaps add a cron/scheduled build to catch these)

sbinet:
I've filed #448 for the test failure.

PR #449, opened 2018-05-29
Vgo support

sbinet:
Please take a look.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,12 @@
+module gonum.org/v1/plot
+
+require (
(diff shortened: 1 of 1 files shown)


kortschak:
Can the tests not be put in a script that runs them conditionally on `$(go version)`?

sbinet:
I'd be tempted to remove (for now) the `vgo` test and only enable them when `Go-1.11` hits the door.
having a `go.mod` file now will allow people living with `go-1.9.x`, `go-1.10.x` and `master` to experiment and play with `vgo`.

what do you think?

sbinet:
Updated for the new gonum vgo-support.

Issue #451, opened 2018-05-31
vg/vgpdf: pdfs don't render circle glyphs properly


btracey:
Example code:

```
package main

import (
	"gonum.org/v1/plot"
	"gonum.org/v1/plot/plotter"
	"gonum.org/v1/plot/vg"
)

func main() {
	pts := plotter.XYs{{1, 1}, {2, 2}, {3, 3}, {1.5, 2}, {2.5, 1}}
	scat, err := plotter.NewScatter(pts)
	if err != nil {
		panic(err)
	}
	p, err := plot.New()
	if err != nil {
		panic(err)
	}
	p.Add(scat)
	p.Save(5*vg.Inch, 3*vg.Inch, "plt.pdf")
	p.Save(5*vg.Inch, 3*vg.Inch, "plt.png")
}
```

The png file looks normal

![plt](https://user-images.githubusercontent.com/3680859/40809683-8ce3792a-64e8-11e8-9d00-6cc2dbfad5de.png)

The pdf, however, renders the circles with dashed lines instead of solid ones.
[plt.pdf](https://github.com/gonum/plot/files/2060168/plt.pdf)

I suspect this is related to using `gofpdf`, because this problem surfaced when I pulled to the current master. 

cc @sbinet

kortschak:
They look like they are slightly rotated. At a guess, it's an issue with `vgpdf.Canvas.arc`; this is made more likely by the observation that the quarter segments in the example are clearly not quarter circles.

PR #454, opened 2018-06-09
vg/vgpdf: use Fpdf.Arc to draw circle segments

sbinet:
Fixes gonum/plot#451.

Please take a look.


kortschak:
Can you add a test for this.

Very nice by the way. I like seeing complicated code vanish.

sbinet:
PTAL.

I had noticed some warnings when opening the PDF (with `mupdf`):
```
$> mupdf plt.pdf
warning: lineto with no current point
warning: closepath with no current point
warning: lineto with no current point
warning: closepath with no current point
```
turned out `Fpdf.Arc` didn't advance the stylus so the resulting path was invalid (but rendering was ok.)

I think I got it right now.

vg/vgpdf/vgpdf.go
```diff
@@ -261,64 +260,21 @@ func (c *Canvas) pdfPath(path vg.Path, style string) {
...
-	x3r := x3*cosar - y3*sinar + x
-	y3r := x3*sinar + y3*cosar + y
-	x4 = r*vg.Length(math.Cos(a2)) + x
-	y4 = r*vg.Length(math.Sin(a2)) + y
-	c.doc.Curve(c.unit(x2r), c.unit(y2r), c.unit(x3r), c.unit(y3r), c.unit(x4), c.unit(y4), style)
+	r := c.unit(comp.Radius)
+	const deg = 180 / math.Pi
+	angle := comp.Angle * deg
+	beg := comp.Start * deg
+	end := deg * (beg + angle)
```
kortschak: Why is this second `deg` factor here?

sbinet: nice catch.

sbinet: this has sent me on an interesting adventure, testing when `beg != end`.
it seems we don't have a very good test suite of various shapes and their rendering across `vg`-backends...

kortschak: You might like to try using github.com/biogo/graphics/rings as a possible test suite. I don't generate PDFs, but it does otherwise exercise a lot of different features.

TBH, I have been considering forking that package for general purpose graph rendering, but it's a lot of work.

kortschak:
Is this waiting for anything?

sbinet:
oops.
let me rebase and squash.

PR #460, opened 2018-07-19
plotter: fix histogram hang for zero-width data range

### LINKED ISSUE #459: Infinite loop drawing histogram
The following code seems to loop infinitely on `p.Save`:

```go
p, err := plot.New()
if err != nil {
	panic(err)
}

hist, err := plotter.NewHist(plotter.Values([]float64{1.0}), 60)
hist.Normalize(1)

p.Add(hist)
err = p.Save(4*vg.Inch, 4*vg.Inch, "hist.png")
if err != nil {
	panic(err)
}
```

The problem is in the function below:
```go
// Normalize normalizes the histogram so that the
// total area beneath it sums to a given value.
func (h *Histogram) Normalize(sum float64) {
	mass := 0.0
	for _, b := range h.Bins {
		mass += b.Weight
	}
	for i := range h.Bins {
		h.Bins[i].Weight *= sum / (h.Width * mass)
	}
}
```

Since `h.Width` is zero, an infinity appears that breaks the render to the canvas

It's not totally obvious what should be done here, perhaps set the bin width to 1.0 for trivial histograms?

kortschak:
Please take a look.

Fixes #459.

### DIFF
--- a/plotter/histogram_test.go
+++ b/plotter/histogram_test.go
@@ -10,6 +10,7 @@ import (
 	"math"
 	"math/rand"
 	"testing"
+	"time"
 
 	"gonum.org/v1/plot"
 	"gonum.org/v1/plot/internal/cmpimg"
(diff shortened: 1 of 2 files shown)


plotter/histogram_test.go
```diff
@@ -62,3 +63,33 @@ func ExampleHistogram() {
...
+func TestSingletonHistogram(t *testing.T) {
+	done := make(chan struct{}, 1)
+	go func() {
+		defer close(done)
+		p, err := plot.New()
+		if err != nil {
+			t.Errorf("unexpected error from plot.New: %v", err)
+		}
+
+		hist, err := NewHist(Values([]float64{1.0}), 60)
```
rosshemsley: I suspect the same problem would be triggered by many equal values, consider adding a test to make sure the fix solves that case too?

plotter/histogram_test.go
```diff
@@ -62,3 +63,33 @@ func ExampleHistogram() {
...
 	cmpimg.CheckPlot(ExampleHistogram, t, "histogram.png")
 }
+
+func TestSingletonHistogram(t *testing.T) {
+	done := make(chan struct{}, 1)
+	go func() {
+		defer close(done)
+		p, err := plot.New()
+		if err != nil {
+			t.Errorf("unexpected error from plot.New: %v", err)
```
rosshemsley: nit: consider `t.Fatalf()` if the rest of the test could panic if the condition fails.

plotter/histogram_test.go
```diff
@@ -62,3 +63,33 @@ func ExampleHistogram() {
...
+		}
+
+		hist, err := NewHist(Values([]float64{1.0}), 60)
+		if err != nil {
+			t.Errorf("unexpected error from NewHist: %v", err)
```
rosshemsley: nit: consider t.Fatalf()

kortschak:
I have reconsidered the erroring - PTAL.

rosshemsley:
LGTM 👍I pulled this change into my own project and I it's working fine.

Thanks!

PR #462, opened 2018-07-23
use gonum.org/v1/tools + golang.org/x/exp/rand

### LINKED ISSUE #461: meta: use gonum.org/v1/tools to detect invalid imports
we should leverage `gonum.org/v1/tools` and `gonum.org/v1/tools/cmd/check-imports` to detect invalid imports.

this would have helped with https://github.com/gonum/plot/pull/457.

sbinet:
- add `goimports` check
- add `gofmt -s -l` check
- add `check-imports`

Fixes #461


sbinet:
some golden files had to be regenerated b/c of the migration to `golang.org/x/exp/rand`.
some of them look less pleasing to the eye.

palette/moreland/luminance_test.go
```diff
@@ -7,9 +7,9 @@ package moreland
 import (
 	"fmt"
 	"image/color"
-	"math/rand"
 	"testing"
 
+	"golang.org/x/exp/rand"
```
kortschak: Please put golang.org/x/exp/rand in its own stanza throughout.

sbinet:
`plotter/testdata/verticalBoxPlot_golden.png` looks slightly less nice because of the units of the Y axis (that used to be displayed with only 1 digit and now are displayed with 2 digits after the dot), this makes the X axis labels slightly scrambled...

sbinet:
ok. fixed. PTAL.

PR #465, opened 2018-08-07
plotter: Ensure Sankey can be drawn more than once

ctessum:
Previously, rendering the same diagram multiple times produced differing results.

Please take a look.

### DIFF
--- a/plotter/sankey.go
+++ b/plotter/sankey.go
@@ -100,11 +100,6 @@ type stock struct {
 
 	// max is min plus the larger of receptorValue and sourceValue.
 	max float64
-
-	// sourceFlowPlaceholder and receptorFlowPlaceholder track
-	// the current plotting location during
-	// the plotting process.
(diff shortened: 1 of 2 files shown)


plotter/sankey.go
```diff
@@ -209,18 +204,24 @@ func NewSankey(flows ...Flow) (*Sankey, error) {
 func (s *Sankey) Plot(c draw.Canvas, plt *plot.Plot) {
 	trCat, trVal := plt.Transforms(&c)
 
+	// sourceFlowPlaceholder and receptorFlowPlaceholder track
+	// the current plotting location during
+	// the plotting process.
+	sourceFlowPlaceholder := make(map[*stock]float64)
```
sbinet: might want to also give a hint for the map size:
```go
sourceFlowPlaceholder := make(map[*stock]float64, len(s.flows))
receptorFlowPlaceholder := make(map[*stock]float64, len(s.flows))
```

PR #478, opened 2018-10-29
plotutil: Add *plotter.Function to AddLines

jonas-schulze:
Please take a look.


plotutil/add.go
```diff
@@ -170,10 +170,11 @@ func AddScatters(plt *plot.Plot, vs ...interface{}) error {
 
 // AddLines adds Line plotters to a plot.
 // The variadic arguments must be either strings
-// or plotter.XYers.  Each plotter.XYer is added to
+// or plotter.XYers or *plotter.Function ("plot types" for short).
```
kortschak:
```
// Each variadic argument must be a string or
// one of a plotting type, plotter.XYers or *plotter.Function.
```

plotutil/add.go
```diff
@@ -170,10 +170,11 @@ func AddScatters(plt *plot.Plot, vs ...interface{}) error {
...
+// Each plot type is added to
 // the plot using the next color and dashes
 // shape via the Color and Dashes functions.
-// If a plotter.XYer is immediately preceeded by
+// If a plot type is immediately preceeded by
```
kortschak:
```
// If a plotting type is ...
```

jonas-schulze:
One could also use `func(float64) float64` instead of `plotter.Function`. What do you think about that?

jonas-schulze:
@kortschak is there a procedure on how and where to discuss this? Do you have any updates on this? I'd be happy to help.

kortschak:
A thread at gonum-dev would probably be worthwhile, but otherwise waiting for @sbinet and @ctessum to chime in here would be OK.

ctessum:
I haven't used the plotutil package, but after looking at it it seems fine to me. An alternative could be to have a separate `AddFunction` function, but I'm not sure that would be better than the current proposal

plotutil/add.go
```diff
@@ -203,8 +204,18 @@ func AddLines(plt *plot.Plot, vs ...interface{}) error {
...
+			i++
+			ps = append(ps, t)
+			if name != "" {
+				items = append(items, item{name: name, value: t})
+				name = ""
+			}
+
 		default:
-			panic(fmt.Sprintf("AddLines handles strings and plotter.XYers, got %T", t))
+			panic(fmt.Sprintf("AddLines handles strings, plotter.XYers and *plotter.Function, got %T", t))
```
sbinet: we usually try to have a panic string of the form `"[package name]: [error message]"`
could you change this along these lines?

PR #479, opened 2018-10-31
plot: allow to use functions as Tick.Markers

egurnov:
Allows to create a Ticker using a function with a matching signature.

Please take a look.


kortschak:
Can you please also send a PR adding yourself to A+C in gonum/gonum with the commit message "A+C: add Alexander Egurnov".

PR #480, opened 2018-10-31
fix imports

jedi108:
Please take a look.


kortschak:
We do not import github.com/gonum packages.

PR #483, opened 2018-11-23
Step line plotter implemented

igrmk:
This pull request implements step line plotter.
It supports filled step plot as well.


sbinet:
Just a quick note that this didn't drop off my radar.
I'll try to review this PR shortly. Thanks for your contribution and interest!

plotter/step.go
```diff
@@ -0,0 +1,173 @@
...
+	StepTypePost
+)
+
+// Step implements the Plotter interface, drawing a stepped line.
+type Step struct {
+	// XYs is a copy of the points for this line.
+	XYs
+
+	// StepStyle is the type of step line
+	StepType StepType
```
sbinet: `s/StepType StepType/StepStyle StepKind/`

plotter/step.go
```diff
@@ -0,0 +1,173 @@
...
+	// StepTypePost means that two consecutive points are connected by two lines in the following order: horizontal, vertical.
+	StepTypePost
+)
+
+// Step implements the Plotter interface, drawing a stepped line.
+type Step struct {
+	// XYs is a copy of the points for this line.
+	XYs
+
+	// StepStyle is the type of step line
```
sbinet: `s/type/kind/`

plotter/step.go
```diff
@@ -0,0 +1,173 @@
...
+	StepType StepType
+
+	// LineStyle is the style of the line connecting the points.
+	LineStyle *draw.LineStyle
+
+	// ShadeColor is the color of the shaded area.
```
sbinet: we should probably document that when left alone it wouldn't fill anything.

plotter/step.go
```diff
@@ -0,0 +1,173 @@
...
+	if pts.ShadeColor != nil {
+		xmin, xmax, ymin, ymax = XYRange(pts)
+		ymin = math.Min(ymin, 0.)
+		ymax = math.Max(ymax, 0.)
+		return
+	}
+	return XYRange(pts)
+}
+
+// Thumbnail the thumbnail for the Step,
```
sbinet: `s/Thumbnail/Thumbnail returns/`

you could also get rid of the extra line feed and put the whole documentation on a single line.

plotter/step.go
```diff
@@ -0,0 +1,173 @@
...
+	if pts.ShadeColor != nil {
+		points := []vg.Point{
+			{X: c.Min.X, Y: c.Min.Y},
+			{X: c.Min.X, Y: c.Max.Y},
+			{X: c.Max.X, Y: c.Max.Y},
+			{X: c.Max.X, Y: c.Min.Y},
+		}
+		poly := c.ClipPolygonY(points)
+		c.FillPolygon(pts.ShadeColor, poly)
+	} else if pts.LineStyle != nil {
```
sbinet: I wouldn't make these exclusive. what if somebody creates 2 step plots with the same fill color but different line styles?

plotter/step.go
```diff
@@ -0,0 +1,173 @@
...
+		y := c.Center().Y
+		c.StrokeLine2(*pts.LineStyle, c.Min.X, y, c.Max.X, y)
+	}
+}
+
+// NewStepPoints returns both a Step and a
+// Points for the given point data.
+func NewStepPoints(xys XYer) (*Step, *Scatter, error) {
```
sbinet: what about `s/NewStepPoints/NewStepScatter/` instead?
that's what this function returns...

also, could you add a test or an example for this function?

plotter/step_test.go
```diff
@@ -0,0 +1,84 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
```
sbinet: `s/2015/2018/`

plotter/step.go
```diff
@@ -0,0 +1,173 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
```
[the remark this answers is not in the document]
igrmk: I have not considered transformation function because I think that step line is a quite common thing. It would be nice if I can use it out-of-box. Also it differs little bit in filling algorithm. We should not try to fill lonely vertical line. You probably would not nice difference anyway but it is just cleaner approach.

I have not considered `Line` enhancement because `Step` has additional data (`StepType` or better `StepKind`) that is useless in `Line`. However it can make a sense to make it a pointer or to add `NoStep` default constant and merge it with `Line`. We can reuse some code this way. It looks like a good idea for me. Let me reimplement it this way?

Also setting `ShadeColor` of 'Line' doesn't make its y-axis include 0 (zero) by default. I'm not sure if it is a bug or intentional. But it doesn't make a sense for `Step` because most bottom step wouldn't have a shade and it looks ugly. However this is the same for `Line` if bottom line is horizontal. So I think it is a bug in `Line`. And then there is no obstacles for merging them. This is fixed in `Step` already.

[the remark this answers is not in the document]
igrmk: And surely I'll update a date.

plotter/step.go
```diff
@@ -0,0 +1,173 @@
...
+	XYs
+
+	// StepStyle is the type of step line
+	StepType StepType
+
+	// LineStyle is the style of the line connecting the points.
+	LineStyle *draw.LineStyle
+
+	// ShadeColor is the color of the shaded area.
+	ShadeColor color.Color
```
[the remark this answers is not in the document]
igrmk: It is called `ShadeColor` in `Line` and `FillColor` in `Histogram` so I would like you to check it up and clarify.

plotter/step_test.go
```diff
@@ -0,0 +1,84 @@
...
+	stepPost.StepType = StepTypePost
+	stepPost.LineStyle = nil
+	stepPost.ShadeColor = color.RGBA{B: 255, A: 40}
+
+	p.Add(stepPre, stepMid, stepPost)
+	p.Legend.Add("pre", stepPre)
+	p.Legend.Add("mid", stepMid)
+	p.Legend.Add("post", stepPost)
+
+	err = p.Save(200, 200, "testdata/step.pdf")
```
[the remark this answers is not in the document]
igrmk: OK. But I've encountered some appearance bugs with filling with semitransparent colors. Anyway currently I can fill with solid ones.

plotter/step.go
```diff
@@ -0,0 +1,173 @@
...
+	"image/color"
+	"math"
+
+	"gonum.org/v1/plot"
+	"gonum.org/v1/plot/vg"
+	"gonum.org/v1/plot/vg/draw"
+)
+
+// StepType specifies a form of a connection of two consecutive points.
+type StepType int
```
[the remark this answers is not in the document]
igrmk: Good idea.

plotter/step.go
```diff
@@ -0,0 +1,173 @@
...
+
+const (
+	// StepTypePre means that two consecutive points are connected by two lines in the following order: vertical, horizontal.
+	StepTypePre StepType = iota
```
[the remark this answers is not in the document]
igrmk: Good idea.

plotter/step.go
igrmk: Sure!

plotter/step.go
igrmk: Sure!

plotter/step.go
igrmk: Sure!

plotter/step.go
igrmk: This is interesting. We can end up with a very dark line on a dark background for example. Maybe fill only bottom half if a line style is specified? Thus it should look exactly as on a plot. By the way it can be a good enhancement for `Line` as well.

plotter/step.go
igrmk: There is `NewLinePoints` in `Line` so here consistency is the question.

igrmk: I can use it in test because we decided to use PNG. I've encountered non-deterministic behavior of scatter plot in PDF, so cannot compare PDFs. It should be OK for PNG.

plotter/step_test.go
igrmk: Sure!

plotter/step.go
igrmk: Sure!

plotter/step.go
igrmk: Especially if we will merge them.

plotter/step.go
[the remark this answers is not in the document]
kortschak: This should be `FillColor`. I think the `Line` field should be changed. Shade is not clearly talking about the fill area.

plotter/step.go
kortschak: TBH, I don't understand why `NewLinePoints` exists.

plotter/step.go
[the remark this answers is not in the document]
igrmk: I will rename it.

plotter/step.go
igrmk: I agree. I'll better remove it. The only reason I see is we copy only once instead of two times.

plotter/step.go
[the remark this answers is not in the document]
igrmk: Done

plotter/step_test.go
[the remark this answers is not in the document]
igrmk: Done

plotter/step.go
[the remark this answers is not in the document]
igrmk: Done

plotter/step.go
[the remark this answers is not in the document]
igrmk: Done

plotter/step.go
igrmk: Done

plotter/step.go
igrmk: Done

plotter/step.go
igrmk: Done

plotter/step.go
igrmk: Done

plotter/step.go
igrmk: Done

plotter/step.go
igrmk: Done

plotter/step_test.go
igrmk: Done

plotter/step.go
```diff
@@ -0,0 +1,168 @@
+// Copyright ©2018 The gonum Authors. All rights reserved.
```
kortschak: s/gonum/Gonum/

plotter/step_test.go
```diff
@@ -0,0 +1,85 @@
+// Copyright ©2018 The gonum Authors. All rights reserved.
```
kortschak: s/gonum/Gonum/

plotter/step_test.go
```diff
@@ -0,0 +1,85 @@
...
+	p, err := plot.New()
+	if err != nil {
+		log.Panic(err)
+	}
+	p.Title.Text = "Step Example"
+	p.X.Label.Text = "X"
+	p.Y.Label.Text = "Y"
+	p.Add(NewGrid())
+
+	stepPre, err := NewStep(randomPoints(n, 0.))
```
kortschak: s/0./0/; it is an ideal constant in a float64 context.

plotter/step_test.go
```diff
@@ -0,0 +1,85 @@
...
+	stepPre.FillColor = color.RGBA{A: 40}
+
+	stepMid, err := NewStep(randomPoints(n, 3.5))
+	if err != nil {
+		log.Panic(err)
+	}
+	stepMid.StepStyle = MidStep
+	stepMid.LineStyle = &draw.LineStyle{Color: color.RGBA{R: 196, B: 128, A: 255}, Width: vg.Points(1)}
+
+	stepPost, err := NewStep(randomPoints(n, 7.))
```
kortschak: Same here.

plotter/step.go
igrmk: Shouldn't it be changed in some bulk way for the entire repository? Currently non-capitalized version is used in every other file.

plotter/step_test.go
igrmk: Shouldn't it be changed in some bulk way for the entire repository? Currently non-capitalized version is used in every other file.

plotter/step_test.go
igrmk: Done.

plotter/step_test.go
igrmk: Done.

plotter/step.go
kortschak: Yes, after I saw this here I sent #485.

igrmk: OK

plotter/step_test.go
igrmk: OK

plotter/step.go
```diff
@@ -0,0 +1,168 @@
...
+	"gonum.org/v1/plot"
+	"gonum.org/v1/plot/vg"
+	"gonum.org/v1/plot/vg/draw"
+)
+
+// StepKind specifies a form of a connection of two consecutive points.
+type StepKind int
+
+const (
+	// PreStep means that two consecutive points are connected by two lines in the following order: vertical, horizontal.
```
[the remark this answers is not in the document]
igrmk: Done. Before approval... Do you think it is worth to merge `Step` with `Line`? I wrote about this in the top comment. Basically we can just add `NoStep` default to `Line`. Seems logical to me. Line stays line but we can control its shape in this case.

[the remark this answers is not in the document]
kortschak: I'll leave that up to @sbinet.

sbinet:
apologies for the belated answer. I am running a Go tutorial this week and thus a bit behind with my open source duties. (I'll circle back to this PR tomorrow evening.)

plotter/step.go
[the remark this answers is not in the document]
sbinet: I think it would be interesting to see what it would look like.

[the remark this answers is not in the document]
sbinet: maybe @ctessum has something to say about this as well?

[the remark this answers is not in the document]
igrmk: @sbinet I can add another commit to this pull request to show what it would look like. If we decide that separate `Step` is better then I'll just revert it. I'm busy right now, so I'll do it in 1 or 2 days.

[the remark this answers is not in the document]
sbinet: SGTM. Thanks

sbinet:
ping?
(could you also rebase?)

igrmk:
Hello all, @sbinet 
! Sorry for so long absence. I was busy with my work.
I merged `Step` and `Line`. And I'm quite happy with result.
I need to highlight some things:
* Thumbnail is changed. Originally if both `LineStyle` and `FillColor` were specified then thumbnail would be a filled box with a line in the middle. Now it is the line in the middle filled below.
* `ShadeColor` is renamed to `FillColor` and it is just an interface now not a pointer to interface because pointer to interface doesn't make sense.
* Originally a filled line was not clipped at all. It is fixed and tests are added.
* Now zero width means that lines are disabled. Originally we tried to draw zero width lines. It is specified in documentation now. So you can leave only filling without lines.

So `FillColor` is breaking change (renamed and an interface instead of a pointer). I think it should be major release then if you are OK with this change.

Everything is rebased to latest upstream. Thanks.

plotter/filledLine_test.go
```diff
@@ -0,0 +1,64 @@
...
+	"log"
+	"testing"
+
+	"golang.org/x/exp/rand"
+
+	"gonum.org/v1/plot"
+	"gonum.org/v1/plot/internal/cmpimg"
+)
+
+func filledLine() {
```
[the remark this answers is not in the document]
igrmk: Done

plotter/filledLine_test.go
```diff
@@ -0,0 +1,64 @@
...
+				pts[i].X = x + rnd.Float64()
+			} else {
+				pts[i].X = pts[i-1].X + 0.5 + rnd.Float64()
+			}
+			pts[i].Y = -5. + 10*rnd.Float64()
+		}
+		return pts
+	}
+
+	n := 4
```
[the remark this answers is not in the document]
igrmk: Done

plotter/filledLine_test.go
```diff
@@ -0,0 +1,64 @@
...
+	p.Add(NewGrid())
+
+	filled, err := NewLine(randomPoints(n, 0))
+	if err != nil {
+		log.Panic(err)
+	}
+	filled.FillColor = color.RGBA{R: 196, G: 255, B: 196, A: 255}
+
+	p.Add(filled)
+	p.Y.Max = -1
```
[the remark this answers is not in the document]
igrmk: I added a comment to code.

plotter/step_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+	"golang.org/x/exp/rand"
+
+	"gonum.org/v1/plot"
+	"gonum.org/v1/plot/internal/cmpimg"
+	"gonum.org/v1/plot/vg"
+	"gonum.org/v1/plot/vg/draw"
+)
+
+func stepLine() {
```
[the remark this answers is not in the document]
igrmk: Done

plotter/filledLine_test.go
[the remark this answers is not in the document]
sbinet: ah, I see... I guess I fixed the X-axis part of that bug here: gonum/plot#489.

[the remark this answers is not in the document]
igrmk: Hmm.. I think the fix for https://github.com/gonum/plot/pull/489 was not quite correct. It works only for X axis clipping. It is better to trait filled part as a polygon. This way it should work for every case.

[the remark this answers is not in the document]
sbinet: yep. as I said in #489, your fix is also working *and* is addressing the Y-range as well :)

igrmk:
Oops, pushed to my own branch, sorry :) Now it is here.

sbinet:
@kortschak WDYT?

igrmk:
Tests are extracted from `plotter` package to `plotter_test` one. Very good issue. Now you can copy an example and it should just work.

plotter/filledLine_test.go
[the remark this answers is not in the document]
kortschak: I don't think this is the right place for this. If there is a bug that was fixed, it should be tested, but a code example is something that people will be referring to for use. Can you make a stand alone (non-example) test for this and refer to the issue that it's testing explicitly?

plotter/filledLine_test.go
```diff
@@ -16,7 +16,7 @@ import (
 	"gonum.org/v1/plot/plotter"
 )
 
-func ExampleLine_filledLine() {
+func clippedFilledLine() {
```
kortschak: Add comment linking to #488. `// See https://github.com/gonum/plot/issues/488`.

kortschak:
Also, can you send a PR to gonum/gonum adding yourself to AUTHORS and CONTRIBUTORS with the commit message "A+C: add Igor Mikushkin".

plotter/filledLine_test.go
igrmk: Done

igrmk:
Sure, will do, thanks.

kortschak:
Thanks

PR #484, opened 2018-11-25
Fixes for PDF comparison

igrmk:
Current PDF comparison doesn't compare actual images.
You can easily check it by changing color and run tests.

This PR fixes PDF comparison.

P.S. Also I've found that scatter plot produces non-deterministic PDF's so you cannot compare them. We have no such PDFs now so it's not a big issue.


sbinet:
ping?

PR #476, opened 2018-10-16
vg/vgimg: change background color to transparent

MiguelPires:
This PR changes vgmimg.go to draw graphs in a transparent background rather than white.

PR #486, opened 2018-12-05
vgimg: Add option to specify image background color.

ctessum:
Previously, the background color was always white, with no 
straightforward way to specify a transparent backgroun.

This is an alternative to #476 .

Please take a look.

### DIFF
--- a/vg/vgimg/vgimg.go
+++ b/vg/vgimg/vgimg.go
@@ -39,6 +39,10 @@ type Canvas struct {
 
 	// width is the current line width.
 	width vg.Length
+
+	// backgroundColor is the background color, set by
+	// UseBackgroundColor.
+	backgroundColor color.Color
 }
 
 const (
@@ -54,7 +58,7 @@ const (
 
 // New returns a new image canvas.
 func New(w, h vg.Length) *Canvas {
-	return NewWith(UseWH(w, h))
+	return NewWith(UseWH(w, h), UseBackgroundColor(color.White))
 }
 
 // NewWith returns a new image canvas created according to the specified
@@ -67,6 +71,7 @@ func New(w, h vg.Length) *Canvas {
 // passed).
 func NewWith(o ...option) *Canvas {
 	c := new(Canvas)
+	c.backgroundColor = color.White
 	var g uint32
 	for _, opt := range o {
 		f := opt(c)
@@ -101,7 +106,7 @@ func NewWith(o ...option) *Canvas {
 		c.gc.Scale(1, -1)
 		c.gc.Translate(0, -h)
 	}
-	draw.Draw(c.img, c.img.Bounds(), image.White, image.ZP, draw.Src)
+	draw.Draw(c.img, c.img.Bounds(), &image.Uniform{c.backgroundColor}, image.ZP, draw.Src)
 	c.color = []color.Color{color.Black}
 	vg.Initialize(c)
 	return c
@@ -111,8 +116,9 @@ func NewWith(o ...option) *Canvas {
 // used when initializing a canvas are compatible with
 // each other.
 const (
-	setsDPI = 1 << iota
+	setsDPI uint32 = 1 << iota
 	setsSize
+	setsBackground
 )
 
 type option func(*Canvas) uint32
@@ -148,7 +154,7 @@ func UseDPI(dpi int) option {
 func UseImage(img draw.Image) option {
 	return func(c *Canvas) uint32 {
 		c.img = img
-		return setsSize
+		return setsSize | setsBackground
 	}
 }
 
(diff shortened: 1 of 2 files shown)


vg/vgimg/vgimg.go
```diff
@@ -111,8 +117,9 @@ func NewWith(o ...option) *Canvas {
 // used when initializing a canvas are compatible with
 // each other.
 const (
-	setsDPI = 1 << iota
-	setsSize
+	setsDPI        uint32 = 1
```
kortschak: `setsDPI uint32 = 1 << iota` and remove the assignments below.

vg/vgimg/vgimg.go
```diff
@@ -161,7 +168,20 @@ func UseImageWithContext(img draw.Image, gc draw2d.GraphicContext) option {
...
+	}
+}
+
+// UseBackgroundColor specifies the image background color.
+// The default color is white.
+func UseBackgroundColor(c color.Color) option {
+	return func(canvas *Canvas) uint32 {
+		// Add this operation as a initHook so that it happens after
+		// canvas.img is initialized.
+		canvas.initHooks = append(canvas.initHooks, func(canvas *Canvas) {
```
kortschak: This is subtle. It think it needs clearer commentary explaining that this operation happens on the first loop over the opts to queued for the second.

vg/vgimg/vgimg.go
```diff
@@ -39,6 +39,10 @@ type Canvas struct {
 
 	// width is the current line width.
 	width vg.Length
+
+	// initHooks is a list of operations to be performed at the
+	// end of initialization. options can add to this list.
```
kortschak: s/options/options passed into NewWith/

vg/vgimg/vgimg.go
```diff
@@ -101,8 +105,10 @@ func NewWith(o ...option) *Canvas {
 		c.gc.Scale(1, -1)
 		c.gc.Translate(0, -h)
 	}
-	draw.Draw(c.img, c.img.Bounds(), image.White, image.ZP, draw.Src)
 	c.color = []color.Color{color.Black}
+	for _, o := range c.initHooks {
```
kortschak: For the sake of consistency s/o/opt/.

vg/vgimg/vgimg.go
ctessum: But then, would `setsBackground` == 3? I believe I tested and confirmed that if it was 3 rather than 4 then things wouldn't work correctly, but I may be wrong.

ctessum:
@kortschak I'm not sure what you have in mind regarding setting a background field. If we add a field to `vgimg.Canvas` we would have to add some logic to make sure that the background was applied before anything else happened. It seems like that would need to include something like a `sync.Once` field in `vgimg.Canvas`, and a `once.Do()` added to every method, which doesn't seem very elegant. Let me know if I'm misunderstanding.

vg/vgimg/vgimg.go
kortschak: https://play.golang.org/p/2qcq_z_KJ0S

kortschak:
The approach used in #476 without the extra rigmarole seemed OK. You set the background colour in the first loop in the preamble and then use that colour when you create backgrounds.

vg/vgimg/vgimg.go
```diff
@@ -161,7 +166,16 @@ func UseImageWithContext(img draw.Image, gc draw2d.GraphicContext) option {
 		c.img = img
 		c.gc = gc
 		c.dpi = gc.GetDPI()
-		return setsDPI | setsSize
+		return setsDPI | setsSize | setsBackground
+	}
+}
+
+// UseBackgroundColor specifies the image background color.
+// The default color is white.
```
kortschak: `// Without UseBackgroundColor the default color is white.` just to be clear since the context here is a little smeared.

ctessum: fixed

vg/vgimg/vgimg.go
```diff
@@ -161,7 +166,16 @@ func UseImageWithContext(img draw.Image, gc draw2d.GraphicContext) option {
 		c.img = img
 		c.gc = gc
 		c.dpi = gc.GetDPI()
-		return setsDPI | setsSize
+		return setsDPI | setsSize | setsBackground
+	}
+}
+
+// UseBackgroundColor specifies the image background color.
+// Without UseBackgroundColor, the default color is white.
```
kortschak: I've just realised that this is not necessarily true. If a user calls `NewWith()` without `UseBackgroundColor`, then the background color is nil. For this to be true, `c.backgroundColor` should be set to ` image.White` before the options loop.

After that, LGTM.

ctessum: done

Issue #316, opened 2016-10-06
How to LogScale in reverse?


kellabyte:
I have duration in seconds on the Y scale and percentages on the X scale but when I set LogScale I get the top and right parts of the graph squished. How can I log scale in reverse so the later values are higher resolution instead?

kortschak:
That's not achievable in the general sense, at least not without further configuration. To do this you need to have a known upper bound, otherwise you are attempting to increase resolution as to also increase the size of the function range. Are you sure you really want to do this? Can you explain your end goal?

If you do have an upper bound, the trivial solution is to plot the difference from that upper bound, thus inverting the sign of your function. An alternative is to implement your own `Normalizer` and `Ticker` types that know your upper bound. To get that to work well, you will probably need to copy the `plot.precisionOf` and `plot.formatFloatTick` functions (maybe we should export them to aid this kind of endeavour.

kellabyte:
The log scale it does currently looks like this on the X axis and compresses the tail.
```
-----------------------------------------------
0                    30 |        50 |   70 | 90          
```

But when measuring latency, showing the 90-100 percentiles in greater detail is more important than 0-30 percentiles so I would rather something that compresses the head like this:

```
-----------------------------------------------
0 | 30 | 50    | 70        | 90                         
```

kortschak:
Yeah, that makes sense. Either of the approaches I mention above would work for you.

kellabyte:
> If you do have an upper bound, the trivial solution is to plot the difference from that upper bound, thus inverting the sign of your function.

Could you explain or give an example? I'm not sure I understand.

kortschak:
The trivial approach is to convert the value you are plotting from a latency percentile one hundred minus the latency percentile (I can't think of a good name for this - which is probably a long-term argument against it). The alternative approach is to fork the code [here](https://github.com/gonum/plot/blob/master/axis.go#L431) and [here](https://github.com/gonum/plot/blob/master/axis.go#L171) to implement a `Ticker` and a `Normalizer` that does what you want. Actually, thinking more, you don't need the `Ticker`, since you can just use a `ConstantTicks` - then it's just the `Normalizer`, which is trivial. Something like this:

```
// ReverseLogScale can be used as the value of an Axis.Scale function to
// set the axis to a log scale.
type ReverseLogScale struct{}

// Normalize returns the fractional logarithmic distance of
// x between min and max.
func (ReverseLogScale) Normalize(min, max, x float64) float64 {
[...]

PR #487, opened 2018-12-05
gonum/plot: add support for inverting an axis

npotts:
This change modifies gonum/plot to add direct library support to invert an axis

This change was spurned by attempting to plot atmospheric data, which is
commonly plotted against an inverted pressure axis such as this [example](https://www.ncl.ucar.edu/Applications/Images/h_lat_1_lg.png).   At the suggestion of Sebastion Binet I am submitting this patch. This is somewhat related to #316.

Signed-off-by: npotts <npotts@users.noreply.github.com>

Please take a look.


plotter/invertedaxis_example_test.go
```diff
@@ -0,0 +1,51 @@
...
+	p.Add(f, plotter.NewGrid())
+	p.Legend.Add("exp(x)", f)
+
+	//Notice that both .Min and .Max for X and Y are both in 'normal' order
+	p.X.Min = f.XMin
+	p.X.Max = f.XMax
+	p.Y.Min = math.Exp(f.XMin)
+	p.Y.Max = math.Exp(f.XMax)
+
+	err = p.Save(10*vg.Centimeter, 10*vg.Centimeter, "testdata/invertedlogscale.png")
```
sbinet: could you add a test (so we make sure everything is alright) ?
Examples without a:
```go
// Output:
// blabla
```
aren't run as part of `go test`...

something like that perhaps:
```go
func TestInvertedScale(t *testing.T) {
	cmpimg.CheckPlot(Example_invertedScale, t, "invertedscale.png")
}
```

you'll probably have to add the png file :)

plotter/invertedaxis_example_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+// Example_invertedScale shows how to create a plot with an inverted Y-axis.
+// This is nearly identical to the Log example, except it inverts the Y axis.
+func Example_invertedScale() {
+	p, err := plot.New()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	p.Title.Text = "Exampled of an Inverted Axis"
```
sbinet: `s/Exampled/Example/`
`s/an Inverted Axis/inverted axes/` (as both x and y are inverted)

plotter/invertedaxis_example_test.go
```diff
@@ -0,0 +1,57 @@
...
+	f := plotter.NewFunction(math.Exp)
+	f.XMin = 0.2
+	f.XMax = 10
+
+	f.Color = color.RGBA{R: 255, A: 255}
+
+	p.Add(f, plotter.NewGrid())
+	p.Legend.Add("exp(x)", f)
+
+	//Notice that both .Min and .Max for X and Y are both in 'normal' order
```
sbinet: `s|//Notice|// Notice|`

plotter/invertedaxis_example_test.go
npotts: Ok I think I did it right.  Let me know otherwise

axis.go
```diff
@@ -171,6 +171,18 @@ func (LogScale) Normalize(min, max, x float64) float64 {
 	return (log(x) - logMin) / (log(max) - logMin)
 }
 
+// InvertedScale can be used as the value of an Axis.Scale function to
+// invert the axis using any Normalizer. This can be useful for plots that
```
ctessum: The second sentence here should be deleted, I think. It's not adding information

plotter/invertedaxis_example_test.go
```diff
@@ -0,0 +1,57 @@
...
+	"testing"
+
+	"gonum.org/v1/plot"
+	"gonum.org/v1/plot/internal/cmpimg"
+	"gonum.org/v1/plot/plotter"
+	"gonum.org/v1/plot/vg"
+)
+
+// Example_invertedScale shows how to create a plot with an inverted Y-axis.
+// This is nearly identical to the Log example, except it inverts the Y axis.
```
ctessum: I think the second sentence should be deleted, and the first sentence changed to say that both axes are inverted.

npotts:
Ill change those ASAP.

Are you all open to the addition of some limited number of "helpers" like this?  I have some basic helper functions that aid in labeling, enforcing a minimal number of ticks and some other useful helper structs that I have been abusing quite heavily.

plotter/invertedaxis_example_test.go
npotts: 👍

axis.go
npotts: 👍

axis.go
```diff
@@ -171,6 +171,17 @@ func (LogScale) Normalize(min, max, x float64) float64 {
 	return (log(x) - logMin) / (log(max) - logMin)
 }
 
+// InvertedScale can be used as the value of an Axis.Scale function to
+// invert the axis using any Normalizer.
+type InvertedScale struct{ Normalizer }
+
+var _ Normalizer = InvertedScale{}
+
+// Normalize returns a normalized [0, 1] value for the position of x
```
kortschak: Final dot.

plotter/invertedaxis_example_test.go
```diff
@@ -0,0 +1,56 @@
+// Copyright ©2017 The Gonum Authors. All rights reserved.
```
kortschak: 2018

plotter/invertedaxis_example_test.go
```diff
@@ -0,0 +1,56 @@
...
+	f := plotter.NewFunction(math.Exp)
+	f.XMin = 0.2
+	f.XMax = 10
+
+	f.Color = color.RGBA{R: 255, A: 255}
+
+	p.Add(f, plotter.NewGrid())
+	p.Legend.Add("exp(x)", f)
+
+	// Notice that both .Min and .Max for X and Y are both in 'normal' order
```
kortschak: s/Notice that b/B/ and final dot. I think something other than 'normal' would be better here.

plotter/invertedaxis_example_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"math"
+	"testing"
+
+	"gonum.org/v1/plot"
+	"gonum.org/v1/plot/internal/cmpimg"
+	"gonum.org/v1/plot/plotter"
+	"gonum.org/v1/plot/vg"
+)
+
+// Example_invertedScale shows how to create a plot with an inverted axes.
```
kortschak: This comment will not be rendered. If you want commentary in an example, it must be in the body. In that case you don't need to use godoc phrasing, just normal prose is OK.

npotts: I don't write too many Test Examples - Its just means I need more practice

I moved the comment into the body as my intent was to provide more prose on how to exactly use it. gonum/plot has a particular cadence/style/feel when compared to other plotting libraries - I feel that more hints as to design patterns should make first time users get moving a bit faster.

axis.go
npotts: ✔️

plotter/invertedaxis_example_test.go
kortschak: Yeah, no worries. You're doing good work and sometime the vagaries of godoc rendering a difficult (I usually have a godoc server running and check what the docs look like to make sure everything is good, particularly when examples are part of the docs.).

sbinet:
needs https://github.com/gonum/gonum/pull/726

plotter/invertedaxis_example_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"testing"
+
+	"gonum.org/v1/plot"
+	"gonum.org/v1/plot/internal/cmpimg"
+	"gonum.org/v1/plot/plotter"
+	"gonum.org/v1/plot/vg"
+)
+
+func Example_invertedScale() {
+	// This samples is nearly identical to the LogScale, other than
```
kortschak: "This example"?

sbinet:
thanks again and welcome

kortschak:
This needed fixes in the commit message.

Issue #491, opened 2019-01-07
Change vgsvg.DPI from 90 to 96


chmike:
### What are you trying to do?
Generate Plots saved in SVG.

### What did you do?
Saved the plot as SVG. Displayed the plot uisng the SVG previewer of studio code, and displayed the plot using Inkscape.

### What did you expect to happen?
Get same plot image as with PNG.

### What actually happened?
The image frame was bigger than the plot frame. The plot size was requested to be 4\*vg.Inch x 4\*vg.Inch. The generated svg xml had an svg tag with width and hight defined as 4in as expected. But the plot was smaller. 

![image](https://user-images.githubusercontent.com/235856/50763626-9e1f4f80-1270-11e9-8721-ab6ddaab0e0c.png)

This is because vgsvg assume that the DPI is 90 while it should be 96. According to [this stackoverflow question](
https://stackoverflow.com/questions/1346922/svg-and-dpi-absolute-units-and-user-units-inkscape-vs-firefox-vs-imagemagick) the DPI of svg has been aligned to the DPI of CSS which is 96. For that reason Inkscape changed it from 90 to 96 in version 0.92. In firefox, the DPI is also 96, but the resulting framing error is not directly visible since the margin is transparent.

[Here is another reference](https://stackoverflow.com/a/24127463/75517) stating that the DPI of SVG is aligned to the DPI of CSS and it is 96. 

### What version of Go and Gonum/plot are you using?
Latest as today, installed with go get. 

### Does this issue reproduce with the current master?
Don't know.

sbinet:
thanks for the issue.
could you test #492 and report whether it fixes it on your end?

thanks.

chmike:
Test is successful. Your change fix the issue.

PR #492, opened 2019-01-07
vg/vgsvg: bump DPI to 96, introduce functional options

sbinet:
Fixes gonum/plot#491.

Please take a look.


vg/vgsvg/vgsvg.go
```diff
@@ -46,32 +54,69 @@ type context struct {
...
+			panic("vgsvg: w and h must both be > 0")
+		}
+		c.w = w
+		c.h = h
+	}
+}
+
+// UseDPI sets the dots per inch of a canvas. It should only be
+// used as an option argument when initializing a new canvas.
+func UseDPI(dpi int) option {
```
kortschak: Any reason for integral DPI?

chmike:
I suggested to change the DPI value because the current value is incorrect. This should be fixed ASAP because it is a bug. 

It's OK to leave it as `int` for now because in reality we should be using the `viewBox` <svg> parameter instead of this DPI scaling. If the units of the input data are in Inches, we can define the <svg> like this: `<svg width="4in" height="4in" viewBox="0 0 4 4">` and don’t need to multiply each value by `vgsvg.DPI` anymore. A point with coordinate 0,0 will be top left, and a point with coordinate 4,4 will be bottom right.

I suggest, we first fix the DPI value to get correct svg images, and I work on a PR to replace the DPI values with the `viewBox` parameter. Is this suggestion OK ?

sbinet:
ping?

vg/vgsvg/vgsvg.go
kortschak: Let's revisit that in another issue.

Issue #377, opened 2017-08-11
plotter: plotting an histogram in logscale


sbinet:
hi,

consider the following code:

```go
func TestHistogramLogScale(t *testing.T) {
	rnd := rand.New(rand.NewSource(1))
	n := 1000
	vals := make(Values, n)
	for i := 0; i < n; i++ {
		vals[i] = rnd.NormFloat64()
	}

	p, err := plot.New()
	if err != nil {
		log.Panic(err)
	}
	p.Title.Text = "Histogram in logy"
	p.Y.Min = 0.1
	p.Y.Scale = plot.LogScale{}
	p.Y.Tick.Marker = plot.LogTicks{}

	h, err := NewHist(vals, 16)
	if err != nil {
		t.Fatal(err)
	}
	h.Normalize(1)
	p.Add(h)

	err = p.Save(200, 200, "testdata/histogram_log.png")
	if err != nil {
		t.Fatal(err)
	}
}
```

this will panic:
```
=== RUN   TestHistogramLogScale
--- FAIL: TestHistogramLogScale (0.01s)
panic: Values must be greater than 0 for a log scale. [recovered]
	panic: Values must be greater than 0 for a log scale.

goroutine 5 [running]:
testing.tRunner.func1(0xc42005a9c0)
	/usr/lib/go/src/testing/testing.go:622 +0x29d
panic(0x7d7160, 0xc4200f7790)
	/usr/lib/go/src/runtime/panic.go:489 +0x2cf
github.com/gonum/plot.LogTicks.Ticks(0x0, 0x3fda71c2dac35ff2, 0xc420123420, 0x4cbec8, 0x857cfb)
	/home/binet/work/gonum/src/github.com/gonum/plot/axis.go:429 +0x400
github.com/gonum/plot.(*LogTicks).Ticks(0xc89e68, 0x0, 0x3fda71c2dac35ff2, 0x4cc5a5, 0x857cfb, 0x9)
	<autogenerated>:42 +0x5d
github.com/gonum/plot.(*verticalAxis).size(0xc420123af8, 0x40278e0000000000)
	/home/binet/work/gonum/src/github.com/gonum/plot/axis.go:282 +0x7a
github.com/gonum/plot.(*Plot).Draw(0xc4200b3000, 0x7f6fa2e5d428, 0xc420058de0, 0x0, 0x0, 0x4069000000000000, 0x406712b000000000)
[...]

PR #493, opened 2019-01-07
plot,plotter: properly handle Log-Y scale in Histogram

sbinet:
Fixes gonum/plot#377.

Please take a look.


eaburns:
The golden image has only one labeled tick mark on the y-axis, so it's not possible to tell the scale. Shouldn't there be at least two, or is that a separate but?

eaburns:
Sorry for butting in again with little context, but it also looks like the h1 values at -2 don't show in the golden plot either. What happened to them?

sbinet:
>  but it also looks like the h1 values at -2 don't show

ah! quite right...

sbinet:
the values at -2 are at the bottom.
`Histogram` should probably implement either `GlyphBoxer` or "artificially" pull down the `ymin` value `DataRange` returns... arf!

sbinet:
PTAL.

I ended up modifying `Histogram.DataRange`, keeping track of both `ymin` and `ylow`, where `ymin` is the smallest Y value and `ylow` the smallest non-zero Y value.
if `ymin` is zero, `DataRange` returns `0.5*ylow` as `ymin`, to make sure the smallest bin is still displayed.

open to a more robust algorithm, though...

plotter/histogram.go
```diff
@@ -106,6 +115,16 @@ func (h *Histogram) DataRange() (xmin, xmax, ymin, ymax float64) {
...
 		}
+		if bin.Weight < ymin {
+			ymin = bin.Weight
+		}
+		if 0 < bin.Weight && bin.Weight < ylow {
+			ylow = bin.Weight
+		}
+	}
+	if ymin == 0 {
+		// reserve a bit of space for the smallest bin to be displayed still.
```
kortschak: s/reserve/Reserve/

plotter/histogram.go
```diff
@@ -106,6 +115,16 @@ func (h *Histogram) DataRange() (xmin, xmax, ymin, ymax float64) {
 		if bin.Weight > ymax {
 			ymax = bin.Weight
 		}
+		if bin.Weight < ymin {
+			ymin = bin.Weight
+		}
+		if 0 < bin.Weight && bin.Weight < ylow {
```
kortschak: This reads as a range check; bin.Weight in (0, ylow). But I don't think this is what you are doing; you are checking whether bin.Weight is non-zero and only using it then. Is this correct? If so, please rephrase this as `if bin.Weight != 0 && bin.Weight < ylow {`. If it's a check that bin.Weight is positive non-zero, then replace the `!=` in that with a `>`.

sbinet:
@decibelcooper 
> Also, what about adding a flag to Histogram for whether or not to suppress zeros? Perhaps it's better to have this off by default.

yeah... this has bothered me as well.
perhaps we should bite the bullet and just go with introducing a thin `HistogramLogY` type that wraps `Histogram` but has the zero-suppression built-in.

Commit 9ffa89d adds a special mode to `Histogram` to explictly support Y log-scale.
Less invasive but also less automated.

sbinet:
ping?

kortschak:
I don't understand the rationale for the first commit. We don't import "log" in plot/axis.go (or anywhere in plot).

sbinet:
the `func log(...)` function is quite annoying (to me at least) when I sprinkle code with `log.Printf("boo")` statements during my debugging sessions.

I can remove it and send it as part of another PR. (or not.)

plotter/histogram.go
```diff
@@ -32,6 +32,9 @@ type Histogram struct {
 	// LineStyle is the style of the outline of each
 	// bar of the histogram.
 	draw.LineStyle
+
+	// LogY enables support for histograms displayed with a Y log-scale.
```
kortschak: This comment is not quite right; `// LogY allows rendering with a log-scaled Y axis.` (or something similar), but also a brief explanation of what it is doing to make this possible.

kortschak:
>I can remove it and send it as part of another PR. (or not.)

SGTM

plotter/histogram.go
```diff
@@ -77,25 +77,34 @@ func (h *Histogram) Plot(c draw.Canvas, p *plot.Plot) {
...
 }
 
 // DataRange returns the minimum and maximum X and Y values
 func (h *Histogram) DataRange() (xmin, xmax, ymin, ymax float64) {
-	xmin = math.Inf(1)
+	xmin = math.Inf(+1)
 	xmax = math.Inf(-1)
+	ymin = math.Inf(+1)
 	ymax = math.Inf(-1)
+	ylow := math.Inf(+1) // smallest non-zero y value
```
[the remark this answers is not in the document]
sbinet: done.

plotter/histogram.go
sbinet: done.

sbinet:
(had to force-push b/c of the `logf` commit removal...)

plotter/histogram.go
kortschak: It would still be nice to have an explanation of what it does.

sbinet: tentatively: aca1101

decibelcooper:
> > I can remove it and send it as part of another PR. (or not.)
> 
> SGTM

Too bad... I, too, thought having a function called `log` is needlessly cumbersome.  Yay for the merge, though.

sbinet:
@decibelcooper don't worry, it's been removed from the top-level (but local) namespace of the `plot` package:
- https://github.com/gonum/plot/pull/502

:)

PR #505, opened 2019-01-19
vg/vgsvg: fix resolution handling for the general case

### LINKED ISSUE #504: vg/vgsvg: #492 has odd impact on image scale
Fix [regression](https://travis-ci.org/biogo/ncbi/jobs/481638043) in github.com/biogo/ncbi graphics package.

### What did you do?
The test cases for [github.com/biogo/ncbi/blast/graphic](https://github.com/biogo/ncbi/blob/master/blast/graphic/graphic_test.go) are not minimal, and I can possibly get something smaller if needed.

Test to pass generating the following image (among others).

A png resave of the correct image.

### What actually happened?
Test failed during cron job travis run (at https://github.com/biogo/ncbi/commit/c9503af863dd6f9fb527f83bab89d15128dcf3b1)

What appears to be happening is that the image is scaled to be larger than it should be. I suspected it was due to the change in default DPI from 90 to 96 in #492 and have confirmed this by changing that value back to 90 and fixing the issue.

A png resave of the failing svg output (note that it is scaled in x and y directions).

The [code in question](https://github.com/biogo/ncbi/blob/master/blast/graphic/graphic.go) does not refer to the DPI value at all. This is confirmed by replacing the vgsvg canvas with a vgimg canvas and getting expected results. This is what would be expected given that all the coordinates are represented in `vg.Length` units.

The following is the png output (substantively the same as the previous svg output and the svg output with DefaultDPI reverted to 90).

I suspect that somewhere in #492 (or missing from #492) there is a reference to the old value.

11e71620383853792e05e2ffd1ecb7c70877b363

Yes.

/cc @sbinet

kortschak:
See https://www.w3.org/TR/SVG11/coords.html#Units

Please take a look.

Fixes #504.


kortschak:
It's not, but it is related. I agree we should do that, but this is just bug fix.

When we do do that, I vote we change to points rather than inches.

kortschak:
I didn't realise with have stale branch checks on this repo. We don't now.

Can you please re-approve?

sbinet:
Going to pt sounds ok. (But I feel I am reaching my level of incompetence here...)

kortschak:
It's fairly straightforward. You specify the dimensions of the canvas using the units you want in the `width` and `height` attribs and then you make a mapping of user units to that with the `viewBox` attrib. If we use `<svg width="%.*gin" height="%.*gin" viewBox="0 0 %.*g %.*g"` with `pr, c.w.Length(), pr, c.h.Length(), pr, c.w.Length(), pr, c.h.Length()` and remove all the `Dots` calls, that should do it.

kortschak:
Which PR? This one?

chmike:
Yes, this one. 
Here is the code. It’s not wrong. I mean that it’s the default `viewBox` value.

``Go
<svg width="%.*gin" height="%.*gin" viewBox="0 0 %.*g %.*g"
	xmlns="http://www.w3.org/2000/svg"
	xmlns:xlink="http://www.w3.org/1999/xlink">`+"\n",
		pr, c.w/vg.Inch,
		pr, c.h/vg.Inch,
		pr, c.w.Dots(c.dpi),
		pr, c.h.Dots(c.dpi),
	)
```

chmike:
I just looked at #504 and I don’t understand the cause of the problem you have. 

This might be a plotter issue that tried to fix the DPI value error.

chmike:
@kortschak 
Could you test the svg image I gave to see if you might have a different DPI value ? 
The viewBox you suggest could partially solve the problem. But we would have a font scaling problem.

kortschak:
And yet it fixed a broken state. The plotter is not doing anything to the DPI value (as demonstrated by the png canvas output).

I suspect that what is going on is that it is indeed the default `viewBox` value, but despite claims to the contrary, different people believe different things about what the default DPI value is.

kortschak:
The svg you pasted above renders like so...

test

kortschak:
And with the view box...

fixed

Never trust SO, or claims of standardisation.

chmike:
@kortschak 
The link to the SVG 1.1 specification is in fact giving DPI 90 as an example. It say that it is defined by the context and that it could vary. 

In deep it shows that your DPI (defined by the context) is different from 96. It might be 90 for you.

Here is a screen capture of your image with the border visible.

image

chmike:
With your viewBox parameter I see this. 

image

I admit that it fixes the scaling issue when the effective dpi is not 96. 

I do have some doubts left about font scaling. The font scale is currently expressed in points in the generated graphs. It should then be specified in px units to adapt to the context dpi difference.

PR #507, opened 2019-01-19
vg/vgsvg: make user units equivalent to points

kortschak:
This *should* make us use points for user units, but it causes the text size to be too large.

Please take a look.

/cc @chmike


chmike:
The font size problem could be solved if the font size is specified in user units. Currently  a font size of 12pt is specified as "12pt". I suggest to specify it as "12". Logicaly it is the same since we now use points everywhere. But apparently svg makes an error when scaling the font.

kortschak:
@sbinet *ping*

vg/vgsvg/vgsvg.go
```diff
@@ -39,7 +35,7 @@ const (
 type Canvas struct {
 	svg  *svgo.SVG
 	w, h vg.Length
-	dpi  float64
+	// dpi  float64
```
sbinet: remove?

kortschak: Thanks

kortschak:
PTAL

vg/vgsvg/vgsvg.go
kortschak: BTW Do you know what `ht` is for. It is not used AFAICS.

sbinet: no idea :)
according to blame, @eaburns is the one who introduced it:
https://github.com/gonum/plot/blame/07f44254b21cff40a59809f4c926aa023286107d/vg/vgsvg/vgsvg.go

PR #509, opened 2019-01-23
vg/svg: remove unused field

kortschak:
Please take a look.

AFAICS this field is not used and never was.


kortschak:
In the first commit it existed, in vecsvg, it did the same work.

Issue #514, opened 2019-01-30
program stops when saving an svg plot


chmike:
### What are you trying to do?
Save a simple svg plot with 2 points and 2 lines. The program hangs.

### What did you do?
Every day I generate ~80 svg plots, and attach them to a mail that I send to myself.

### What did you expect to happen?
As usual, get the mail with the graphs. 

### What actually happened?
Didn't receive the mail this morning. Investigated, and saw that the program blocked when saving one particular graph which has only two points, and 2 lines. The problem occurs only with that particular graph. If I skip that graph, no problem. If I try to save only this particular graph, the program blocks. 

By using the debugger, to execute the Draw function step by step, the program stops when executing 
```Go 
marks := a.Tick.Marker.Ticks(a.Min, a.Max)        // in file `axis.go` at line 296
```
with `a.Min = 99.99999999999996` and `a.Max = 100`. 

The debugger never manages to step into that function. I have no idea why. 
This is for the Y axis. It uses the default ticks and legend. 

The plot has only two points, and two horizontal lines which have the following values:

```
Points: [{X:1.5486336e+09 Y:0.77} {X:1.54872e+09 Y:0.77}]
Line1: [{X:1.5486336e+09 Y:0.77} {X:1.54872e+09 Y:0.77}]
Line2: [{X:1.5486336e+09 Y:0.7699999999999998} {X:1.5486768e+09 Y:0.7699999999999998} {X:1.54872e+09 Y:0.7699999999999998}]
```
Before passing the values to Plot, I normalize the Y values by multiplying them by (100/0.77) which yields 100. Could the different, but very close, values be the cause of the problem ? 

### What version of Go and Gonum/plot are you using?
Go1.11.4

### Does this issue reproduce with the current master?
Yes.

sbinet:
could you show us a simple reproducer program?
thanks.

sbinet:
> Could the different, but very close, values be the cause of the problem ?

(and, yes, I think that's the culprit)

chmike:
Here is the most simple program producing the problem. 

``` Go
package main

import (
	"gonum.org/v1/plot"
	"gonum.org/v1/plot/plotter"
	"gonum.org/v1/plot/vg"
)

func main() {
	p, err := plot.New()
	if err != nil {
		panic(err)
	}
	//p.Title.Text = "Test plot"
[...]

PR #515, opened 2019-01-30
plot: make sure there is enough ULP space for axis ranges

sbinet:
Fixes gonum/plot#514.

Please take a look.

### DIFF
--- a/axis.go
+++ b/axis.go
@@ -145,6 +145,19 @@ func (a *Axis) sanitizeRange() {
 		a.Min--
 		a.Max++
 	}
+
+	var (
+		min  = math.Float64bits(a.Min)
+		max  = math.Float64bits(a.Max)
+		nULP = max - min
+	)
+	const minULP = 20 // arbitrary value found experimentally with gonum/plot#514.
+	// add enough ULP space between min and max
+	// see https://github.com/gonum/plot/issues/514
+	if nULP <= minULP {
+		a.Min = math.Float64frombits(min - minULP/2)
+		a.Max = math.Float64frombits(max + minULP/2)
+	}
 }
 
 // LinearScale an be used as the value of an Axis.Scale function to
--- a/axis.go
+++ b/axis.go
@@ -145,20 +145,6 @@ func (a *Axis) sanitizeRange() {
 		a.Min--
 		a.Max++
 	}
-
-	var (
-		min  = math.Float64bits(a.Min)
-		max  = math.Float64bits(a.Max)
-		nULP = max - min
-	)
-	const minULP = 20 // arbitrary value found experimentally with gonum/plot#514.
-	// Add enough ULP space between min and max
-	// see https://github.com/gonum/plot/issues/514
-	if nULP < minULP {
-		ulps := minULP - nULP
-		a.Min = math.Float64frombits(min - ulps/2)
-		a.Max = math.Float64frombits(max + ulps/2)
-	}
 }
 
 // LinearScale an be used as the value of an Axis.Scale function to
--- a/axis.go
+++ b/axis.go
@@ -152,11 +152,12 @@ func (a *Axis) sanitizeRange() {
 		nULP = max - min
 	)
 	const minULP = 20 // arbitrary value found experimentally with gonum/plot#514.
-	// add enough ULP space between min and max
+	// Add enough ULP space between min and max
 	// see https://github.com/gonum/plot/issues/514
-	if nULP <= minULP {
-		a.Min = math.Float64frombits(min - minULP/2)
-		a.Max = math.Float64frombits(max + minULP/2)
+	if nULP < minULP {
+		ulps := minULP - nULP
+		a.Min = math.Float64frombits(min - ulps/2)
+		a.Max = math.Float64frombits(max + ulps/2)
 	}
 }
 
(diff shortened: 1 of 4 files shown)


sbinet:
@chmike

axis.go
```diff
@@ -145,6 +145,19 @@ func (a *Axis) sanitizeRange() {
...
+	var (
+		min  = math.Float64bits(a.Min)
+		max  = math.Float64bits(a.Max)
+		nULP = max - min
+	)
+	const minULP = 20 // arbitrary value found experimentally with gonum/plot#514.
+	// add enough ULP space between min and max
+	// see https://github.com/gonum/plot/issues/514
+	if nULP <= minULP {
+		a.Min = math.Float64frombits(min - minULP/2)
```
kortschak: Not that this is incorrect, but this gives a range that is higher than minULP for all cases. Is that what you want?

axis.go
```diff
@@ -145,6 +145,19 @@ func (a *Axis) sanitizeRange() {
...
 		a.Max++
 	}
+
+	var (
+		min  = math.Float64bits(a.Min)
+		max  = math.Float64bits(a.Max)
+		nULP = max - min
+	)
+	const minULP = 20 // arbitrary value found experimentally with gonum/plot#514.
+	// add enough ULP space between min and max
```
kortschak: s/add/Add/

axis.go
sbinet: It's an oversight.
Thanks for spotting this.

Are you onboard with the general direction of this PR?
And the choice of handling this issue at the level of the Axis instead of at the level of the Talbot algorithm?

kortschak: Yes, it seems reasonable to me.

sbinet:
PTAL

kortschak:
I've had a look at the failure and compared it to the reference implementation. It fails with a warning. The actual fault is a pretty straightfoward failure to progress a float increment. I think that that should be fixed (as well? - notwithstanding my SGTM above). The choice of what to do here with the axis range comes down to aesthetics and I'll leave that up to you, but I'd like the following changes made as well (this prevents a failure in the ULP magiv number from missing any pathological cases).

```diff
diff --git a/labelling.go b/labelling.go
index c5c66ea..410812f 100644
--- a/labelling.go
+++ b/labelling.go
@@ -123,7 +123,7 @@ outer:
 
                                        minStart := (math.Floor(dMax/step) - float64(have-1)) * float64(skip)
                                        maxStart := math.Ceil(dMax/step) * float64(skip)
-                                       for start := minStart; start <= maxStart; start++ {
+                                       for start := minStart; start <= maxStart && start != start-1; start++ {
                                                lMin := start * fracStep
                                                lMax := lMin + kStep
 
diff --git a/labelling_test.go b/labelling_test.go
index 8c6d20f..101bdd8 100644
--- a/labelling_test.go
+++ b/labelling_test.go
@@ -19,7 +19,7 @@ var talbotLinHanrahanTests = []struct {
        wantStep      float64
        wantMagnitude int
 }{
-       // Expected values confirmed against R reference imlpementation.
+       // Expected values confirmed against R reference implementation.
        {
                dMin:        -1.9846500878911073,
                dMax:        0.4370974820125605,
@@ -220,6 +220,19 @@ var talbotLinHanrahanTests = []struct {
                wantStep:      15,
                wantMagnitude: -2,
        },
+
+       // The following fails in the reference implementation with a
+       // warning of "probable complete loss of accuracy in modulus".
+       {
+               dMin:        99.99999999999996,
+               dMax:        100,
+               want:        3,
+               containment: free,
+
+               wantValues:    []float64{99.99999999999996, 99.99999999999997, 100},
+               wantStep:      2,
+               wantMagnitude: -14,
+       },
 }
 
 func TestTalbotLinHanrahan(t *testing.T) {
```

This will give arguably reasonable aesthetics for all values of `dmin`, `dmax` where `ulp(dmax)-ulp(dmin) > 2`. So there probably is some argument that there should be some range sanitisation. However, the converse argument is that people should not be expecting float values to behave like real numbers; they aren't and don't have infinite precision.

sbinet:
PTAL

sbinet:
taking a step back, I see a couple of strategies when the difference between `axis.Min` and `axis.Max` is less than `N` ULPs:

- consider them as equal and do `a.Min--; a.Max++`
- extend the range so its at least `N` ULPs
- allow to pass a "sanitize" function (a `nil` field of `Axis`?) that will kick in and do its thing

kortschak:
... or do nothing and ensure that the code doesn't hang. That is the change in labelling.go — without that change, the test I added fails to complete.

kortschak:
For low ULP range, make the range a (the closest above) multple of the number of major ticks (we use 3).

I share @chmike concerns; just ramping the space up when it's below some arbitrary threshold is not a good idea. Though the labels there are not entirely relevant to this - that is just the rendering and is under control of the user.

sbinet:
ok, then 6bd75af0bbbc2ed3b65e50e275f37fc66d061922 has only the Talbot fix.
(the comment of the final commit should be modified to reflect that.)

PTAL ?

kortschak:
That's fine. Do we want to keep an issue open for an approach to improving the aesthetics if necessary?

sbinet:
> Do we want to keep an issue open for an approach to improving the aesthetics if necessary?

meh. if it annoys somebody they will file an issue?

chmike:
Have we merged this PR ? If yes, can we close it, as well as the corresponding Issue #514 that I raised ?

Issue #519, opened 2019-02-04
vgsvg enhancement: use <style> to reduce svg code size


chmike:
### What are you trying to do?
Make the svg code more compact. 

### What you suggest to do ?
Define styles in style blocks to avoid describing the full style repeatedly in each tag.  Use [style definition](https://developer.mozilla.org/fr/docs/Web/SVG/Element/style) in the svg code.

chmike:
Code ready, but there is one caveat. 

Styles defined inside `<style>` have a file wide scope. If multiple svg are copied inside the same html file, we would have class name collisions.

The recommended solution is to add a unique ID (e.g. randomly picked) to the svg tag and prepend class selector in the style with this tag id. 

I would like to know what I should do about it. I can use a predefined id (e.g. svgID) that the user may  then replace with another one when needed. This replacement can be automated without too much trouble: `id="svgID"` -> `id="newID"` and ` #svgID .st` -> ` #newID .st`

Another solution could be to generate a random id, but this makes testing by comparison with a golden svg no possible anymore. 

What should I do ?

chmike:
Opted for a predefined svg id "svgID" to ease testing by comparison. 

The absence of class name conflicts when two plots with different style are displayed by the same html file remains to be tested.

chmike:
I discovered three regressions with this change. 

The first regression is that the image viewer of Gnome can’t display images with CSS styles. It can display images without this optimization. 

The second regression is that we can’t view, edit or annotate the plots with InkScape (a vector image editor). I don’t know the status with Adobe Illustrator. 

A third regression is that the svg with styles can’t be processed by imgur. I guess they use the same lib to render svg as eye of Gnome, the image viewer of Gnome.

In my opinion, these are sufficient to justify a canceling of the PR.

kortschak:
Please file an issue and we can send a PR to revert this.

PR #520, opened 2019-02-07
vg/vgsvg: add class style for compactness

chmike:
Fixes gonum/plot#519

Please take a look.

- I verified that there is no class name conflict when an html displays two plots (\) with different styles. 
- the size of scatter.svg is halved with this change !


decibelcooper:
ooo, much excite :+1:

kortschak:
We'll need to have all the SVG golden images regenerated. You can do this with the `-regen` flag when you run tests. Please don't commit any other changed golden images, only SVG.

chmike:
Sorry, I don't understand your comment. At which level should I run the 
`go test -regen` ?

sbinet:
I'd suggest:
```
$> cd gonum.org/v1/plot
$> go test -regen ./...
```
and then only `git add` the golden `.svg` files.

chmike:
I don’t fully understand the testing framework. 

The files in `vg/testdata/width_-1.svg`, etc. are not regenerated with the parameter -regen. It seam that the test `TestLineWidth` is not executed.

chmike:
Here is he trick. 

The command *must* be `go test -v ./... -regen` and not `go test -v -regen ./...`. When I use the former command, the tests are executed recursively in sub-directories, but I get a lot of errors because the flag `-regen` is not defined for most tests. With the later command the tests are not executed in sub-directories. After that, the files are all regenerated, and subsequent calls to `go test ./...` succeed (no failures).

vg/vgsvg/vgsvg.go
```diff
@@ -461,3 +482,13 @@ func opacityString(clr color.Color) string {
 	_, _, _, a := clr.RGBA()
 	return fmt.Sprintf("%.*g", pr, float64(a)/math.MaxUint16)
 }
+
+// return the class name
```
sbinet: this comment doesn't add too much, and doesn't follow the usual Go code documentation conventions.
could you please add some more context?

vg/vgsvg/vgsvg.go
```diff
@@ -483,8 +483,9 @@ func opacityString(clr color.Color) string {
 	return fmt.Sprintf("%.*g", pr, float64(a)/math.MaxUint16)
 }
 
-// return the class name
-func (c *Canvas) getClass(style string) string {
+// getClassName returns the class name associate to the given style.
```
sbinet: `s/associate to the/associated with a/`

vg/vgsvg/vgsvg.go
```diff
@@ -483,8 +483,9 @@ func opacityString(clr color.Color) string {
...
+// A new class name is generated and returned for a new style.
```
sbinet: I would perhaps go with:
```go
// A different class name is generated for each new style.
```

chmike:
PTAL

vg/vgsvg/vgsvg.go
```diff
@@ -461,3 +482,14 @@ func opacityString(clr color.Color) string {
 	_, _, _, a := clr.RGBA()
 	return fmt.Sprintf("%.*g", pr, float64(a)/math.MaxUint16)
 }
+
+// getClassName returns the class name associated to the given style.
+// A different class name is generated for each new style.
+func (c *Canvas) getClassName(style string) string {
```
kortschak: I'd also change the name of this method. Getters in Go are generally spelled without the "get" prefix unless that would make them zero length. I'd call this `className` or preferably `classNameOf`.

vg/vgsvg/vgsvg.go
```diff
@@ -37,8 +39,10 @@ type Canvas struct {
 	svg  *svgo.SVG
 	w, h vg.Length
 
-	buf   *bytes.Buffer
-	stack []context
+	buf    *bytes.Buffer
+	stack  []context
+	styles map[string]string
+	svgID  string
```
kortschak: Why does this need to exist, it is never mutated after construction with `NewWith` as far as I can see.

vg/vgsvg/vgsvg.go
```diff
@@ -73,11 +77,13 @@ func New(w, h vg.Length) *Canvas {
...
-		svg:   svgo.New(buf),
-		w:     DefaultWidth,
-		h:     DefaultHeight,
-		buf:   buf,
-		stack: []context{{}},
+		svg:    svgo.New(buf),
+		w:      DefaultWidth,
+		h:      DefaultHeight,
+		buf:    buf,
+		stack:  []context{{}},
```
kortschak: Please make this `make([]context, 1)` if you want an explicit unit-length slice.

vg/vgsvg/vgsvg.go
chmike: The idea was to define the svgID in one location, and I initially generated a random svgID. Now that it's a constant, I should indeed use a const string.

vg/vgsvg/vgsvg.go
chmike: That's not my code. It was there before. The line was modified because of the alignment change. 

I'm not sure what a context is. I suspect it is a shorthand notation to instantiate a slice of slice structure, or something like that.

chmike: You were right. I fixed it.

chmike:
PTAL

kortschak:
Thanks

PR #525, opened 2019-02-15
Revert "vg/vgsvg: add class style for compactness"

### LINKED ISSUE #524: vg/svg: due to regressions, cancel PR #520 using styles for svg compactness
I discovered three regressions resulting from the use of \<style\> in svg plots to reduce svg size.

The first regression is that the image viewer of Gnome ("Eye of Gnome") can’t display the plots. It displays the image as a black rectangle. 

The second regression is that imgur, an image sharing site, display the svg plots as black rectangles. I guess they use the same lib as eye of Gnome to render svg (librsvg).

The third regression is that Inkscape can’t interpret the styles and show a plot as a black rectangle. It is thus not possible to view, edit or annotate the plots with Inkscape. I don’t know the status with Adobe Illustrator.

I thus suggest to cancel the PR #520 which added the styles for compactness.

kortschak:
This reverts commit bbc78a95e31f693ae360e101cd1a53ad3775fc87.

If libsvg fixes its behaviour we can revisit this.

Please take a look.

Fixes #524.


kortschak:
Failure on master from this is due to coveralls.io inability to pull the repo with any degree of consistence lemurheavy/coveralls-public#1264.

PR #527, opened 2019-02-20
vgimg: translate to gg arc end point concept

kortschak:
This concerts from (theta,phi) that is used by vg to (theta_1,theta_2) that is used by gg.

Please take a look.


kortschak:
There will be tests when rings is moved to plot.

Issue #536, opened 2019-03-11
vg/vgpdf: latest gofpdf (24315ac) confuses vgpdf


sbinet:
with latest gofpdf commit (`24315ac`), we get these "stray" characters in the resulting PDF file:

```
$> cd vg/vgpdf
$> head testdata/arc*.pdf
==> testdata/arc_golden.pdf <==
%PDF-1.3
3 0 obj
<</Type /Page
/Parent 1 0 R
/Resources 2 0 R
/Contents 4 0 R>>
endobj
4 0 obj
<</Filter /FlateDecode /Length 9580>>
stream

==> testdata/arc.pdf <==
/GS1 gs
%PDF-1.4
3 0 obj
<</Type /Page
/Parent 1 0 R
/Resources 2 0 R
/Group <</Type /Group /S /Transparency /CS /DeviceRGB>>
/Contents 4 0 R>>
endobj
4 0 obj
```

notice the `/GS1 gs` ?
it's also there for the `multipage.pdf` example.

the latest working revision is `v1.0.2` (`6b7c17b`)

PR #537, opened 2019-03-11
vg/vgpdf: create a PDF page context before setting PDF attributes

sbinet:
Fixes gonum/plot#536.

Please take a look.


kortschak:
Does this work with both the previous version of gofpdf and the current? Should go.mod/go.sum be updated?

sbinet:
yes*2.

this works with the previous version of gofpdf (but previous version sets `%PDF-1.3` while the latest one sets `%PDF-1.4`)

I've updated go.mod + go.sum.

PR #538, opened 2019-03-11
ci: enable Go modules

sbinet:
Please take a look.


sbinet:
I might have been lazy by just copying the `gopher.png` logo from `gonum/gonum` into `gonum/plot/plotter/testdata`...

PR #541, opened 2019-03-11
vg/vg{img,pdf}: do not display text with zero font size

sbinet:
Fixes gonum/plot#540.

PR #542, opened 2019-03-12
vg: document behaviour of zero font size text

sbinet:
Please take a look.

### DIFF
--- a/vg/vg.go
+++ b/vg/vg.go
@@ -74,6 +74,7 @@ type Canvas interface {
 
 	// FillString fills in text at the specified
 	// location using the given font.
+	// If the font size is zero, the text is not drawn.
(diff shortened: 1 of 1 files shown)


sbinet:
as requested by @kortschak in https://github.com/gonum/plot/pull/541#issuecomment-471743842.

the behaviour for zero-width lines is actually already documented in `vg.Canvas`:
https://github.com/gonum/plot/blob/master/vg/vg.go#L20
```go
type Canvas interface {
	// SetLineWidth sets the width of stroked paths.
	// If the width is not positive then stroked lines
	// are not drawn.
	//
	// The initial line width is 1 point.
	SetLineWidth(Length)
...
}
```

PR #543, opened 2019-04-04
plotter: add vector field plotter

kortschak:
Please take a look.

/cc @chewxy


plotter/field_test.go
```diff
@@ -0,0 +1,146 @@
+// Copyright ©2019 The Gonum Authors. All rights reserved.
```
ctessum: To make the examples more legible in godoc, it might make sense to split the examples and tests up into individual files. From the [documentation](https://blog.golang.org/examples):

> A whole file example is a file that ends in _test.go and contains exactly one example function, no test or benchmark functions, and at least one other package-level declaration. When displaying such examples godoc will show the entire file.

plotter/field_test.go
```diff
@@ -0,0 +1,146 @@
...
+type field struct {
+	c, r int
+	fn   func(x, y float64) plotter.XY
+}
+
+func (f field) Dims() (c, r int)           { return f.c, f.r }
+func (f field) Vector(c, r int) plotter.XY { return f.fn(f.X(c), f.Y(r)) }
+func (f field) X(c int) float64 {
+	if c < 0 || c >= f.c {
+		panic("index out of range")
```
ctessum: "x index out of range"

plotter/field_test.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+	return float64(c - f.c/2)
+}
+func (f field) Y(r int) float64 {
+	if r < 0 || r >= f.r {
+		panic("index out of range")
```
ctessum: "y index out of range"

plotter/field.go
```diff
@@ -0,0 +1,188 @@
...
+	// vector glyph. The function should draw a unit
+	// vector to (1, 0) on the vg.Canvas, c.
+	// The direction and magnitude of v can be used
+	// to determine properties of the glyph drawing
+	// but should not be used to determine size or
+	// directions of the glyph.
+	//
+	// If DrawGlyph is nil, a simple black arrow will
+	// be drawn.
+	DrawGlyph func(c vg.Canvas, v XY)
```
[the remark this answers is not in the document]
ctessum: or maybe the normalized magnitude.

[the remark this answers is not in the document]
ctessum: or, actually, providing the row and column indexes as arguments would probably be better, because then the user can pull in whatever external information they need.

plotter/field.go
```diff
@@ -0,0 +1,188 @@
...
+	"gonum.org/v1/plot/vg/draw"
+)
+
+// FieldXY describes a two dimensional vector field where the
+// X and Y coordinates are arranged on a rectangular grid.
+type FieldXY interface {
+	// Dims returns the dimensions of the grid.
+	Dims() (c, r int)
+
+	// Vector returns the value of a grid value at (c, r).
```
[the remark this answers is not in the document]
kortschak:
```
	// Vector returns the value of a vector field at (c, r).
	// It will panic if c or r are out of bounds for the field.
```

field.go is derived heavily from heat.go and there are copy/paste errors that reflect that. This is one of those.

plotter/field.go
```diff
@@ -0,0 +1,188 @@
...
+	// It will panic if c or r are out of bounds for the grid.
+	Vector(c, r int) XY
+
+	// X returns the coordinate for the column at the index x.
```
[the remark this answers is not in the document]
kortschak: Done. This will also need to be fixed in heat.go.

plotter/field.go
```diff
@@ -0,0 +1,188 @@
...
+	// If DrawGlyph is nil, a simple black arrow will
+	// be drawn.
+	DrawGlyph func(c vg.Canvas, v XY)
+
+	// max define the dynamic range of the field.
+	max float64
+}
+
+// NewField creates as new heat map plotter for the given data,
+// using the provided palette.
```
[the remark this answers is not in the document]
kortschak: Fixed

plotter/field.go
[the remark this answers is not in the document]
kortschak: If the user wants to change the colour based on the magnitude, they can calculate the magnitude from `v`. This was the major reason v was included (which is the normalised vector).

I don't like the idea of a completely free approach to transforms. It opens up whole new world of issues to have to answer. This is simple and works for the vast majority of cases with close to zero work for the user.

I can see some merits in providing the row/column, but I need to think about it some more.

plotter/field_test.go
kortschak: I think this is another PR. The example code that this is adapted from, heat_test.go is the same. In fact I think they all are because the `TestX` funcs are in the same files.

plotter/field_test.go
kortschak: This is example code, but OK.

Same in heat_test.go.

plotter/field_test.go
kortschak: Ditto

plotter/field.go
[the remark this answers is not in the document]
ctessum: I somehow missed the fact that `v` was already included as an argument.

kortschak:
I'll wait for @sbinet as well.

plotter/field.go
```diff
@@ -0,0 +1,196 @@
...
+		LineStyle: DefaultLineStyle,
+		max:       max,
+	}
+}
+
+// Plot implements the Plot method of the plot.Plotter interface.
+func (f *Field) Plot(c draw.Canvas, plt *plot.Plot) {
+	c.Push()
+	c.SetLineStyle(f.LineStyle)
+	defer c.Pop()
```
sbinet: any reason we don't schedule that `defer` closer to its `c.Push` counterpart?
(if there are, they should be mentioned.)

plotter/field.go
```diff
@@ -0,0 +1,196 @@
...
+
+			c.Push()
+			c.Translate(vg.Point{X: (x + dx) / 2, Y: (y + dy) / 2})
+			v := f.FieldXY.Vector(i, j)
+			c.Rotate(math.Atan2(v.Y, v.X))
+			s := math.Hypot(v.X, v.Y) / (2 * f.max)
+			v.X *= s
+			v.Y *= s
+			c.Scale(s*float64(dx-x), s*float64(dy-y))
+			if f.DrawGlyph == nil {
```
sbinet: I'd use a `switch` there, to be consistent with Go style, but ok.

kortschak: I think this is fine; it's a single case.

plotter/field.go
kortschak: No, none real. I can change it.

kortschak:
PTAL

kortschak:
AppVeyor is horrifically slow here. Do we use your identity on that @sbinet? We probably shouldn't since you're doing other things as well. Do they have organisation accounts like travis does? I have to say, the service they have is pretty disappointing.

sbinet:
their handling of org accounts is a bit confusing, and, at the very least not very straighforward, indeed...

PR #546, opened 2019-04-06
plotter: fix documentation typo and clarify example

### LINKED ISSUE #544: plotter: fix documentation in HeatMap and examples
See https://github.com/gonum/plot/pull/543#discussion_r272209426, https://github.com/gonum/plot/pull/543#discussion_r272245652 and https://github.com/gonum/plot/pull/543#discussion_r272245771.

kortschak:
Please take a look.

Fixes #544.

### DIFF
--- a/plotter/heat.go
+++ b/plotter/heat.go
@@ -24,7 +24,7 @@ type GridXYZ interface {
 	// It will panic if c or r are out of bounds for the grid.
 	Z(c, r int) float64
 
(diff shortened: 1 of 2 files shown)


kortschak:
I'm thinking also that maybe `DrawGlyph` should take a `draw.LineStyle`. This would mean that the function does not need to be a closure and we can provide a set of them (or others can). WDYT?

PR #547, opened 2019-04-06
plotter: add colour magnitude vector field example

kortschak:
Please take a look.


kortschak:
... and now in the correct PR.

I'm thinking also that maybe `DrawGlyph` should take a `draw.LineStyle`. This would mean that the function does not need to be a closure and we can provide a set of them (or others can). WDYT?

chewxy:
Is the `draw.LineStyle`  such that a pre-specified linestyle may be passed onto the `DrawGlyph` function? Chances are people are gonna write custom `DrawGlyph` functions anyway... but I suppose it may be a nice option to have a default `DrawGlyph` function, then you just pass in the `LineStyle`

kortschak:
Yes, that was my thinking. If we have common arrows, coloured arrows etc, that would be helpful. Otherwise we need constructors to insert the style into a closure.

I'll push what I have here.

kortschak:
PTAL

kortschak:
*ping*

chewxy:
The default DrawGlyph function is not in yet?

kortschak:
It `DrawGlyph` is nil `Field` uses `drawVector`.

PR #551, opened 2019-05-15
github: add note about using play.golang.org

kortschak:
Please take a look.

### DIFF
--- a/.github/ISSUE_TEMPLATE
+++ b/.github/ISSUE_TEMPLATE
@@ -11,7 +11,14 @@ https://groups.google.com/forum/#!forum/gonum-dev
 
 
 ### What did you do?
-<!-- Please include a link to a minimal reproducer here. -->
+<!-- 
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE
```diff
@@ -11,7 +11,14 @@ https://groups.google.com/forum/#!forum/gonum-dev
 
 
 ### What did you do?
-<!-- Please include a link to a minimal reproducer here. -->
+<!-- 
+Please include a link to a minimal reproducer here. Preferably this
+should be a complete runnable example at https://play.golang.org/.
```
sbinet: same comment than in https://github.com/gonum/gonum/pull/983#discussion_r284112067

(the template could even pre-import the `x/tour/pic` package)

kortschak: Same answer.

I thought about doing that, but I don't know how helpful it will be. If it becomes an issue we can link to an example.

PR #555, opened 2019-05-26
plotter: Implement XYers interface for polygons

### LINKED ISSUE #442: plotter: NewPolygon should take an interface rather than a list of interfaces
Right now, plotter.NewPolygon takes `...XYer` as an input. However, it seems like it is difficult to actually get data in that format.

Would it make sense to add an interface something like:

```go
XYers interface{
    Len() int
    LenAt(int) int
    XY(int, int) (x, y float64)
}
```

That might be easier to implement.

ctessum:
This makes it possible for external types to implement the interface 
without copying all data points.

Fixes #442.

Please take a look.

### DIFF
--- a/plotter/plotter.go
+++ b/plotter/plotter.go
@@ -119,6 +119,27 @@ type XYer interface {
 	XY(int) (x, y float64)
 }
 
+// XYers wraps a set of (x, y) pair sets.
+type XYers interface {
+	// Len returns the number of XYers.
+	Len() int
+
+	// LenAt returns the length of XYer i.
+	LenAt(i int) int
+
+	// XYer returns a set of x, y pairs.
+	XY(i, j int) (x, y float64)
+}
+
+// xyerAt implements the XYer interface for an item in an XYers.
+type xyerAt struct {
+	XYers
+	i int
+}
+
+func (xy xyerAt) Len() int                { return xy.XYers.LenAt(xy.i) }
+func (xy xyerAt) XY(j int) (x, y float64) { return xy.XYers.XY(xy.i, j) }
+
 // XYRange returns the minimum and maximum
 // x and y values.
 func XYRange(xys XYer) (xmin, xmax, ymin, ymax float64) {
--- a/plotter/polygon.go
+++ b/plotter/polygon.go
@@ -34,11 +34,11 @@ type Polygon struct {
 // differently, but all built-in backends treat inner rings
 // with the opposite winding order from the outer ring as
 // holes.
-func NewPolygon(xys ...XYer) (*Polygon, error) {
-	data := make([]XYs, len(xys))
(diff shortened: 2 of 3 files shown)


plotter/plotter.go
```diff
@@ -119,6 +119,27 @@ type XYer interface {
 	XY(int) (x, y float64)
 }
 
+// XYers wraps a set of (x, y) pair sets.
+type XYers interface {
+	// Len returns the number of XYers.
+	Len() int
+
+	// LenAt returns the length of XYer i.
```
kortschak: Is there a reason to not define this as

```
type XYers interface {
    // Len returns the number of XYers.
    Len() int

    // XYer returns the XYer at position i.
    XYer(i int) XYer
}
```

This seems simpler for the user to create if needed.

ctessum: My understanding is that if the interface returns `plotter.XYer`, then every type that implements it also needs to return `plotter.XYer` (returning an interface with the same signature is not allowed). So every package that implements this interface would have to import `plotter`, and the same interface could never really be used for anything else.

This problem is avoided if the interface only includes built-in types.

For my specific use case, I have a general purpose [geometry](https://godoc.org/github.com/ctessum/geom) library and I'm trying to make it convenient to plot the polygons, but I don't really want to include `plotter` as a dependency.

kortschak: Yeah, OK. The API is not very nice like this though. I'll let someone else comment. @sbinet?

sbinet: just a quick note to say I haven't forgotten about this.
I'll circle back to it before the end of the week.

sbinet: I'd be more in favour of Dan's API as well.

I understand you don't want `plotter` as a dependency of `github.com/ctessum/geom`, and I agree it isn't great for the separation of concerns (or just the sheer amount of dependencies that come with `gonum/plot`.)
I also have had that concern and this kind of issue for my `hep/hbook` package that deals with histograms, but just the statistical aspects of histograms (1D, 2D, ...), not dealing with actually displaying them.

I didn't want `hbook` to be tied to a specific graphical library having been bitten by this kind of hard dependency in the (C++) particle physics world.
what I did was to create a separate package would depend on `hep/hbook` and `gonum/plot` to hold types that would implement the needed interfaces for `gonum/plot` to work, wrapping `hep/hbook` values. (it's `hep/hplot`: https://godoc.org/go-hep.org/x/hep/hplot)

wouldn't this kind of approach (the usual additional indirection switcheroo trick) work for you as well?

ctessum: It seems to me like that would result in a worse overall API experience, and I feel like there's a lot to be said for the interfaces only returning built-in types, but if that's the consensus, I'm alright with it I guess.

PR #583, opened 2020-02-27
plot: introduce unexported orientation type

sbinet:
Please take a look.


sbinet:
one could also consider adding a field `Orientation` to the `Axis` type (and export that new `orientation` type)

PR #556, opened 2019-06-07
show the value of each bar on top of it.

ffhelicopter:
show the value of each bar on top of it.

#475

PR #585, opened 2020-03-11
Show value of each bar on top of it

### LINKED ISSUE #475: plot: Can't add data labels to the barchart
I am trying to put values over a bar chart.

### What did you do?
	bars, err := plotter.NewBarChart(groupTest, vg.Points(100))
	if err != nil {
		panic(err)
	}

	bars.LineStyle.Width = vg.Length(0)
	bars.Color = plotutil.Color(0)

The data labels (value of each bar) on the top of each bar

### What actually happened?
Right now there is no field in the BarChart struct which enables one to inserts data labels.

Go: 1.10.3
Gonum: [5f3c436](https://github.com/gonum/plot/commit/5f3c436ce602d80b13f2936b6bf149515198d6a4)

agzuniverse:
This change modifies the bar charts to show their values as a label on top of each bar if the user sets the new `b.ShowLabel` option to true.

This PR builds on #556 
Closes #475


agzuniverse:
@kortschak you mentioned #556 cannot be merged because it does not pass the CI build, so here's a PR which does.

plotter/barchart.go
```diff
@@ -48,6 +49,10 @@ type BarChart struct {
 	// stackedOn is the bar chart upon which
 	// this bar chart is stacked.
 	stackedOn *BarChart
+
+	// ShowLabel determines whether the value of the bars should be
+	// shown above it or not.
+	ShowLabel bool
```
kortschak: `ShowValue`?

xiaofeng-zerone:
When can master branches use this feature?

kortschak:
It would need to be merged first, and then it will be included in the next minor release. In order to be merged it needs to pass review, which will require that the change requests are satisfied.

plotter/barchart.go
agzuniverse: I've changed it to `ShowValue`.

agzuniverse:
I'll look into adding a test for this soon.

lynxplay:
Is there any plan on moving forward with this feature ?
It seems like a really nice addition and I'd personally love to use the master branch of gonum/plot rather than my own fork just for this feature.

agzuniverse:
I think the only reason this PR wasn't merged is because I didn't write tests for this change. I thought of doing it but unfortunately couldn't find the time. @lynxplay if you could add a test the maintainers will probably merge this change in to master.

lynxplay:
Yeah I tried looking into this but the generated plots don't respect the written values on top of the bars when cropping so one value is always cropped off. 

Tests on this are mainly just a single generated graph and an image comparison no?

sbinet:
yes, we like to have a simple `testdata/foo_golden.{png,svg,...}` reference file and a test (or an example) generating that file to compare with and make sure we don't introduce regressions when refactoring code.

here is an example:
- https://github.com/gonum/plot/blob/v0.9.0/plotter/barchart_example_test.go#L159 (example generating `testdata/barChart_positiveNegative.png`)
- https://github.com/gonum/plot/blob/v0.9.0/plotter/barchart_test.go#L19 (the test that runs `ExampleBarChart_positiveNegative` and leverages the `cmpimg` package that compares with reference files)
- https://github.com/gonum/plot/blob/v0.9.0/plotter/testdata/barChart_positiveNegative_golden.png (the reference file)

(with my apologies for the belated answer.)

PR #587, opened 2020-03-20
vg/draw: add support for .tex to NewFormattedCanvas and plot.Plot.Save

sbinet:
Please take a look.

### DIFF
--- a/vg/draw/draw_test.go
+++ b/vg/draw/draw_test.go
@@ -5,6 +5,7 @@
 package draw
 
 import (
+	"fmt"
 	"image/color"
 	"reflect"
 	"testing"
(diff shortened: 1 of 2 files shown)


vg/draw/draw_test.go
```diff
@@ -115,3 +116,44 @@ func TestTile(t *testing.T) {
...
+			case err != nil && test.err != nil:
+				if got, want := err.Error(), test.err.Error(); got != want {
+					t.Fatalf("invalid error.\ngot= %v\nwant=%v", got, want)
+				}
+			case err != nil && test.err == nil:
+				t.Fatalf("unexpected error: %+v", err)
+			case err == nil && test.err != nil:
+				t.Fatalf("expected an error (got=%v)", err)
+			case err == nil && test.err == nil:
+				// ok
```
kortschak: The set of cases is exhaustive, so this can be omitted since there's no default to properly condition, no?

sbinet: sure.
I just wanted the `switch` to be thorough (and explicit).
but ok.

sbinet:
PTAL

Issue #589, opened 2020-04-06
plot: allow to specify axis label text position


sbinet:
This was raised in go-hep/hep#613 (item `3)`).

the feature request was to be able to, say, put the X-axis label text at the far right of that axis (not unlike the plot attached in go-hep/hep#612)

(implementing this directly in `go-hep/hplot` would probably mean duplicating a fair bit of `gonum/plot` code for `Axis`)

I'd propose to add a new field to `Axis.Label`:
```go
type Axis struct {
    Label struct {
        Text string
        Pos draw.Position // or Position draw.Position ?
        draw.TextStyle
    }
}
```

with:
```go
package draw

type Position int
const (
    PosCenter Position = 0
    PosLeft   Position = -10 // is this really needed?
    PosRight  Position = +10
    PosBottom Position = -20 // ditto?
    PosTop    Position = +20
)
```

what do you think?

sbinet:
see https://github.com/gonum/plot/pull/595 for what it could look like.

dp1140a:
Would this allow me to draw a Cartesian grid?  As in my axis can be positioned to cross in the middle of my image or anywhere I specify?  Or as some people like to call it a Rectangular orthogonal coordinate plane with axes X and Y on squared grid?
![blank-cartesian-coordinate-system-in-two-dimensions-rectangular-orthogonal-coordinate-plane-with-axes-x-and-y-on-squared-grid-math-scale-template-2J8R36X](https://github.com/user-attachments/assets/fde48e84-2e2c-4863-8e70-9d340a91d349)

If not any tips or examples on how to do so? Or is it even possible?

sbinet:
no, it "just" allows to translate the axis label, not how the axes are laid out:

![image](https://github.com/user-attachments/assets/c2c6dfdc-8580-4bfe-a16e-46d6b8cc80e7)

dp1140a:
looking at some of the code it seems that this should be possible to implement:

[https://github.com/gonum/plot/blob/9b8e196748f157bb608ba15ff0f69b91f76caaf3/plot.go#L168-L169](https://github.com/gonum/plot/blob/9b8e196748f157bb608ba15ff0f69b91f76caaf3/plot.go#L168-L169)

Would need an additional option on Axis that would place it:
For X-Axis:

- Top
- Middle
- Bottom -- default

For Y-Axis:

- Left -- default
- Center
- Right

Thoughts??

PR #595, opened 2020-04-07
plot,vg/draw: add ability to set the axis label position

sbinet:
Fixes gonum/plot#589.

Please take a look.


example_align_test.go
```diff
@@ -90,3 +91,25 @@ func ExampleAlign() {
 		panic(err)
 	}
 }
+
+func ExampleAxisLabels() {
```
golangcibot, line 95: tests: ExampleAxisLabels refers to unknown identifier: AxisLabels (from `govet`)

kortschak:
What do you think about using a `float64` relative offset from the middle of the axis; -1 far left/bottom, 0 middle, 1 far right/top?

sbinet:
Sure can do.

sbinet:
there's probably some amount of weight to the argument of giving the same ability to the `plot.Plot.Title` field...
(that can be part of a follow-up PR).

also: we could actually reuse `draw.XAlignment` and `draw.YAlignment` for the `Position` field; thus:
```go
type Axis struct {
    Label struct {
        XPos draw.XAlignement
        YPos draw.YAlignement
    }
}
```

or add the pair `(XPos, YPos)` directly to `draw.TextStyle`.
it makes sense for `Axis.Label` and `Plot.Title`, but I am not sure it makes total sense for all the places where `draw.TextStyle` is used.
also, re-using `{X,Y}Alignment` for the position could be a bit confusing.

axis.go
```diff
@@ -44,6 +44,9 @@ type Axis struct {
 		// counterclockwise will be added to the label
 		// text before drawing.
 		draw.TextStyle
+
+		// Position is where the axis label string should be drawn.
```
kortschak, line 48: I think this needs documentation explaining that it's relative to the center and that -1 and 1 are the ends.

vg/draw/canvas.go
```diff
@@ -71,6 +71,17 @@ const (
 	YBottom YAlignment = 0
 )
 
+// Position specifies the text position.
+type Position float64
+
+const (
+	PosLeft   Position = -1
```
kortschak, line 78: Do these need to be a new type?

kortschak:
I'd leave off configuring the title until someone asks for it.

sbinet:
somebody already requested it (the title thing) a couple of months back (the simple hack was to add some whitespace).

axis.go
```diff
@@ -46,7 +46,13 @@ type Axis struct {
 		draw.TextStyle
 
 		// Position is where the axis label string should be drawn.
-		Position draw.Position
+		// The default value is draw.PosCenter, displaying the label
+		// at the center of the axis.
+		// Use draw.PosRight to display the label string at the far
+		// right of an X-axis.
+		// Use draw.PosTop to display the label string at the top
+		// of an Y-axis.
```
kortschak, line 54: I don't think these are necessary.

sbinet: adapted to mention the `[-1,+1]` valid interval.

PTAL.

Issue #598, opened 2020-04-08
vg/vgtex: line style is leaking to other scopes


sbinet:
this was reported in https://github.com/go-hep/hep/issues/619 (the item `1)`).
_slightly reworded:_

---
1. Changing the LineStyle of the background grid affect the appearance of the error bar style and the function line too **only with `pdflatex`**. Left: grid with dashed-lines. Right: grid without changing line style (wrt `plotter.NewGrid()`).
![image](https://user-images.githubusercontent.com/30081646/78678051-be611280-78e8-11ea-9f66-8e41b7fa381b.png)
---

this can be seen w/:

```go
func TestLineLatex(t *testing.T) {
	p, err := plot.New()
	if err != nil {
		t.Fatalf("error: %+v", err)
	}
	p.X.Min = -10
	p.X.Max = +10
	p.Y.Min = -10
	p.Y.Max = +10

	f1 := plotter.NewFunction(func(float64) float64 {
		return -7
	})
	f1.LineStyle.Color = color.Black
	f1.LineStyle.Width = 2
	f1.LineStyle.Dashes = []vg.Length{2, 1}

	f2 := plotter.NewFunction(func(float64) float64 {
		return -1
	})
	f2.LineStyle.Color = color.RGBA{R: 255, A: 255}
	f2.LineStyle.Width = 2
	f2.LineStyle.Dashes = []vg.Length{2, 1, 3}

	f3 := plotter.NewFunction(func(float64) float64 {
		return +7
	})
	f3.LineStyle.Color = color.Black
	f3.LineStyle.Width = 2
	f3.LineStyle.Dashes = []vg.Length{2, 1}

	p.Add(f1, f2, f3)
	p.Add(plotter.NewGrid())

	err = p.Save(10*vg.Centimeter, 10*vg.Centimeter, "testdata/linestyles.tex")
	if err != nil {
		t.Fatalf("error: %+v", err)
	}

	err = p.Save(10*vg.Centimeter, 10*vg.Centimeter, "testdata/linestyles.png")
	if err != nil {
		t.Fatalf("error: %+v", err)
	}
}
```
![linestyles](https://user-images.githubusercontent.com/1070920/78806440-f046a780-79c2-11ea-9456-4ceaf3260068.png)
[linestyles.pdf](https://github.com/gonum/plot/files/4451532/linestyles.pdf)

in the PDF, the grid style is the same than `f3`.

PR #599, opened 2020-04-08
vg/vgtex: properly scope line-style modifications

sbinet:
Fixes gonum/plot#598.

Please take a look.


sbinet:
ah, that's probably because `pgf` doesn't handle dash specifications with an odd-number of dashes:
```
\pgfsetdash{list of even length of dimensions}{phase}
Sets the dashing of a line. The first entry in the list specifies the length of the first solid part of the list.
The second entry specifies the length of the following gap. Then comes the length of the second solid
part, following by the length of the second gap, and so on. The `phase` specifies where the first solid
part starts relative to the beginning of the line.
```
our documentation of what is expected for dashes is also a bit lacking, I think:
- https://godoc.org/gonum.org/v1/plot/vg/draw#LineStyle
- https://godoc.org/gonum.org/v1/plot/plotutil#Dashes
(perhaps we should specify/document that a bit better...)

let me fix the test, than I'll merge

Issue #596, opened 2020-04-07
plot: add a Padding field to plot.Axis.Label


sbinet:
this was raised here: https://github.com/go-hep/hep/issues/620 by @rmadar 

---

This one is probably for `gonum/plot`. I would find useful to be able to set some padding between an axis label and its axis, as for `hplot.Title`. One can see on the plot below, that it would gain in clarity - I think. More precisely:
```go
p := hplot.New()
p.Title.Padding = 20        // possible 
p.X.Tick.Label.Padding = 10 // doesn't exist yet?
```

![image](https://user-images.githubusercontent.com/30081646/78680866-50b6e580-78ec-11ea-8e03-37b66c205dba.png)

---

kortschak:
What is the difference between the output of what is proposed and the current behaviour?

sbinet:
see https://github.com/gonum/plot/pull/600 for examples w/ padding=0 (what we have currently, still the default with that PR), padding=5 and padding=10.

kortschak:
OK

PR #600, opened 2020-04-08
plot: add a Padding field to Axis.Label

sbinet:
Fixes gonum/plot#596.

Please take a look.


axis.go
```diff
@@ -39,6 +39,14 @@ type Axis struct {
 		// Text is the axis label string.
 		Text string
 
+		// Padding is the amount of padding
+		// between the top of the label and
+		// the bottom of the X-axis.
+		// Padding is the amount of padding
+		// between the right of the label and
+		// the left of the Y-axis.
```
kortschak, line 47: "Padding is the distance between the label and the axis."

PR #601, opened 2020-04-09
vg/vgtex: handle solid lines through empty pgfsetdash

sbinet:
Please take a look.

### DIFF
--- a/vg/vgtex/canvas.go
+++ b/vg/vgtex/canvas.go
@@ -211,6 +211,7 @@ func (c *Canvas) wstyle() {
 
 func (c *Canvas) wdash() {
 	if len(c.context().dashArray) == 0 {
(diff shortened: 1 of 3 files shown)


kortschak:
What is this for?

sbinet:
it's the "suspenders" part of #599 (being the "belt" part).

#599 made sure no context was bleeding "sidewise" (e.g. Grid wouldn't "inherit" the line style of a (preceding) sibling context).

but it didn't ensure contexts weren't bleeding out (from parent context to child context).
well, it didn't prevent that for one "corner case":
- context-1 sets a dash like `-- - `
- sub-context sets no dash line (ie: solid line).
currently, `vgtext.Canvas.wdash` would return early (as the solid line is represented by a `[]vg.Length{}`, leaving the `pgf` `dash` to what ever it was in the parent context (context-1).

PR #602, opened 2020-04-10
vg/vgtex: implement Canvas.Fill without stroking path

sbinet:
We used to call pgfusepath with both 'stroke' and 'fill' when filling a
path.
This didn't show without transparency/opacity.
But with an opacity-enabled color, it does, as shown in go-hep/hep#618,
where the legend for a filled box is overlined with a not-closed
rectangle.

This CL now calls pgfusepath with only 'fill' as an argument.

Updates go-hep/hep#618.

Please take a look.


sbinet:
PTAL
[before.pdf](https://github.com/gonum/plot/files/4473871/before.pdf)
[after.pdf](https://github.com/gonum/plot/files/4473876/after.pdf)

kortschak:
Thanks

sbinet:
thanks for the review in these trying times.

Issue #603, opened 2020-04-10
plot: legend entries text isn't vertically centered


sbinet:
as mentioned here https://github.com/go-hep/hep/issues/618, the legends' text doesn't seem to be vertically centered:

![img](https://user-images.githubusercontent.com/30081646/78676819-4219ff80-78e7-11ea-8371-319c40cc0ab3.png)

sbinet:
playing a bit with adding a `YPadding` to the `Legend` struct, I get:
![new](https://user-images.githubusercontent.com/1070920/78998930-e9966c80-7b49-11ea-830f-d3306fe86335.png)

compared to (the current behaviour):
![ref](https://user-images.githubusercontent.com/1070920/78998946-f024e400-7b49-11ea-831f-9a4ac0ea1947.png)

but this probably means there is some mistake in how the legend entry height is computed and/or how to place the string within that vertical space.

right now, the legend entry height is used as the total vertical space for an entry:
- https://github.com/gonum/plot/blob/master/legend.go#L98

which is the largest value among all entries:
- https://github.com/gonum/plot/blob/master/legend.go#L155

that is:
- https://github.com/gonum/plot/blob/eddda93a22770fd2b8f43ee2f61e91bc62deed15/vg/draw/canvas.go#L688
```go
func (sty TextStyle) Height(txt string) vg.Length {
	nl := textNLines(txt)
	if nl == 0 {
		return vg.Length(0)
	}
	e := sty.Font.Extents()
	return e.Height*vg.Length(nl-1) + e.Ascent
}
```
so, for a single-line text, it's basically the ascent of the font.

and we start drawing the label at (x,y), where y is the bottom of the baseline:
- https://github.com/gonum/plot/blob/master/legend.go#L117

but, according to:
![apple-font](https://developer.apple.com/library/archive/documentation/TextFonts/Conceptual/CocoaTextArchitecture/Art/glyph_metrics_2x.png)

we should probably start drawing at (x,y) where y is shifted upward by `(Ascent-CapHeight)/2`.
but AFAICT, that information isn't available.
so I used `Descent/2` instead.

see:
![legend_functions](https://user-images.githubusercontent.com/1070920/79000854-b35aec00-7b4d-11ea-85e4-00f4dd2144b5.png)
[legend_functions.pdf](https://github.com/gonum/plot/files/4462106/legend_functions.pdf)

what do you think?
ship it?

PR #604, opened 2020-04-10
plot: offset legend entry by Font.Descent/2 to better vertically center text

sbinet:
Fixes gonum/plot#603.

Please take a look.


legend.go
```diff
@@ -40,6 +40,11 @@ type Legend struct {
 	// final position.
 	XOffs, YOffs vg.Length
 
+	// YPosition specify the vertical position of a legend entry.
```
kortschak, line 43: s/specify/specifies/

legend.go
```diff
@@ -114,6 +120,17 @@ func (l *Legend) Draw(c draw.Canvas) {
...
 		}
 		yoffs := (enth - sty.Rectangle(e.text).Max.Y) / 2
+		switch l.YPosition {
+		case draw.PosBottom:
+			// no-op
+		case draw.PosCenter:
+			delta := -sty.Font.Extents().Descent / 2
+			yoffs += delta
+		case draw.PosTop:
+			// FIXME(sbinet): does this make sense?
```
kortschak, line 130: It does if the whole thing does.

legend.go
```diff
@@ -40,6 +40,11 @@ type Legend struct {
 	// final position.
 	XOffs, YOffs vg.Length
 
+	// YPosition specify the vertical position of a legend entry.
+	// Valid values are [-1,+1], with +1 being the top of the
```
kortschak, line 44: This is not true from what I see in the code. It is in {-1,0,1}, rather than in [-1,+1]. I think it should match the documentation rather than changing the docs.

PR #605, opened 2020-04-14
vg: introduce MultiCanvas, a tee for canvases

sbinet:
Please take a look.

### DIFF
--- a/vg/tee.go
+++ b/vg/tee.go
@@ -24,8 +24,6 @@ type teeCanvas struct {
 // SetLineWidth sets the width of stroked paths.
 // If the width is not positive then stroked lines
 // are not drawn.
-//
-// The initial line width is 1 point.
 func (tee teeCanvas) SetLineWidth(w Length) {
 	for _, c := range tee.cs {
 		c.SetLineWidth(w)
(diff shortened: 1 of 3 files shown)


vg/tee.go
```diff
@@ -0,0 +1,139 @@
...
+
+type teeCanvas struct {
+	cs []Canvas
+}
+
+// SetLineWidth sets the width of stroked paths.
+// If the width is not positive then stroked lines
+// are not drawn.
+//
+// The initial line width is 1 point.
```
kortschak, line 28: For these comments on "The initial ...", I would rather say that the initial state depends on the underlying `Canvas`.

vg/tee.go
```diff
@@ -0,0 +1,139 @@
...
+// The initial color is black.  If SetColor is
+// called with a nil color then black is used.
+func (tee teeCanvas) SetColor(c color.Color) {
+	for _, canvas := range tee.cs {
+		canvas.SetColor(c)
+	}
+}
+
+// Rotate applies a rotation transform to the
+// context.  The parameter is specified in
```
kortschak, line 63: Reduce double space to single.

vg/tee.go
```diff
@@ -0,0 +1,139 @@
...
+
+// DrawImage draws the image, scaled to fit
+// the destination rectangle.
+func (tee teeCanvas) DrawImage(rect Rectangle, img image.Image) {
+	for _, c := range tee.cs {
+		c.DrawImage(rect, img)
+	}
+}
+
+var (
```
kortschak, line 137:
```
var _ Canvas = (*teeCanvas)(nil)
```

vg/tee.go
sbinet: I removed those.

vg/tee.go
sbinet: (also removed from the original `vg.Canvas` documentation)

sbinet:
PTAL

PR #610, opened 2020-05-09
all: introduce vg vect-ops

sbinet:
Please take a look.

### DIFF
--- a/plot.go
+++ b/plot.go
@@ -169,6 +169,7 @@ func (p *Plot) Draw(c draw.Canvas) {
 	}
 
 	p.Legend.Draw(draw.Crop(c, ywidth, 0, xheight, 0))
+	c.Draw()
 }
 
 // DataCanvas returns a new draw.Canvas that
--- a/plot.go
+++ b/plot.go
@@ -446,10 +446,7 @@ func (p *Plot) WriterTo(w, h vg.Length, format string) (io.WriterTo, error) {
 		return nil, err
 	}
 
-	wc := draw.NewWriter(vg.WriterFrom(c))
-	defer wc.Flush()
-
-	p.Draw(draw.NewCanvasWriter(wc, w, h))
+	p.Draw(draw.New(c))
 	return c, nil
 }
 
--- a/plot.go
+++ b/plot.go
@@ -169,7 +169,6 @@ func (p *Plot) Draw(c draw.Canvas) {
 	}
 
 	p.Legend.Draw(draw.Crop(c, ywidth, 0, xheight, 0))
-	c.Draw()
 }
 
 // DataCanvas returns a new draw.Canvas that
@@ -214,7 +213,7 @@ func padX(p *Plot, c draw.Canvas) draw.Canvas {
 	n := (lx*maxx - rx*minx) / (lx - rx)
 	m := ((lx-1)*maxx - rx*minx + minx) / (lx - rx)
 	return draw.Canvas{
-		Canvas: vg.Canvas(c),
+		Canvas: c,
 		Rectangle: vg.Rectangle{
 			Min: vg.Point{X: n, Y: c.Min.Y},
 			Max: vg.Point{X: m, Y: c.Max.Y},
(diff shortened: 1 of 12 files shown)


sbinet:
note I am not happy with the change `draw.Canvas -> *draw.Canvas` in the API.

it's just a corner I've cut, trying not to introduce a new interface.
I think it's possible to leave it as `draw.Canvas` but change its `Canvas` field to a `vgop.Writer` interface (or a concrete type that buffers calls to the underlying `vgop.Writer`, much like `io.Writer` and `bufio.Writer`)

I am also not too happy with the current `ops []vgop.Op` field of `draw.Canvas` that can grow very large (and allocates/reallocates).
another corner I've cut.

but it was to get a first idea of the ball park figures :)

sbinet:
what's the general feeling about this first draft?
do you want me to try to remove the `*draw.Canvas` indirection? (by, probably, modifying the `vg.Canvas` interface to be just what I called `vgops.Writer`)

kortschak:
I may be able to get to this over the weekend.

kortschak:
I've had a look. It looks fairly complex for something that is giving admittedly fairly significant gains but in code that's unlikely to be in a hot path; how many plots does a person need to see?

The change in API is not nice. I agree.There is an obvious way around this, but it is probably worse (make the `ops` field a `*[]vgop.Op` - but this needs now that the `draw.Canvas` must now have a constructor).

I think having a vgop package is wrong. The operations should be a type in vg, and would be `type Op interface { ApplyTo(vg.Canvas) }` which gets rid of the warty `isOp` method and the big switche in `draw.Canvas.Draw`.

What is the `Write` method for? I would call this `Append` if it needs to exist (this also becomes possible with the inversion of the interface architecture).

sbinet:
> It looks fairly complex for something that is giving admittedly fairly significant gains but in code that's unlikely to be in a hot path; how many plots does a person need to see?

well, it depends on the use of gonum/plot.
in a monitoring application (_e.g.:_ monitoring the parameters of a (fairly complex) detector) it can quickly go into the hundreds of plots that will be created and dutifully stored by the automated system (and perhaps looked back at by a human operator later on to assess data quality for a time period.)
in a typical physics analysis, it can clock into the 20-50 plots, when you want to check a bunch of systematics.

admittedly, not all plots will be looked at by a human (some will probably be analyzed by an automated system, in the monitoring use case), but you still want to create them nonetheless.

as for the `vgop` package, it was really just because I surmised it would be easier to lump it back into `vg` rather than the other way around (extracting a subset of a package into a new different one doesn't lend itself neatly into mechanical/automatized work.)

Let me work a bit on the no-plot-draw-api-change aspect.

I believe I can turn this into an opt-in optimization for canvas implementations - much like the `io.WriterTo`/`io.ReaderFrom` optimization in `io.Copy`.
this could also turn `vg.Canvas` (and `vg.Writer` + a new `vg.Reader`) into something much like `io` and all the things around it (`io.Copy`, `io.Tee`, `io.MultiWriter`, ...)

it could also open up other optimizations such as op-coalescing (ie: `vg.Op`s that could be dropped because a few op later it will be undone by another) or vectorization.

kortschak:
>well, it depends on the use of gonum/plot.
in a monitoring application (e.g.: monitoring the parameters of a (fairly complex) detector) it can quickly go into the hundreds of plots that will be created and dutifully stored by the automated system (and perhaps looked back at by a human operator later on to assess data quality for a time period.)
in a typical physics analysis, it can clock into the 20-50 plots, when you want to check a bunch of systematics.

>admittedly, not all plots will be looked at by a human (some will probably be analyzed by an automated system, in the monitoring use case), but you still want to create them nonetheless.

This seems odd. If not all of them are being looked at by a human and they are being checked by an automated system ISTM that the best approach is to leave them in the raw data form and render the plots lazily when either the automated system identifies and issue and wants to show them or even more lazily when that happens and a human wants to see them.

sbinet:
introducing a `draw.Canvas.Canvas` field as `vg.Writer` doesn't improve things.
it's only when introducing it as a concrete buffered `*draw.Writer` that we start to see the original performance improvements.

but then people would have to be careful with the creation of `draw.Canvas` (and not forget to call `Flush`).
retro-fitting this API sounds much more error prone.
I tried a bit just inside the `gonum/plot` repository and I didn't manage to get all the tests back to green.

it was a fun experiment.
(I did notice a couple of low hanging fruits wrt allocations)

Issue #611, opened 2020-05-13
modules: go.mod updated unexpectedly


ShawnROGrady:
### What are you trying to do?
I'm using this library to plot the output of a benchmark run (called `benchplot`, [source code](https://github.com/ShawnROGrady/benchplot)). From within another go project this is run as: `go test ./... -run xxx -bench . | benchplot`

### What did you do?
The bare minimum to reproduce is to create a test module containing only `go.mod`:
```
module github.com/ShawnROGrady/foo

go 1.14

require (
        github.com/ajstarks/svgo v0.0.0-20200320125537-f189e35d30ca // indirect
        github.com/jung-kurt/gofpdf v1.16.2 // indirect
        golang.org/x/image v0.0.0-20200430140353-33d19683fad8 // indirect
        gonum.org/v1/plot v0.7.0
)
```
 `go.sum`, and a basic `main.go`:

```
package main

import (
	"fmt"

	_ "gonum.org/v1/plot"
)

func main() {
	fmt.Println("foo")
}
```

### What did you expect to happen?
I should be able to build the above example (using `go build`) then run it in a different go module without anything being changed. 

### What actually happened?
Running the compiled binary from within a go module results in `go.mod` and `go.sum` being unexpectedly updated in that module. Specifically the following line is added to `go.mod`:
```
gonum.org/v1/plot v0.7.0 // indirect 
```

### What version of Go and Gonum/plot are you using?

```
% go version
go version go1.14.2 darwin/amd64
```

Using go modules, so not seeing anything in `$(go env GOPATH)/src/gonum.org`, but in my `go.mod` I've been able to reproduce with `gonum.org/v1/plot v0.0.0-20200212202559-4d97eda4de95` and `gonum.org/v1/plot v0.0.0-20200212202559-4d97eda4de95` and `gonum.org/v1/plot v0.7.0`

### Does this issue reproduce with the current master?
Yes.

kortschak:
This is expected behaviour, but more significantly this is go tool chain behaviour, and not really anything to do with us. If there is a go.mod file it will have its contents updated to reflect the dependencies the code has.

ShawnROGrady:
@kortschak:

1. `go.mod` isn't empty, either in the module importing this library or the directory where the compiled binary is run. Maybe I should have clarified that when describing my minimal reproduction case
2. the same issue doesn't happen if I recreate that minimal example using the [assert](https://godoc.org/github.com/stretchr/testify/assert) library.

[...]

PR #612, opened 2020-05-13
vg: use x/tools/go/packages to locate vg/fonts dir

sbinet:
Fixes gonum/plot#611.

Please take a look.


go.sum
```diff
@@ -8,14 +9,50 @@ github.com/golang/freetype v0.0.0-20170609003504-e2365dfdc4a0 h1:DACJavvAHhabrF0
...
 golang.org/x/tools v0.0.0-20180525024113-a5b4c53f6e8b/go.mod h1:n7NCudcB/nEzxVGmLbDWY5pfWTLqBcC2KZ6jyYvM4mQ=
+golang.org/x/tools v0.0.0-20190206041539-40960b6deb8e/go.mod h1:n7NCudcB/nEzxVGmLbDWY5pfWTLqBcC2KZ6jyYvM4mQ=
+golang.org/x/tools v0.0.0-20190312151545-0bb0c0a6e846/go.mod h1:LCzVGOaR6xXOjkQ3onu1FJEFr0SW1gC7cKk1uF8kGRs=
+golang.org/x/tools v0.0.0-20191119224855-298f0cb1881e/go.mod h1:b+2E5dAYhXwXZwtnZ6UAqBI28+e2cm9otk0dWdXHAEo=
+golang.org/x/tools v0.0.0-20200513022744-65e69ff2d148 h1:X3wiF+nWEwv11fBoJVQN0dSMu7rlvPkRKYAHuuJoyvc=
+golang.org/x/tools v0.0.0-20200513022744-65e69ff2d148/go.mod h1:EkVYQZoAsY45+roYkvgYkIh4xh/qjgUK9TdY2XT94GE=
+golang.org/x/xerrors v0.0.0-20190717185122-a985d3407aa7/go.mod h1:I/5z698sn9Ka8TeJc9MKroUUfqBBauWjQqLJ2OPfmY0=
+golang.org/x/xerrors v0.0.0-20191011141410-1b5146add898/go.mod h1:I/5z698sn9Ka8TeJc9MKroUUfqBBauWjQqLJ2OPfmY0=
+golang.org/x/xerrors v0.0.0-20191204190536-9bdfabe68543 h1:E7g+9GITq07hpfrRu66IVDexMakfv52eLZ2CXBWiKr4=
+golang.org/x/xerrors v0.0.0-20191204190536-9bdfabe68543/go.mod h1:I/5z698sn9Ka8TeJc9MKroUUfqBBauWjQqLJ2OPfmY0=
```
kortschak, line 49: How come we get all these multiple sum additions?

vg/font_syscall.go
```diff
@@ -7,31 +7,28 @@
...
 
-	if pkg, err := build.Import(importString, "", build.FindOnly); err == nil {
-		p := filepath.Join(pkg.Dir, "fonts")
-		if _, err := os.Stat(p); err == nil {
-			dirs = append(dirs, p)
-		}
+	cfg := &packages.Config{
+		Mode: packages.NeedFiles,
+	}
+	pkgs, err := packages.Load(cfg, fonts)
```
ShawnROGrady, line 29: @sbinet  after messing around a bit I think I understand the issue I was seeing a bit better. This call makes the underlying build tool load the package, which results in a `go.mod` file if called within another module.

I was actually able to get this to work by changing this line to:
```
pkgs, err := packages.Load(cfg, fonts, "mod=readonly")
```

Per the [docs](https://godoc.org/golang.org/x/tools/go/packages):
```
Load passes most patterns directly to the underlying build tool....
```

kortschak: Yes, readonly doesn't mean it won't modify, but that it will fail if it can't modify. See https://github.com/golang/go/issues/26850#issuecomment-411903910

go.sum
sbinet: after a bit of experimenting, I am fairly confident those are coming from `x/tools`.

kortschak: Thanks

Issue #613, opened 2020-05-14
vg: streamline font handling


sbinet:
#611 (and its fix #612 ) uncovered the fact we have a bit of code rot in vg's handling of fonts (and especially locating fonts)

originally, `vg` would load ttf files according to a list of directories (`vg.FontDirs`, optionally seeded from `$VGFONTPATH`) with some default pre-packaged fonts under `vg/fonts` directory.

nowadays, `vg/fonts` doesn't hold any ttf files, just a regular Go package with `[]byte` of the content of ttf fonts.

we should probably revamp or at least consolidate a bit how fonts are handled in gonum/plot.

- [x] drop completely the code that was at fault in #611 
- [ ] expose an API that relies on either `[]byte` as ttf fonts input or just on `freetype/truetype.Font` (or on `x/image/font/sfnt.Font`, when it's ready)
- [ ] consider relying on `x/image/font.Face` (so we can switch between `truetype.Font` and `sfnt.Font` more easily)
- [ ] consider exposing a `fonts.Handler` to gather a set of fonts (_e.g.:_ normal, bold, italics, math). it could be passed to `plot.New` (and thus not require to return `(*Plot, error)` but just `*Plot`)

sbinet:
_wrt_ `fonts.Handler`.
perhaps have a look at what [gioui](https://gioui.org) is doing?
_e.g._ 
- https://godoc.org/gioui.org/text
- https://godoc.org/gioui.org/text#FontRegistry
- https://godoc.org/gioui.org/text#Shaper

PR #614, opened 2020-05-14
vg: remove FontDirs seeding with vg/fonts directory

sbinet:
Updates gonum/plot#613.

Please take a look.


vg/font.go
```diff
@@ -237,12 +237,9 @@ func fontData(name string) ([]byte, error) {
 // If the first font file found is unreadable or cannot be parsed, then
 // subsequent directories are not tried, and the font will fail to load.
 //
-// The default slice contains, in the following order, the values of the
-// environment variable VGFONTPATH if it is defined, then the vg
-// source fonts directory if it is found (i.e., if vg was installed by
-// go get).  If the resulting FontDirs slice is empty then the current
-// directory is added to it.  This slice may be changed to load fonts
-// from different locations.
+// The default slice contains the values of the environment variable
```
kortschak, line 240: I would say that it is initialised with the contents of `VGFONTPATH` is it is defined rather than this; it's exported and mutable.

sbinet:
PTAL

vg/font.go
kortschak: Sorry, that last clause was a justification of my comment. The point was that it may not necessarily contain the values, since because it is exported and mutable, it may have been changed.

```
// The default slice is initialised with the contents of the VGFONTPATH
// environment variable if it is defined.
// This slice may be changed to load fonts from different locations.
```

sbinet: ok. done.

sbinet:
ping?

PR #616, opened 2020-05-22
plot/vg/draw: introduce and use TextHandler

sbinet:
Please take a look.


sbinet:
ping?

kortschak:
This will take some time to look through. I will try to get to it in the next couple of days (task added to personal assistant).

sbinet:
I've dropped the HTML handler out of this PR (to reduce the number of files to review)
If we proceed with this PR (or some version of it), I'll file it back in another one.

(I've also rebased off the latest master, and removed the `panic`s for the `Canvas.DPI` implementations (putting a default/hardcoded 72 value) for canvas implementations that do not expose/use an explicit DPI value)

gob/gob.go
```diff
@@ -43,4 +44,7 @@ func init() {
 	gob.Register(plotter.XYZs{})
 	gob.Register(plotter.XYValues{})
 
+	// vg/draw.TextStyle
+	gob.Register(plot.DefaultTextHandler)
+	gob.Register(text.Plain{})
```
kortschak, line 49: `plot.DefaultTextHandler == text.Plain{}`; why do we register both?

text/latex.go
```diff
@@ -0,0 +1,143 @@
...
+
+import (
+	"fmt"
+	"math"
+
+	"github.com/go-latex/latex/drawtex"
+	"github.com/go-latex/latex/font/ttf"
+	"github.com/go-latex/latex/mtex"
+	"github.com/go-latex/latex/tex"
+	"gonum.org/v1/plot/vg"
```
kortschak, line 15: Insert blank line.

text/latex.go
```diff
@@ -0,0 +1,143 @@
...
+	sty draw.TextStyle
+	pt  vg.Point
+}
+
+func (r *latex) Render(width, height, dpi float64, c *drawtex.Canvas) error {
+	//var (
+	//	w = width * dpi
+	//	h = height * dpi
+	//)
+
```
kortschak, line 91: Delete?

text/latex.go
```diff
@@ -0,0 +1,143 @@
...
+		{X: x1, Y: y1},
+	}
+
+	r.cnv.FillPolygon(r.sty.Color, pts)
+}
+
+var (
+	_ draw.TextHandler = (*Latex)(nil)
+	_ mtex.Renderer    = (*latex)(nil)
+)
```
kortschak, line 143: Can you move this to the top? It should be a documentation of the interface that the type is intended to satisfy.

text/plain.go
```diff
@@ -0,0 +1,16 @@
+// Copyright ©2020 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package text // import "gonum.org/v1/plot/text"
+
+import (
+	"gonum.org/v1/plot/vg/draw"
+)
```
kortschak, line 9: `import "gonum.org/v1/plot/vg/draw"`

text/plain.go
```diff
@@ -0,0 +1,16 @@
...
+
+// Plain is a text/plain handler.
+type Plain = draw.PlainTextHandler
+
+var (
+	_ draw.TextHandler = (*Plain)(nil)
+)
```
kortschak, line 16: `var _ draw.TextHandler = (*Plain)(nil)`

text/text.go
```diff
@@ -0,0 +1,6 @@
+// Copyright ©2020 The Gonum Authors. All rights reserved.
```
kortschak, line 1: I think usually we call these "doc.go".

vg/recorder/recorder.go
```diff
@@ -112,6 +112,11 @@ func (c *Canvas) append(a Action) {
 	c.Actions = append(c.Actions, a)
 }
 
+func (c *Canvas) DPI() float64 {
+	// FIXME(sbinet): add a way to specify a value?
```
kortschak, line 116: If we need to add these we can either make canvases require a constructor which takes a DPI, or add a resolution field that is interpreted as 72 if it's 0, or make it a conditional interface where 72 is the default value if there is no DPI method. Given that LaTeX is the only one that needs it, the last approach feels the nicest to me.

vg/vg.go
```diff
@@ -17,6 +17,8 @@ import (
 // graphics.
 // The origin is in the bottom left corner.
 type Canvas interface {
+	DPI() float64
+
```
kortschak, line 21: Why did we need to add the `DPI` method?

gob/gob.go
sbinet: mainly for "forward coding" (in the perhaps remote case when we'd switch DefaultTextHandler to default to another implementation)

I can remove if you prefer.

vg/vg.go
sbinet: for the text.Latex handler.

vg/recorder/recorder.go
sbinet: I am a bit worried a `xyz.Canvas` implementing this `DPI()` method could be wrapped by a couple layers of `vg.Canvas` (sans the `DPI()` method) interfaces and then presented to a `draw.Canvas` value.
I think the type switch wouldn't be able to dig that `DPI()` method out of the interface layers.

but perhaps it's just a remote, theoretical and uncommon failure mode....

gob/gob.go
kortschak: OK. That's fine.

sbinet:
ping?

kortschak:
I'm still concerned about the `DPI` method.

sbinet:
PTAL: removed DPI from vg.Canvas, set as a field of Latex.

text/latex.go
```diff
@@ -18,15 +18,17 @@ import (
 )
 
 // Latex parses, formats and renders LaTeX.
-type Latex struct{}
+type Latex struct {
+	DPI float64
```
kortschak, line 22: Add a doc comment that this sets that LaTeX font resolution and also that resolution defaults to 72 if this is unset.

sbinet: done

PR #617, opened 2020-05-22
ci: add golangci-lint as a github action

sbinet:
Please take a look.


sbinet:
close in favor of #618

PR #618, opened 2020-05-22
ci: add golangci-lint as a github action

sbinet:
Please take a look.


.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,21 @@
+name: golangci-lint
+on:
+  push:
+    tags:
+      - v*
+    branches:
+      - master
```
kortschak, line 7: Turns out these lines are a bad idea. See gonum#1330.

sbinet: ah, indeed, what I had in go-hep was:
```
on:
  push:
    branches: [ master ]
  pull_request:
    branches: [ master ]
  schedule:
    - cron: '0 2 * * 1-5'
```

sbinet:
PTAL.

I've removed the "check only linting in diff _wrt_ `origin/master`" option.

sbinet:
merging in (I just rebased _wrt_ what you'd approved)

Issue #619, opened 2020-05-22
plotter: consider adding an example for Contour


sbinet:
contours are being generated+tested (to some extend).
but:
- there are no examples,
- there are no reference plots to check against.

PR #620, opened 2020-05-23
plotter: add example for Contour

sbinet:
Fixes gonum/plot#619.

Please take a look.


sbinet:
ok, so the non-naive Contour plot doesn't yield reproductible plots (apologies to come late to that party).
and I didn't manage to use the naive strategy (that leads to reproductible plots) in a sneaky way.

so, just give up on the `cmpimg.CheckPlot` test (but keep the example and the reference file)

kortschak:
I suspect the non-reproducibility comes from the iteration over `conts` in `contourPaths`. I think this can be made deterministic by canonicalising the cycles being returned by Johnson's algorithm (this would be a circular permutation to the minimum node and conditional reversal so that the slice satisfies `s[1] < s[len(s)-1]`).

If you think this is work doing, please file an issue.

sbinet:
I tried to do that earlier with the attached patch. (but forgot the iteration on the `contourSet`.)
to no avail. (even with the iteration on `contourSet` fixed, I think.)

```diff
diff --git a/plotter/contour.go b/plotter/contour.go
index 281929f..631ee06 100644
--- a/plotter/contour.go
+++ b/plotter/contour.go
@@ -316,7 +316,7 @@ func contourPaths(m GridXYZ, levels []float64, trX, trY func(float64) vg.Length)
 	// of another path. This needs to be investigated.
 
 	// Excise loops from crossed paths.
-	for c := range conts {
+	for _, c := range conts.keys() {
 		// Always try to do quick excision in production if possible.
 		c.exciseLoops(conts, true)
 	}
@@ -333,6 +333,17 @@ func contourPaths(m GridXYZ, levels []float64, trX, trY func(float64) vg.Length)
 // contourSet hold a working collection of contours.
 type contourSet map[*contour]struct{}
 
+func (cs contourSet) keys() []*contour {
+	o := make([]*contour, 0, len(cs))
+	for c := range cs {
+		o = append(o, c)
+	}
+	sort.Slice(o, func(i, j int) bool {
+		return o[i].z < o[j].z
+	})
+	return o
+}
+
 // endMap holds a working collection of available ends.
 type endMap map[point]*contour
 
@@ -628,7 +639,7 @@ func (p path) linearPathsIn(g graph) []path {
 			if len(g[u]) > 1 {
 				panic("contour: not a linear path")
 			}
-			for v := range g[u] {
+			for _, v := range g[u].keys() {
 				curr = append(curr, p[u])
 				u = v
 				break
diff --git a/plotter/example_contour_test.go b/plotter/example_contour_test.go
index 760b1ad..fa59078 100644
--- a/plotter/example_contour_test.go
+++ b/plotter/example_contour_test.go
@@ -7,11 +7,13 @@ package plotter_test
 import (
 	"log"
 	"math"
+	"testing"
 
 	"golang.org/x/exp/rand"
 
 	"gonum.org/v1/gonum/mat"
 	"gonum.org/v1/plot"
+	"gonum.org/v1/plot/cmpimg"
 	"gonum.org/v1/plot/palette"
 	"gonum.org/v1/plot/plotter"
 	"gonum.org/v1/plot/vg"
@@ -77,3 +79,7 @@ func (g unitGrid) Y(r int) float64 {
 	}
 	return float64(r)
 }
+
+func TestContour(t *testing.T) {
+	cmpimg.CheckPlot(ExampleContour, t, "contour.png")
+}
diff --git a/plotter/johnson.go b/plotter/johnson.go
index 709f8df..c05270f 100644
--- a/plotter/johnson.go
+++ b/plotter/johnson.go
@@ -4,6 +4,8 @@
 
 package plotter
 
+import "sort"
+
 // johnson implements Johnson's "Finding all the elementary
 // circuits of a directed graph" algorithm. SIAM J. Comput. 4(1):1975.
 //
@@ -44,7 +46,7 @@ func cyclesIn(g graph) [][]int {
 		// s = least vertex in V_k
 		for _, v := range j.adjacent {
 			s := len(j.adjacent)
-			for n := range v {
+			for _, n := range v.keys() {
 				if n < s {
 					s = n
 				}
@@ -75,7 +77,7 @@ func (j *johnson) circuit(v int) bool {
 	j.blocked[v] = true
 
 	//L1:
-	for w := range j.adjacent[v] {
+	for _, w := range j.adjacent[v].keys() {
 		if w == j.s {
 			// Output circuit composed of stack followed by s.
 			r := make([]int, len(j.stack)+1)
@@ -94,7 +96,7 @@ func (j *johnson) circuit(v int) bool {
 	if f {
 		j.unblock(v)
 	} else {
-		for w := range j.adjacent[v] {
+		for _, w := range j.adjacent[v].keys() {
 			j.b[w][v] = struct{}{}
 		}
 	}
@@ -106,7 +108,7 @@ func (j *johnson) circuit(v int) bool {
 // unblock is the UNBLOCK sub-procedure in the paper.
 func (j *johnson) unblock(u int) {
 	j.blocked[u] = false
-	for w := range j.b[u] {
+	for _, w := range j.b[u].keys() {
 		delete(j.b[u], w)
 		if j.blocked[w] {
 			j.unblock(w)
@@ -168,7 +170,7 @@ func (t *tarjan) strongconnect(v int) {
 	t.onStack[v] = true
 
 	// Consider successors of v.
-	for w := range t.g[v] {
+	for _, w := range t.g[v].keys() {
 		if t.indexTable[w] == 0 {
 			// Successor w has not yet been visited; recur on it.
 			t.strongconnect(w)
@@ -237,6 +239,15 @@ func (t *tarjan) sccSubGraph(min int) graph {
 // set is an integer set.
 type set map[int]struct{}
 
+func (s set) keys() []int {
+	ks := make([]int, len(s))
+	for k := range s {
+		ks = append(ks, k)
+	}
+	sort.Ints(ks)
+	return ks
+}
+
 // graph is an edge list representation of a graph.
 type graph []set
 
@@ -256,7 +267,7 @@ func (g graph) subgraph(s int) graph {
 		g[u] = nil
 	}
 	for u, e := range g[s:] {
-		for v := range e {
+		for _, v := range e.keys() {
 			if v < s {
 				delete(g[u+s], v)
 			}
@@ -269,7 +280,7 @@ func (g graph) subgraph(s int) graph {
 func (g graph) clone() graph {
 	c := make(graph, len(g))
 	for u, e := range g {
-		for v := range e {
+		for _, v := range e.keys() {
 			if c[u] == nil {
 				c[u] = make(set)
 			}
```

sbinet:
so... it's a bit sad but I can surely live with it.

kortschak:
I'm happy to look at it (the solution should be simpler than what you have - just a canonicalisation at the end of `johnson` since the order of iteration on `conts` should not matter). I'll forget if there is no issue though.

kortschak:
This should do it.

```diff
diff --git a/plotter/johnson.go b/plotter/johnson.go
index 709f8df..6068395 100644
--- a/plotter/johnson.go
+++ b/plotter/johnson.go
@@ -4,6 +4,8 @@
 
 package plotter
 
+import "sort"
+
 // johnson implements Johnson's "Finding all the elementary
 // circuits of a directed graph" algorithm. SIAM J. Comput. 4(1):1975.
 //
@@ -65,6 +67,9 @@ func cyclesIn(g graph) [][]int {
                j.s++
        }
 
+       // johnson.circuit does range iteration over maps,
+       // so sort to ensure consistent ordering.
+       sort.Sort(byComponentLengthOrStart(j.result))
        return j.result
 }
 
@@ -278,3 +283,11 @@ func (g graph) clone() graph {
        }
        return c
 }
+
+type byComponentLengthOrStart [][]int
+
+func (c byComponentLengthOrStart) Len() int { return len(c) }
+func (c byComponentLengthOrStart) Less(i, j int) bool {
+       return len(c[i]) < len(c[j]) || (len(c[i]) == len(c[j]) && c[i][0] < c[j][0])
+}
+func (c byComponentLengthOrStart) Swap(i, j int) { c[i], c[j] = c[j], c[i] }
diff --git a/plotter/johnson_test.go b/plotter/johnson_test.go
index ff623b9..3154e02 100644
--- a/plotter/johnson_test.go
+++ b/plotter/johnson_test.go
@@ -362,9 +362,6 @@ func TestJohnson(t *testing.T) {
                        g = test.g
                }
                gotCycles := cyclesIn(g)
-               // johnson.circuit does range iteration over maps,
-               // so sort to ensure consistent ordering.
-               sort.Sort(byComponentLengthOrStart(gotCycles))
                if !reflect.DeepEqual(gotCycles, test.wantCycles) {
                        t.Errorf("unexpected johnson result for %d:\n\tgot:%#v\n\twant:%#v", i, gotCycles, test.wantCycles)
                }
@@ -384,11 +381,3 @@ func TestJohnson(t *testing.T) {
                }
        }
 }
-
-type byComponentLengthOrStart [][]int
-
-func (c byComponentLengthOrStart) Len() int { return len(c) }
-func (c byComponentLengthOrStart) Less(i, j int) bool {
-       return len(c[i]) < len(c[j]) || (len(c[i]) == len(c[j]) && c[i][0] < c[j][0])
-}
-func (c byComponentLengthOrStart) Swap(i, j int) { c[i], c[j] = c[j], c[i] }
```

sbinet:
thanks.

turns out, testing w/ your patch applied didn't fix the non-reproducibility.
changing the plot output from `.png` to `.pdf` did, though. (w/o the patch)
changing the plot output from `.png` to `.svg`: again, non reproducibility. (even w/ your patch)

I guess the pdf backend is applying some path optimization of its own?
(diffing the `svg` files leads to 238k of diff file...)

kortschak:
I'll look further.

kortschak:
I've looked further, and it's much more subtle than I remember. The loop excision really depends on a map, but the reference nature of maps makes it much easier to do. I expect it's a fair bit of work to figure this out.

sbinet:
ok. I'll file an issue.

is it ok to leave the example like that? (and defer the `cmpimg.CheckPlot` thing for later?)

kortschak:
I think that's fine. If you want, you could do the test as the PDF with a note explaining why and to change it later.

sbinet:
the PDF is 3x heavier than the PNG.
less churn on the repo to leave as is (and perhaps enable the regression test later)

kortschak:
No worries.

Issue #627, opened 2020-06-11
vg: consider adding a Gio-backed canvas


sbinet:
`x/exp/shiny` seems... not particularly maintained.

but there is a new kid in town: [Gio](https://gioui.org).
and @ajstarks just created a canvas API on top of that:
- https://github.com/ajstarks/giocanvas

(it'd be great to also have some kind of plot interactivity built on top of that...)

- [x] Gio support for affine transforms (translate,rotate,scale)
- [x] Gio support for stroke path
- [x] Gio support for arc path component
- [x] Gio support for dash-lines
- [ ] integrate w/ Gio fonts handling

![img](https://github.com/ajstarks/giocanvas/raw/master/apisheet.png)

kortschak:
Agreed!

sbinet:
Update: Gio is missing a few things just yet (I have added that list to the issue).

in the meantime, one could do the same trick than what I did for `x/exp/shiny`: draw the plot using `vg/vgimg` and display the resulting image in the Gio window.
it wouldn't be as fast as with direct Gio (which IIUC is GPU-accelerated) though.

ajstarks:
FYI, giocanvas has a not-quite working version of arcs: https://github.com/ajstarks/giocanvas/blob/master/giocanvas.go Lines 522-586

sbinet:
here is what it could look like:
- https://github.com/go-hep/exp/pull/9

kevinschoon:
I recently started work on a toy project to do interactive plotting [here](https://github.com/kevinschoon/qviz). Currently it just writes the resulting plot to a file which is monitored by another program, but Gio would be welcomed addition for native support. By _interactive_ I'm referring to the modify->display feedback cycle, not interactivity of the plot itself.

kortschak:
>By *interactive* I'm referring to the modify->display feedback cycle, not interactivity of the plot itself.

Both of these are very valuable.

sbinet:
so, would the submission of a `vg/vggio` (? or `vg/vgio`?) backend using `vg/vgimg` to paint canvas as a (hopefully) temporary work around, be a well received PR?

having used my `x/exp/vgshiny` for some time and now my `x/exp/vggio` for a couple of days, the latter seems much smoother in redrawing things (at least on my machine), even via that hacky scaffolding.

kortschak:
I think so.

sbinet:
since `e7bc1a4` Gioui exposes affine transforms (rotate,scale,...).
[...]

PR #628, opened 2020-06-19
vg/vggio: first import of a Gio-backed canvas

sbinet:
Updates gonum/plot#627.


sbinet:
done.

sbinet:
the lint failure (unable to inspect a dependency that uses `cgo`) is annoying.
it seems related to https://github.com/golangci/golangci-lint/issues/827 that has been closed...

not sure what to do at this point.

vg/vggio/vg.go
```diff
@@ -0,0 +1,106 @@
...
+type Canvas struct {
+	*vgimg.Canvas
+	gtx layout.Context
+}
+
+const (
+	// DefaultDPI is the default dot resolution for image
+	// drawing in dots per inch.
+	DefaultDPI = 96
+)
```
kortschak, line 33:
```
// DefaultDPI is the default dot resolution for image
// drawing in dots per inch.
const DefaultDPI = 96
```

kortschak:
The golangci-lint failure is irritating. I will file an issue to them about that later.

kortschak:
I think I understand this failure; the GitHub action that runs golangci-lint does not have the Cgo dependencies installed (the travis build does, but it's independent). You will need to have pre-install step in a .golangci.yml file that installs those deps, see https://github.com/gonum/gonum/blob/master/.github/workflows/golangci-lint.yml and then adding something like

```
- name: cgo-dependencies
  run: sudo apt install pkg-config libgles2-mesa-dev libxkbcommon-dev libxkbcommon-x11-dev
```

sbinet:
nice catch.

PR #630, opened 2020-06-29
vg/vggio: use layout.Context as hook into Gio

sbinet:
Use `layout.Context` instead of `system.FrameEvent` as the former is what we really use, and is "more general" than the latter.

Please take a look.


sbinet:
travis is reported as "pending" but is actually "green".

PR #631, opened 2020-07-25
plotter/heat: Add RasterHeatMap

davidkleiven:
Please take a look.

This PR is based on the discussion [gonum-dev group](https://groups.google.com/g/gonum-dev/c/TmINmbBZXZc). This PR implements a new heat map type (RasterHeatMap). When plotting a heat map it first creates a raster image and then plots the image. By doing so we avoid creating an artificial grid. Below is an example of a result produced using the new heatmap

demoRaster

I am only familiar with gonum/plot as a user and admittedly I don't know all the internal concepts of the package well. Therefore, I am not sure if I made the most consistent choices. In particular, should there be two heat map types or should there be an option in the current heatmap type to enable/disable raster-based drawing?

Code used to produce the plot

```go
package main

import (
	"fmt"
	"math"

	"gonum.org/v1/plot"
	"gonum.org/v1/plot/palette/moreland"
	"gonum.org/v1/plot/plotter"
)

// Gaussian represents a gaussian sampled on a "virtual" grid
// of dimension NxN.
type Gaussian struct {
	WidthX float64
	WidthY float64
	N      int
}

// X returns the x coordinate corresponding to column c
func (g *Gaussian) X(c int) float64 {
	return float64(c) - float64(g.N/2)
}

// Y returns the y coordinate corresponding to row r
func (g *Gaussian) Y(r int) float64 {
	return float64(r) - float64(g.N/2)
}

// Z returns the value correspondin
func (g *Gaussian) Z(c, r int) float64 {
	exponent := g.X(c)*g.X(c)/(g.WidthX*g.WidthX) + g.Y(r)*g.Y(r)/(g.WidthY*g.WidthY)
	return math.Exp(-exponent)
}

// Dims returns the dimension of the "virtual" grid
func (g *Gaussian) Dims() (int, int) {
	return g.N, g.N
}

func main() {
	data := Gaussian{
		WidthX: 10.0,
		WidthY: 30.0,
		N:      100,
	}

	pal := moreland.Kindlmann().Palette(255)
	plt, _ := plot.New()
	raster := plotter.NewRasterHeatMap(&data, pal)
	plt.Add(raster)
	plt.Save(200, 200, "demoRaster.png")
	fmt.Printf("Picture sucsessfully written\n")
}

```


plotter/heat.go
```diff
@@ -223,3 +224,60 @@ func (h *HeatMap) GlyphBoxes(plt *plot.Plot) []plot.GlyphBox {
...
+
+// Plot implements the Plot method of the plot.Plotter interface.
+func (r *RasterHeatMap) Plot(c draw.Canvas, p *plot.Plot) {
+	cols, rows := r.HeatMap.GridXYZ.Dims()
+	img := image.NewRGBA64(image.Rectangle{
+		Min: image.Point{X: 0, Y: 0},
+		Max: image.Point{X: cols, Y: rows},
+	})
+
+	var pImg *Image
```
sbinet, line 250: I'd remove that line (and use a `:=` declaration+initialization below for the plot image)

plotter/heat.go
```diff
@@ -223,3 +224,60 @@ func (h *HeatMap) GlyphBoxes(plt *plot.Plot) []plot.GlyphBox {
...
+
+			if col != nil {
+				img.Set(i, j, col)
+			}
+		}
+	}
+
+	xmin, xmax, ymin, ymax := r.DataRange()
+	pImg = NewImage(img, xmin, ymin, xmax, ymax)
+	pImg.Plot(c, p)
```
sbinet, line 276:
```suggestion
	plt := NewImage(img, xmin, ymin, xmax, ymax)
	plt.Plot(c, p)
```

plotter/heat.go
davidkleiven: Yes, that's better

sbinet:
Ping @kortschak

plotter/example_rasterheat_test.go
```diff
@@ -0,0 +1,50 @@
+// Copyright ©2015 The Gonum Authors. All rights reserved.
```
kortschak, line 1: 2020

plotter/example_rasterheat_test.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	img := vgimg.New(250, 175)
+	dc := draw.New(img)
+
+	raster := plotter.NewRasterHeatMap(&m, pal)
+	plt.Add(raster)
+	plt.Draw(dc)
+	w, err := os.Create("testdata/rasterHeatMap.png")
+	//w, err := os.Create("/home/davidkleiven/rasterHeatMap.png")
```
kortschak, line 42: Please remove this commented out code.

plotter/heat.go
```diff
@@ -223,3 +224,58 @@ func (h *HeatMap) GlyphBoxes(plt *plot.Plot) []plot.GlyphBox {
...
+	HeatMap *HeatMap
+}
+
+// NewRasterHeatMap creates as new heat map plotter for the given data,
+// using the provided palette.
+func NewRasterHeatMap(g GridXYZ, p palette.Palette) *RasterHeatMap {
+	return &RasterHeatMap{
+		HeatMap: NewHeatMap(g, p),
+	}
+}
```
kortschak, line 240: ISTM that this implementation would be better as an option on `HeatMap`. If we have a `Rasterized bool` field then the implementation can be easily chosen by the user without growing the API surface as significantly as is done here.

plotter/example_rasterheat_test.go
davidkleiven: Fixed

plotter/example_rasterheat_test.go
davidkleiven: Fixed

plotter/heat.go
davidkleiven: I agree with this choice, I will incorporate the changes

davidkleiven:
I added a `Rasterized` flag to the `HeatMap` type and removed the new type.

plotter/example_heat_rasterized_test.go
```diff
@@ -0,0 +1,51 @@
...
+	plt.Draw(dc)
+	w, err := os.Create("testdata/rasterHeatMap.png")
+
+	if err != nil {
+		log.Panic(err)
+	}
+	png := vgimg.PngCanvas{Canvas: img}
+	if _, err = png.WriteTo(w); err != nil {
+		log.Panic(err)
+	}
```
sbinet, line 50: Perhaps simplify a bit the example to just use the `plot.Plot.Save` method?

```suggestion
	raster := plotter.NewHeatMap(&m, pal)
	raster.Rasterized = true
	plt.Add(raster)

	err := plt.Save(5*vg.Centimeter, 3.5*vg.Centimeter, "testdata/rasterHeatMap.png")
	if err != nil {
		log.Panic(err)
	}
```

plotter/heat.go
```diff
@@ -57,6 +58,10 @@ type HeatMap struct {
 	// Min and Max define the dynamic range of the
 	// heat map.
 	Min, Max float64
+
+	// Rasterized is a flag that indicates whether the heatmap
+	// should be produced using raster-based drawing or not
```
sbinet, line 63:
```suggestion
	// Rasterized indicates whether the heatmap
	// should be produced using raster-based drawing or not.
```

plotter/heat.go
```diff
@@ -96,8 +101,43 @@ func NewHeatMap(g GridXYZ, p palette.Palette) *HeatMap {
 	}
 }
 
-// Plot implements the Plot method of the plot.Plotter interface.
-func (h *HeatMap) Plot(c draw.Canvas, plt *plot.Plot) {
+// plotRasterized plots the heatmap using raster-based drawing
```
sbinet, line 104:
```suggestion
// plotRasterized plots the heatmap using raster-based drawing.
```

plotter/heat.go
```diff
@@ -96,8 +101,43 @@ func NewHeatMap(g GridXYZ, p palette.Palette) *HeatMap {
...
+			}
+		}
+	}
+
+	xmin, xmax, ymin, ymax := h.DataRange()
+	pImg := NewImage(img, xmin, ymin, xmax, ymax)
+	pImg.Plot(c, plt)
+}
+
+//  plotVectorized plots the heatmap using vector-based drawing
```
sbinet, line 139:
```suggestion
//  plotVectorized plots the heatmap using vector-based drawing.
```

plotter/heat.go
```diff
@@ -181,6 +221,15 @@ func (h *HeatMap) Plot(c draw.Canvas, plt *plot.Plot) {
...
 
+// Plot implements the Plot method of the plot.Plotter interface.
+func (h *HeatMap) Plot(c draw.Canvas, plt *plot.Plot) {
+	if h.Rasterized {
+		h.plotRasterized(c, plt)
+	} else {
+		h.plotVectorized(c, plt)
+	}
+}
+
```
kortschak, line 232: Callers go above helpers.

plotter/heat.go
kortschak: As @sbinet suggests, but delete the trailing " or not".

davidkleiven:
@sbinet I reverted the dimension of the image from `5*vg.Centimer, 3.5*vg.Centimeter` to the original `250, 175` to make the test pass. Should I have upgraded the image it is compared against instead?

kortschak:
I think this is fine as it is; it reflects the code in the other heat map example.

kortschak:
Thank you for doing this.

sbinet:
thanks as well.

although, I've just tried it on my Go-HEP hplot code and I noticed the Y axis seems to be reversed.
this was actually noticeable in gonum/plot (apologies for not catching this earlier).

here is the golden file for the vector-based golden image:
img-vect

and here is the raster-based one:
img-raster

kortschak:
Ah yes. Graphics Y coordinates run down the screen.

Issue #635, opened 2020-09-02
all: follow xyz_example_test.go naming convention


sbinet:
a former PR named them like `example_xyz_test.go`.
the Go convention is `xys_example_test.go`.

PR #636, opened 2020-09-02
all: follow xyz_example_test naming convention for examples

sbinet:
Fixes gonum/plot#635.

Please take a look.


sbinet:
I propose to tag `plot@v0.8.0` once that's in.

kortschak (approved):

PR #639, opened 2020-09-16
plotter: fix volcano example

### LINKED ISSUE #638: volcano example does not compile
https://github.com/gonum/plot/blob/e9d75d9287226c92c4bf758e99bc992d419251f4/plotter/volcano_example.go#L45

may be fix is to replace
```go
for l := 100.5; l < volcano.Matrix.(*mat.Dense).Max(); l += 5 {
```
with
```go
for l := 100.5; l < mat.Max( volcano.Matrix); l += 5 {
```

kortschak:
Please take a look.

Fixes #638.

### DIFF
--- a/plotter/volcano_example.go
+++ b/plotter/volcano_example.go
@@ -42,7 +42,7 @@ func (g deciGrid) Y(r int) float64 {
 
 func main() {
 	var levels []float64
-	for l := 100.5; l < volcano.Matrix.(*mat.Dense).Max(); l += 5 {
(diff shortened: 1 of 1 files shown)


sbinet (approved):
LGTM.

it'd be great to turn it into a "real" `func Example` to prevent bitrot to occur in the future.

kortschak:
I was thinking the same thing. I will do that tomorrow.

sbinet:
thanks.

"à chaque jour suffit sa peine"
(each day has trouble of its own)

Issue #649, opened 2020-12-14
plot: glyph boxes and text handling (ascent/descent)


sbinet:
it seems our handling of text could be improved.

consider this program:

```go
// +build ignore

package main

import (
	"log"

	"gonum.org/v1/plot"
	"gonum.org/v1/plot/plotter"
	"gonum.org/v1/plot/vg"
	"gonum.org/v1/plot/vg/vgimg"
)

func main() {
	const dpi = vgimg.DefaultDPI

	p, err := plot.New()
	if err != nil {
		log.Fatalf("could not create plot: %+v", err)
	}

	p.Title.Text = "Labels"
	p.X.Min = -1
	p.X.Max = +1
	p.Y.Min = -1
	p.Y.Max = +1

	labels, err := plotter.NewLabels(plotter.XYLabels{
		XYs: []plotter.XY{
			{X: -0.5, Y: +0.5},
			{X: +0.5, Y: +0.5},
			{X: +0.5, Y: -0.5},
			{X: -0.5, Y: -0.5},
		},
		Labels: []string{
			"App0\nApp1", "Bgg0", "Cqq0\nCqq1\nCqq2", "Djj0",
		},
	})
	if err != nil {
		log.Fatalf("could not creates labels plotter: %+v", err)
	}
	for i := range labels.TextStyle {
		labels.TextStyle[i].Font.Size = vg.Length(34)
	}

	p.Add(labels)
	p.Add(plotter.NewGrid())
	p.Add(plotter.NewGlyphBoxes())

	err = p.Save(15*vg.Centimeter, 15*vg.Centimeter, "labels.png")
	if err != nil {
		log.Fatalf("could save plot: %+v", err)
	}
}
```

running it leads to:
![labels](https://user-images.githubusercontent.com/1070920/102065108-6b851680-3df8-11eb-90db-7e294b0157b4.png)

we probably still don't handle descent/ascent well, or at least, not in a coherent fashion.
![glyphterms_2x](https://user-images.githubusercontent.com/1070920/102063368-3b3c7880-3df6-11eb-943d-99d54c07c05a.png)

it seems the glyph box of a text symbol is only considering the bounding box of that symbol.
I'd argue that's not how the glyph box should be defined: as it's used to determine whether some graphic element is "on the plot", it should take into account the descent of the symbol.

it would also seem we don't have a very consistent definition of the baseline (leading to all that head room in those glyphboxes, especially for the multi-line ones)

sbinet:
[...]

PR #650, opened 2020-12-14
all: correctly handle descent in glyphbox of text symbols

sbinet:
Fixes gonum/plot#649.

Please take a look.


plotter/labels_test.go
```diff
@@ -5,12 +5,114 @@
...
+	"gonum.org/v1/plot/vg"
+	"gonum.org/v1/plot/vg/draw"
 )
 
 func TestLabels(t *testing.T) {
 	cmpimg.CheckPlot(ExampleLabels, t, "labels.png")
 	cmpimg.CheckPlot(ExampleLabels_inCanvasCoordinates, t, "labels_cnv_coords.png")
 }
+
+func TestLabelsWithGlyphBoxes(t *testing.T) {
```
kortschak, line 23: It would be good to have some description of the things that this is exercising so that the expectations can be confirmed.

plotter/labels_test.go
```diff
@@ -5,12 +5,114 @@
...
+				XYs: []plotter.XY{
+					{X: -0.8 + 0.00, Y: -0.5},
+					{X: -0.6 + 0.02, Y: -0.5},
+					{X: -0.4 + 0.04, Y: -0.5},
+					{X: -0.8 + 0.00, Y: +0.5},
+					{X: -0.6 + 0.02, Y: +0.5},
+					{X: -0.4 + 0.04, Y: +0.5},
+					{X: +0.0 + 0.00, Y: +0},
+					{X: +0.2 + 0.02, Y: +0},
+					{X: +0.4 + 0.04, Y: +0},
```
kortschak, line 49: Can these have a note explaining why the non-simple form is being used, i.e. what the terms are.

plotter/labels_test.go
```diff
@@ -5,12 +5,114 @@
...
+				},
+				Labels: []string{
+					"Aq", "Aq", "Aq",
+					"Aq\nAq", "Aq\nAq", "Aq\nAq",
+
+					"Bg\nBg\nBg",
+					"Bg\nBg\nBg",
+					"Bg\nBg\nBg",
```
kortschak, line 57: The glyph boxes on 2-high and 3-high labels look tall. Do you know why this is? Is it expected?

sbinet: some amount of that headroom is probably for the descent of a possible line of text, above.
but I agree there is probably too much space.
I tried yesterday to play a bit with various things, to no avail.

there's probably some metric we are not taking into account.

metrics

line gap perhaps isn't subtracted from the glyph box height?

sbinet:
it might actually be a bug in `truetype`.

as part of #613, I am playing with migrating away from `github.com/golang/freetype/truetype` over to `golang.org/x/image/font`

here is what `truetype` (`ttf`) returns for the font's extend, compared with what `x/image/font/sfnt` returns (with `sfnt.HintingNone`):
```
--- FAIL: TestFontExtends (0.01s)
    font_test.go:26: invalid font extents:
        ttf= vg.FontExtents{Ascent:9.814453125, Descent:-3.0322265625, Height:12.8466796875}
        sft= vg.FontExtents{Ascent:8.9111328125, Descent:-2.1630859375, Height:11.4990234375}
```

and with `sfnt.HintingFull`:
```
--- FAIL: TestFontExtends (0.01s)
    font_test.go:26: invalid font extents:
        ttf= vg.FontExtents{Ascent:9.814453125, Descent:-3.0322265625, Height:12.8466796875}
        sft= vg.FontExtents{Ascent:9.0625, Descent:-2.1875, Height:11.5625}
```

(the default of `truetype` seems to be `HintingNone`. I am just posting the `HintingFull` bit for completeness)

here is what I get when using the font metrics from `sfnt`:
labels_glyphboxes
as you can see, the glyphbox "fits" better.
there is probably still an issue, though: the bottom letter (for `YAlign = draw.YBottom`) doesn't land nicely on the requested y coordinate.

the (unscaled) `x/image/font.Metrics` returned by `sfnt` is:
```
font.Metrics{
   Height:2355,
   Ascent:1825,
   Descent:443,
   XHeight:940,
   CapHeight:1341,
   CaretSlope:image.Point{X:0, Y:1},
}
```
where:
```go
type Metrics struct {
	// Height is the recommended amount of vertical space between two lines of
	// text.
	Height fixed.Int26_6

	// Ascent is the distance from the top of a line to its baseline.
	Ascent fixed.Int26_6

	// Descent is the distance from the bottom of a line to its baseline. The
	// value is typically positive, even though a descender goes below the
	// baseline.
	Descent fixed.Int26_6

	// XHeight is the distance from the top of non-ascending lowercase letters
	// to the baseline.
	XHeight fixed.Int26_6

	// CapHeight is the distance from the top of uppercase letters to the
	// baseline.
	CapHeight fixed.Int26_6

	// CaretSlope is the slope of a caret as a vector with the Y axis pointing up.
	// The slope {0, 1} is the vertical caret.
	CaretSlope image.Point
}
```

sbinet:
hacking a bit the Gio library to display the "bounding box" of each text line it draws (in magenta), I get:
labels

from the overlaps of the gio-boxes in the multi-line labels, I'd be enclined to say the `CapHeight` ot the bottom line is the bottom of the line gap.

but applying that formula doesn't pan out.

I am stumped.

kortschak:
What do you want to do with this?

sbinet:
I've asked some help on the Gio mailing list on what's the proper mental model for multi-lines text boxes.

as is, we are already shrinking the glyphbox around a piece of text.
it's not as tight as it could be but it's already an improvement.

I'd go with what we have right now and re-iterate when/if some clever mind on Gio chimes in.
this will incur some churn on the reference files, but oh well...

sbinet:
PTAL

plotter/labels_test.go
```diff
@@ -5,12 +5,119 @@
...
+					{X: -0.6 + 0.02, Y: -0.5}, // Aq + y-align center
+					{X: -0.4 + 0.04, Y: -0.5}, // Aq + y-align top
+
+					{X: -0.8 + 0.00, Y: +0.5}, // ditto for Aq\nAq
+					{X: -0.6 + 0.02, Y: +0.5},
+					{X: -0.4 + 0.04, Y: +0.5},
+
+					{X: +0.0 + 0.00, Y: +0}, // ditto for Bg\nBg\nBg
+					{X: +0.2 + 0.02, Y: +0},
+					{X: +0.4 + 0.04, Y: +0},
```
kortschak, line 54: Can we have above this `NewLabels` call
```
const (
    left   = 0.00
    middle = 0.02
    right  = 0.04
)
```
and use those in place of the literals here and in the following call to `NewLabels`.

kortschak (commented):
After this, LGTM.

sbinet:
FYI, I now get the following `gio` glyphboxes (in magenta):
labels
they snap nicely around the text lines.

(this new image was created with a hacked version of Gio to display those glyphboxes, and after having migrated `gonum/plot` to `x/image/font`.)

sbinet:
(the last squash+rebase+force-push was just to get a nicer git-history, no code change)

PR #652, opened 2020-12-16
vg/vgimg: properly close font.Face after use

sbinet:
Please take a look.

### DIFF
--- a/vg/vgimg/vgimg.go
+++ b/vg/vgimg/vgimg.go
@@ -300,7 +300,10 @@ func (c *Canvas) FillString(font vg.Font, pt vg.Point, str string) {
 	c.ctx.Push()
 	defer c.ctx.Pop()
 
-	c.ctx.SetFontFace(font.FontFace(c.DPI()))
+	face := font.FontFace(c.DPI())
+	defer face.Close()
+
+	c.ctx.SetFontFace(face)
 
 	x := pt.X.Dots(c.DPI())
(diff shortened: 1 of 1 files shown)


vg/vgimg/vgimg.go
```diff
@@ -300,7 +300,10 @@ func (c *Canvas) FillString(font vg.Font, pt vg.Point, str string) {
 	c.ctx.Push()
 	defer c.ctx.Pop()
 
-	c.ctx.SetFontFace(font.FontFace(c.DPI()))
+	face := font.FontFace(c.DPI())
+	defer face.Close()
+
+	c.ctx.SetFontFace(face)
```
kortschak, line 306: Did this come up from linting? The `Close` method on `font.Face` doesn't do anything other than return a [nil `error`](https://github.com/golang/freetype/blob/e2365dfdc4a05e4b8299a783240d4a7d5a65d4e4/truetype/face.go#L250-L251).

kortschak (approved):

vg/vgimg/vgimg.go
sbinet: no, it's just when I went through the modifications for going from `freetype/truetype.Font` to `x/image/font/opentype.Font` that I'd noticed this.

not closing the `font.Face` for these 2 implementations is indeed not too big a deal (right now), but I feel like it's holding that interface wrong.

Issue #608, opened 2020-04-28
vg/vgpdf: jung-kurt/gofpdf is in archive mode


sbinet:
it seems [jung-kurt/gofpdf](https://github.com/jung-kurt/gofpdf) has been archived.

this is a meta-issue to find a new, actively supported, upstream package for generating PDFs.

here is the list of forks:
- https://techgaun.github.io/active-forks/index.html#jung-kurt/gofpdf

PR #654, opened 2020-12-16
Changes PDF handling to use supported libraries.

dlorenc:
The existing PDF functionality was based on two now-unsupported libraries:
rsc.io/pdf and github.com/jung-kurt/gofpdf

This change drops the rsc.io/pdf dependency, and moves the jung-kurt one to
a now maintained fork: github.com/phpdave11/gofpdf

I also had to replace the custom pdf comparison function (built on rsc.io/pdf)
with an existing comparison method in gofpdf. There weren't any tests here, so
I'm not sure how to be super confident the results will be the same.

Please take a look.

### DIFF
--- a/cmpimg/cmpimg.go
+++ b/cmpimg/cmpimg.go
@@ -16,7 +16,7 @@ import (
 	"reflect"
 	"strings"
 
-	"rsc.io/pdf"
+	"github.com/phpdave11/gofpdf"
 
 	_ "image/jpeg"
 	_ "image/png"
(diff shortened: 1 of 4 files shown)


kortschak (approved):
Thank you for doing this. I'll wait for @sbinet. When this is merged, the commit message will be " vg/vgpdf: migrate to phpdave11/gofpdf" (we'll do that in squash).

Please send a PR to gonum/gonum with the commit message "A+C: add Dan Lorenc" adding yourself to CONTRIBUTORS (Google is already in AUTHORS).

sbinet (commented):
LGTM as well.

we should also mention https://github.com/gonum/plot/issues/608 in the commit message.

(even if the fork of jung-kurt/gofpdf is still heavily relying on files (via file names) instead of io.Reader)

dlorenc:
Thanks for the reviews! I'll send that PR to add myself once this is merged.

sbinet:
I believe we actually do this the other way around (or rather, we merge the PRs A+C, then code, once the 2 PRs are ready).

dlorenc:
Oh, sorry! I misunderstood. PR sent.

sbinet:
no worries :) 

needs https://github.com/gonum/gonum/pull/1530

sbinet:
thanks a lot.

PR #655, opened 2020-12-18
vg: use x/image/font/{sfnt,opentype}

sbinet:
Updates gonum/plot#613.

Please take a look.


sbinet:
there's still room for improvement _wrt_ how tight the glyphbox snaps around a piece of text.

but I'll leave this for later (and a revamp of how text is laid out across vg-backends).
(but before that, I'll have to devise my "collection of fonts" plan)

vg/font.go
```diff
@@ -105,16 +111,20 @@ func (f *Font) Name() string {
...
-	return truetype.NewFace(f.font, &truetype.Options{
+	face, err := opentype.NewFace(f.font, &opentype.FaceOptions{
 		Size: f.Size.Points(),
 		DPI:  dpi,
 	})
+	if err != nil {
+		panic(err)
+	}
+	return face
 }
```
kortschak, line 128: We need to consider whether this needs to be altered to return the error. At the moment `opentype.NewFace` only ever returns a nil `error`, but we shouldn't panic here in normal use probably.

vg/font.go
```diff
@@ -149,42 +159,73 @@ type FontExtents struct {
...
+	var (
+		// TODO(sbinet): re-use a Font-level sfnt.Buffer instead?
+		buf  sfnt.Buffer
+		ppem = fixed.Int26_6(f.font.UnitsPerEm())
+	)
+
+	met, err := f.font.Metrics(&buf, ppem, f.hinting)
+	if err != nil {
+		panic(fmt.Errorf("could not extract font extents: %v", err))
+	}
```
kortschak, line 171: The same is true here.

vg/font.go
```diff
@@ -149,42 +159,73 @@ type FontExtents struct {
...
+	scale := f.Size / Points(float64(ppem))
 	return FontExtents{
-		Ascent:  Points(float64(bounds.Max.Y)) * scale,
-		Descent: Points(float64(bounds.Min.Y)) * scale,
-		Height:  Points(float64(bounds.Max.Y-bounds.Min.Y)) * scale,
+		Ascent:  Points(float64(met.Ascent)) * scale,
+		Descent: Points(float64(-met.Descent)) * scale, // sfnt-descent is >0
```
kortschak, line 175: Why do we change sign?

vg/font.go
```diff
@@ -149,42 +159,73 @@ type FontExtents struct {
...
+		Height:  Points(float64(met.Height)) * scale,
 	}
 }
 
 // Width returns width of a string when drawn using the font.
 func (f *Font) Width(s string) Length {
-	// scale converts truetype.FUnit to float64
-	scale := f.Size / Points(float64(f.font.FUnitsPerEm()))
+	var (
+		ppem = fixed.Int26_6(f.font.UnitsPerEm())
```
kortschak, line 183: This name lives long enough that it probably wants a longer name, `pixelsPerEm`?

vg/font_test.go
```diff
@@ -0,0 +1,213 @@
...
+)
+
+func TestFontExtends(t *testing.T) {
+	for _, tc := range []struct {
+		font string
+		want map[Length]FontExtents
+	}{
+		{
+			font: "Times-Roman",
+			want: map[Length]FontExtents{
```
kortschak, line 23: How were the want values obtained?

vg/font_test.go
```diff
@@ -0,0 +1,213 @@
...
+	fnt, err := MakeFont("Times-Roman", 12)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	for _, tc := range []struct {
+		txt  string
+		want Length
+	}{
+		{" ", 3},
```
kortschak, line 129: Same question here.

vg/font_test.go
```diff
@@ -0,0 +1,213 @@
...
+		txt  string
+		want fixed.Int26_6
+	}{
+		{"AV", -264},
+		{"A∇", 0},
+		{"AΔ", 0},
+		{"AA", 0},
+		{"VA", -264},
+		{"∇A", 0},
+		{"∇Δ", 0},
```
kortschak, line 179: Here too; some are surprising for a non-professional-typographer.

kortschak (commented):
I think generally this is an improvement but there are places where things are worse due largely to the differences between above line height rise in single- and multi-line text (for example in plotter/testdata/labels_glyphboxes_golden.png and plotter/testdata/rotation_golden.png among others).

The keys in palette/testdata/reverse_palette_golden.png and plotter/testdata/scatterColor_golden.png are oddly shortened. I don't entirely understand why this is so pronounced.

It improves the separation of the group labels in plotter/testdata/sankeySimple_golden.png.

The legend in testdata/legendAlignment_golden.png is more compressed and looks unbalanced. Again this is related to glyph boxes and the upper height not being as high above the x-height.

I think this also impacts on how high up the page plots are placed (the gap between the top and the title is too small relative to the gap at the bottom - particularly in vg/testdata/width_*).

vg/font.go
sbinet: we change the sign for backward compatibility with `vg.FontExtents` documentation.

I admit I pondered on changing the documentation and propagating the new definition of `Descent` (to follow the one of `x/image/font.Metrics`) but kept it for later.
this could be done in this PR.

vg/font_test.go
sbinet: by extraction from what we had with `truetype`.
I'll add a comment.

vg/font_test.go
sbinet: I'll add a comment (mainly to note where (I think) a Kerning should be had but isn't. probably a shortcoming of the font definition itself rather than the package that decoded it, though)

vg/font.go
kortschak: I think it's worth considering, but later.

kortschak (approved):
There is still the issue of the upper bounding which I think *is* a bug, and one that's important enough to try to fix, but this is good.

sbinet:
as alluded to on the mailing list I had merged this PR with the `ps1` commit message (w/o squashing it appropriately)

I am sorry for this mishap.
should we left it as is?

kortschak:
Yeah, it's fine.

PR #659, opened 2021-01-05
all: define vg.FontExtents.Descent as a positive value to follow x/image/font/sfnt convention

sbinet:
Please take a look.

### DIFF
--- a/axis.go
+++ b/axis.go
@@ -230,7 +230,7 @@ type horizontalAxis struct {
 // size returns the height of the axis.
 func (a horizontalAxis) size() (h vg.Length) {
 	if a.Label.Text != "" { // We assume that the label isn't rotated.
-		h -= a.Label.Font.Extents().Descent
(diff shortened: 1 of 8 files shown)


kortschak (commented):
There is a reference to `Descent` in [vgpdf](https://github.com/gonum/plot/blob/f888cdb60200a6291af8763d83eda6c866a6f53d/vg/vgpdf/vgpdf.go#L185) that looks like it should change too.

sbinet:
indeed, but IIRC, that's not the same `Descent`.
it's the one internal to `gofpdf`.

do you want me to add a comment about that?

kortschak:
Sorry, no. My mistake.

kortschak (approved):

sbinet:
it's finnicky business.

kortschak:
Injudicious use of grep without following up with proper static analysis.

PR #660, opened 2021-01-06
text,vg/draw: extend draw.TextHandler interface

sbinet:
This CL extends the vg/draw.TextHandler interface to be able to handle
how lines are split, which may vary from backend to backend (e.g.: this
differs between text/plain and text/html).
TextHandler's Box method is also modified to follow x/image/font/sfnt
convention to return a positive number for the descent of a glyph
(a.k.a. "depth" in Box lingua.)

This CL also consolidates how multi-lines bounding boxes are computed,
leveraging the underlying draw.TextHandler of a given draw.TextStyle value.

Please take a look.


sbinet:
if somebody understands why rotated labels (both the `y` axis label and the `LaTeX` label) are "funny", I'd be really grateful.

sbinet:
@rmadar: see how it's finally shaping up. (modulo the "funny" business with rotated labels, and the (current) limitations of `go-latex/latex` itself)

text/latex.go
```diff
@@ -26,18 +27,32 @@ type Latex struct {
 
 var _ draw.TextHandler = (*Latex)(nil)
 
-// Box returns the bounding box of the given text where:
+func (hdlr Latex) dpi() float64 {
+	if hdlr.DPI == 0 {
+		return 72
+	}
+	return hdlr.DPI
+}
```
kortschak, line 35: Please put this below its calls.

text/latex.go
```diff
@@ -81,8 +93,26 @@ func (hdlr Latex) Draw(c *draw.Canvas, txt string, sty draw.TextStyle, pt vg.Poi
 		cnv: c,
 		sty: sty,
 		pt:  pt,
+		w:   vg.Length(w / 72 * dpi),
+		h:   vg.Length((h + d) / 72 * dpi),
```
kortschak, line 97: Please give 72 a name.

text/latex_test.go
```diff
@@ -0,0 +1,144 @@
+// Copyright ©2021 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package text_test // import "gonum.org/v1/plot/text"
```
kortschak, line 5: We only have import comments in doc.go files (given we assume modules, we could actually get rid of them all together).

sbinet:
PTAL.

(my comment still applies, though)
https://github.com/gonum/plot/pull/660#issuecomment-755320446

vg/draw/text_plain.go
```diff
@@ -71,21 +67,6 @@ func (hdlr PlainTextHandler) Draw(c *Canvas, txt string, sty TextStyle, pt vg.Po
...
-		if r == '\n' {
-			n++
-		}
-	}
-	return n
-}
-
 var (
 	_ TextHandler = (*PlainTextHandler)(nil)
 )
```
kortschak, line 72: This should be at the top, since it's documentational about intention.

kortschak:
This looks fine. We should wait until someone understands label rotation problem though.

sbinet:
Found it.

PTAL.

text/latex.go
```diff
@@ -49,23 +56,21 @@ func (hdlr Latex) Box(txt string, fnt vg.Font) (width, height, depth vg.Length)
...
 }
 
 // Draw renders the given text with the provided style and position
 // on the canvas.
 func (hdlr Latex) Draw(c *draw.Canvas, txt string, sty draw.TextStyle, pt vg.Point) {
-	dpi := hdlr.DPI
-	if dpi == 0 {
-		dpi = 72
-	}
+	dpi := hdlr.dpi() / latexDPI
```
kortschak, line 66: This isn't needed until L85. Can it be moved down to there?

kortschak (commented):
The padding between the X-axis label and X-axis tick label is very tight (probably Y-axis as well, but this is obscured by the negation sign). A default that's slightly further apart would probably be worthwhile.

text/latex.go
sbinet: done.

sbinet:
yeah, I noticed that as well.

it's coming from the fact that the LaTeX box is much more tightly snapping around glyphs.
perhaps one could artificially insert a linegap or something.

or... would it be ok to file this as an issue to fix for later?

kortschak:
> or... would it be ok to file this as an issue to fix for later?

I think this is OK.

kortschak (approved):

sbinet:
filed https://github.com/gonum/plot/issues/661

Issue #661, opened 2021-01-08
text/latex: tight bounding boxes


sbinet:
as noted in https://github.com/gonum/plot/pull/660#pullrequestreview-564552426, the bounding boxes around text handled by `text.Latex` handler are very close to the glyphs.

this results in, _e.g._, the label for the x or y axes being close to their corresponding titles.

a possible avenue to fix this would be to add `(Ascent-CapHeight)` or `linegap`, as pictured in:
![glyph](https://developer.apple.com/library/archive/documentation/TextFonts/Conceptual/CocoaTextArchitecture/Art/glyph_metrics_2x.png)

kortschak:
I think adding line gap to the bounding box above labels is probably the right thing to do.

PR #662, opened 2021-01-12
text: add a linegap to the LaTeX bounding box

sbinet:
Fixes gonum/plot#661.

Please take a look.


kortschak (approved):
It's still pretty tight, but I think it's OK.

PR #664, opened 2021-01-14
font collection

### LINKED ISSUE #613: vg: streamline font handling

sbinet:
Fixes #613 
Please take a look.


sbinet:
TODO and/or discuss:

- require `plot.New()` to take `text.Handler` and drop the `,error`
- allow `plot.New(nil)` (and default to a plain/text handler with the Liberation fonts collection)
- require `plotter.XXXs` that handle text to take a `text.Handler` as input parameter (and drop `,error` when possible)

there's also the `vg.Font` and `vg.FontMap` stuff to clean-up. (as well as the `vg/fonts.Asset` thing. not everything may be removed as _e.g._ `vgpdf` needs access to some of that to allow embedding fonts...)

text/text.go
```diff
@@ -0,0 +1,188 @@
...
+const (
+	PosLeft   = -1
+	PosBottom = -1
+	PosCenter = 0
+	PosTop    = +1
+	PosRight  = +1
+)
+
+// TextStyle describes what text will look like.
+type TextStyle struct {
```
kortschak, line 72: Is it worth renaming this to `Style` to avoid stutter?

vg/len.go
```diff
@@ -4,37 +4,24 @@
...
-import (
-	"strconv"
-	"strings"
-)
+import "gonum.org/v1/plot/font"
 
 // A Length is a unit-independent representation of length.
 // Internally, the length is stored in postscript points.
-type Length float64
+type Length = font.Length
```
kortschak, line 11: I'm not sure about this direction. I can see philosophical arguments in both directions though. Is the import in the other direction possible?

text/plain.go
```diff
@@ -17,7 +17,7 @@ type Plain struct{}
 var _ Handler = (*Plain)(nil)
 
 // Lines splits a given block of text into separate lines.
-func (hdlr Plain) Lines(txt string) []string {
+func (hdlr *Plain) Lines(txt string) []string {
```
kortschak: Why do this?

kortschak (commented):
This will take me a few iterations since it's not trivial.

text/text.go
sbinet: I did consider it.
but the fact it's widely embedded in many structs in `plot` and `plotter` gave me pause.

I haven't checked how `rsc.io/rf` handles that kind of migration.
(a simple `sed` command could handle that though).

let me check.

kortschak: gopls will do that across package boundaries with the label rename.

vg/len.go
sbinet: I believe that's the only possible import order (if we keep the 2 packages separate).

consider:
```go
package vg

type Canvas interface {
    ...
    FillString(fnt Font, pt Point, txt string)
}

type Font struct {
    Size Length
    ...
}
```

`Font` needs to know about the `Length` type.
`Canvas` needs to know about the `Font` type.

so you can't have `Length` together with `Canvas` and keep `Font` in a different package.
I believe it needs to be:
- a) `(Canvas) -> (Font) -> (Length)`, or
- b) `(Canvas) -> (Font, Length)`, or
- c) `(Canvas, Font, Length)`

for aesthetical reasons, I chose `b)` (and in another thread you seemed to indicate you'd be -like I was- in favor of having a dedicated `font` package)

happy to go either way, though.

text/plain.go
sbinet: this is a remnant of a previous thinking/design for the "font cache" that was designed to be held by value (instead of by pointer-to-value as of right now).
this was made to be sure modifications applied to the font cache (and thru the `text.Plain` handler) were persisted.

I'll revert to a non-ptr-receiver.

sbinet:
> This will take me a few iterations since it's not trivial.

yeah. perhaps I could first do a splinter PR with only the `vg/draw.Text{Handler,Style}` migration?

vg/len.go
kortschak: Yeah. I'm torn. I think it's reasonable the way you have it given that points derive their existence from fonts. I would be good to have someone else's input as well though just to ensure it's not shared delusion.

kortschak:
I think it's broken into parts nicely, so I don't think that's necessary.

text/text.go
sbinet: yes. but I'd like the migration for users of gonum/plot (_ie:_ once the new package version has been tagged) to be as easy as `go fix` (or `rf -ex some-script`)
perhaps the tooling we have right now isn't up to the task. yet.

kortschak: A gofmt rewrite should handle that without issue, or sed.

text/plain.go
sbinet: done.

sbinet: done.

text/text.go
sbinet: done.

sbinet:
ping?

(it'd be great to have this in for the "beginning of the year" Gonum release)

axis.go
```diff
@@ -114,7 +114,7 @@ func makeAxis(o orientation) (Axis, error) {
 		Padding: vg.Points(5),
 		Scale:   LinearScale{},
 	}
-	a.Label.TextStyle = text.TextStyle{
+	a.Label.Style = text.Style{
```
kortschak, line 109: Here I think the field should remain as TextStyle.

axis.go
```diff
@@ -47,7 +47,7 @@ type Axis struct {
 		// For the vertical axis, one quarter turn
 		// counterclockwise will be added to the label
 		// text before drawing.
-		text.TextStyle
+		text.Style
```
kortschak, line 51: Can we keep this as `TextStyle`. This would me stopping it from being embedded, but do we actually use that?

legend.go
```diff
@@ -19,7 +19,7 @@ import (
 type Legend struct {
 	// TextStyle is the style given to the legend
 	// entry texts.
-	text.TextStyle
+	text.Style
```
kortschak, line 23: Same here.

plot.go
```diff
@@ -38,7 +38,7 @@ type Plot struct {
 		// the top of the plot.
 		Padding vg.Length
 
-		text.TextStyle
+		text.Style
```
kortschak, line 46: And here.

text/text.go
```diff
@@ -88,20 +88,20 @@ type TextStyle struct {
 
 // Width returns the width of lines of text
 // when using the given font before any text rotation is applied.
-func (sty TextStyle) Width(txt string) (max vg.Length) {
+func (sty Style) Width(txt string) (max vg.Length) {
```
kortschak, line 100: Single letter receivers?

font/font.go
```diff
@@ -0,0 +1,304 @@
...
+type Variant string
+
+// FontExtents contains font metric information.
+type FontExtents struct {
+	// Ascent is the distance that the text
+	// extends above the baseline.
+	Ascent Length
+
+	// Descent is the distance that the text
+	// extends below the baseline.  The descent
```
kortschak, line 78: Single space after dot.

font/font.go
```diff
@@ -0,0 +1,304 @@
...
+			}
+		}
+		adv, err := f.Face.GlyphAdvance(&buf, idx, pixelsPerEm, hinting)
+		if err != nil {
+			panic(fmt.Errorf("could not retrieve glyph's advance: %v", err))
+		}
+		width += int(adv)
+		prev, hasPrev = idx, true
+	}
+	return Points(float64(width)) * scale
```
kortschak, line 170: `return Points(width) * scale`

font/font.go
```diff
@@ -0,0 +1,304 @@
...
+	faces map[Font]*opentype.Font
+}
+
+func (c *Cache) GobEncode() ([]byte, error) { return nil, nil }
+func (c *Cache) GobDecode([]byte) error {
+	if c.faces == nil {
+		c.faces = make(map[Font]*opentype.Font)
+	}
+	return nil
+}
```
kortschak, line 188: Are these being used?

font/font.go
```diff
@@ -0,0 +1,304 @@
...
+
+	// Weight is the weight of a font, such as Normal or Bold.
+	Weight font.Weight
+
+	// Size is the size of the font.
+	Size Length
+}
+
+// Name returns a fully qualified name for the given font.
+func (f *Font) Name() string {
```
kortschak, line 40: Is this tested?

plot.go
```diff
@@ -486,3 +488,9 @@ func (p *Plot) Save(w, h vg.Length, file string) (err error) {
 	_, err = c.WriteTo(f)
 	return err
 }
+
+func init() {
+	DefaultTextHandler = text.Plain{
+		Fonts: font.NewCache(liberation.Collection()),
+	}
```
kortschak: Any reason to no have this in the `DefaultTextHandler` decl assignment?

gob/gob.go
```diff
@@ -44,7 +44,7 @@ func init() {
 	gob.Register(plotter.XYZs{})
 	gob.Register(plotter.XYValues{})
 
-	// vg/draw.TextStyle
-	gob.Register(plot.DefaultTextHandler)
-	gob.Register(text.Plain{})
+	// text.TextStyle
```
kortschak, line 47: `// text.Style`

vg/vgimg/vgimg_test.go
```diff
@@ -16,12 +16,18 @@ import (
...
+	"gonum.org/v1/plot/font/liberation"
 	"gonum.org/v1/plot/plotter"
 	"gonum.org/v1/plot/vg"
 	"gonum.org/v1/plot/vg/draw"
 	"gonum.org/v1/plot/vg/vgimg"
 )
 
+var (
+	cache = font.NewCache(liberation.Collection())
+)
```
kortschak, line 29: `var cache = font.NewCache(liberation.Collection())`

font/font.go
```diff
@@ -0,0 +1,304 @@
...
+
+// Typeface identifies a particular typeface design.
+// The empty string denotes the default typeface.
+type Typeface string
+
+// Variant denotes a typeface variant, such as "Mono", "Smallcaps" or "Math".
+type Variant string
+
+// FontExtents contains font metric information.
+type FontExtents struct {
```
kortschak, line 72: s/FontExtents/Extents/g?

font/font.go
```diff
@@ -0,0 +1,304 @@
...
+	// is given as a positive value.
+	Descent Length
+
+	// Height is the distance from the lowest
+	// descending point to the highest ascending
+	// point.
+	Height Length
+}
+
+type FontFace struct {
```
kortschak, line 88: s/FontFace/Face/g?

text/text.go
```diff
@@ -0,0 +1,197 @@
...
+	YAlign YAlignment
+
+	// Handler parses and formats text according to a given
+	// dialect (Markdown, LaTeX, plain, ...)
+	// The default is a plain text handler.
+	Handler Handler
+}
+
+// FontExtents returns the FontExtents of this Style's font.
+func (sty Style) FontExtents() font.FontExtents {
```
kortschak, line 94: Though I think this would stay as `FontExtents`; 

```
// FontExtents returns the font.Extents of this Style's font.
func (sty Style) FontExtents() font.Extents
```

font/font.go
```diff
@@ -0,0 +1,304 @@
...
+	"golang.org/x/image/font/sfnt"
+	"golang.org/x/image/math/fixed"
+)
+
+var (
+	// DefaultCache is the global cache for fonts.
+	DefaultCache *Cache = NewCache(nil)
+)
+
+type Font struct {
```
kortschak, line 22: Go doc for `Font`.

font/font.go
```diff
@@ -0,0 +1,304 @@
...
+	"golang.org/x/image/font"
+	"golang.org/x/image/font/opentype"
+	"golang.org/x/image/font/sfnt"
+	"golang.org/x/image/math/fixed"
+)
+
+var (
+	// DefaultCache is the global cache for fonts.
+	DefaultCache *Cache = NewCache(nil)
+)
```
kortschak, line 20:
```
// DefaultCache is the global cache for fonts.
var DefaultCache *Cache = NewCache(nil)
```

font/font.go
sbinet: this is to make `plot/gob` happy.
otherwise, `encoding/gob` will try to (de)serialize `opentype.Font` which doesn't have any exported field (and `gob` doesn't like that.)

I'll add some comment about that, there.

font/font.go
sbinet: `Points` is actually a `func` that takes a single `float64` argument.
so your proposed alternative wouldn't compile.

sbinet:
PTAL.

kortschak (approved):
Thanks

Issue #669, opened 2021-02-16
plot: cannot use fnt.Font() (type *truetype.Font) as type *sfnt.Font in field value


haswalt:
### What are you trying to do?

Use plot to draw a simple line

### What did you do?

This simple example script demonstrates a contrived example based on working code.

```go
import (
	"gonum.org/v1/plot"
	"gonum.org/v1/plot/plotter"
	"gonum.org/v1/plot/vg"
)

func main() {
  p, err := plot.New()
  if err != nil {
	return err
  }

  plotter.DefaultLineStyle.Width = vg.Points(1)

  xys := ...
  l, _ := plotter.NewLine(xys)
  p.Add(l)

  p.HideX()
  p.HideY()

  p.Save(1024, 540, "out.png")
}
```
However when trying to update dependencies within the project (using go modules).

```
go get -u ./...
```

### What did you expect to happen?

All modules to update and the go.mod and go.sum to be updated.

### What actually happened?

This error is output:

```
# gonum.org/v1/plot/text
../../../Go/pkg/mod/gonum.org/v1/plot@v0.8.1/text/latex.go:36:3: cannot use fnt.Font() (type *truetype.Font) as type *sfnt.Font in field value
../../../Go/pkg/mod/gonum.org/v1/plot@v0.8.1/text/latex.go:37:3: cannot use fnt.Font() (type *truetype.Font) as type *sfnt.Font in field value
../../../Go/pkg/mod/gonum.org/v1/plot@v0.8.1/text/latex.go:38:3: cannot use fnt.Font() (type *truetype.Font) as type *sfnt.Font in field value
../../../Go/pkg/mod/gonum.org/v1/plot@v0.8.1/text/latex.go:64:3: cannot use sty.Font.Font() (type *truetype.Font) as type *sfnt.Font in field value
../../../Go/pkg/mod/gonum.org/v1/plot@v0.8.1/text/latex.go:65:3: cannot use sty.Font.Font() (type *truetype.Font) as type *sfnt.Font in field value
../../../Go/pkg/mod/gonum.org/v1/plot@v0.8.1/text/latex.go:66:3: cannot use sty.Font.Font() (type *truetype.Font) as type *sfnt.Font in field value
```

### What version of Go and Gonum/plot are you using?

go version go1.15.8 darwin/amd64
gonum.org/v1/plot v0.8.1

sbinet:
[...]

PR #671, opened 2021-02-23
plot: update go-latex dependency

sbinet:
Updates gonum/plot#669.

Please take a look.


kortschak:
How is this related to #669?

sbinet:
#669 is actually about two things:
- the migration from freetype/truetype.Font to sfnt.Font
- which needs a set of compatible revisions from gonum/plot and go-latex.
the go-latex revision that was encoded in gonum/plot's master was a bit behind.

this removes (or should, at least) the need to `go mod edit replace` go-latex.

kortschak (approved):

PR #688, opened 2021-05-03
Update README.md to fix installation error

centerorbit:
Added a note about needing to install libgles-dev package if you get an error during install.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,4 +36,12 @@ You can get `gonum/plot` using go get:
 
 `go get gonum.org/v1/plot/...`
 
+
+Note: If you get this error during install "Package glesv2 was not found in the pkg-config search path.", install the `libgles-dev` package for your system:
+
+`sudo apt install libgles-dev`
+
+Once complete you may run the `go get` command again.
+
+
 If you write a cool plotter that you think others may be interested in using, please post to the list so that we can link to it in the `gonum/plot` wiki or possibly integrate it into the `plotter` package.


sbinet:
thanks. (it's probably missing a few other Gio dependencies, though)

perhaps we should "guard" `vg/vggio` behind a `// +build vggio` build tag, so `go get gonum.org/v1/plot/...` is kept "pure-Go" ?
@kortschak

kortschak:
Yeah, I'm not sure. That places an obstacle to use of the package that causes its own problems.

We should also decide whether we want to test on travis, since we have minutes back. I'd prefer not to since we're half way through what we have and they are more valuable for testing gonum/gonum on arm64.

centerorbit:
Thanks all. Just for reference, I experienced this on Pop!_OS 20.10

README.md
```diff
@@ -36,4 +36,12 @@ You can get `gonum/plot` using go get:
 
 `go get gonum.org/v1/plot/...`
 
+
+Note: If you get this error during install "Package glesv2 was not found in the pkg-config search path.", install the `libgles-dev` package for your system:
+
+`sudo apt install libgles-dev`
```
sbinet, line 42: perhaps redirect (also? instead?) to Gio's official page for installation?
we should probably rely on their instructions than ours as the latter might sync' off...

ie: https://gioui.org/doc/install

sbinet:
fixed w/ https://github.com/gonum/plot/pull/789

Issue #694, opened 2021-06-01
vg/vgpdf: multiple canvas unecessarily rebuild same fonts


sbinet:
while running a monitoring program that analyzes about 6G of data and produces plots, I noticed half the real time was stemming from producing ~100 plots (PDF).

digging a bit, I noticed a few suboptimal behaviour on the `gofpdf` side:

- https://github.com/phpdave11/gofpdf/pull/15
- https://github.com/phpdave11/gofpdf/pull/16

but also a bigger part coming from `vg/vgpdf` that recreates the pdf fonts (`makeFont`) over and over: one per `vgpdf.Canvas`.
we probably need to create a per-package cache that the `vgpdf.Canvas.fonts` cache would build upon.

PR #695, opened 2021-06-01
vg/vgpdf: add a per-package cache of PDF fonts

sbinet:
This CL adds a package-wide, goroutine-safe cache of fonts data that can be
re-used across vgpdf.Canvas values.

Fixes gonum/plot#694.

Please take a look.

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -9,7 +9,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: cgo-dependencies
-        run: sudo apt install pkg-config libwayland-dev libx11-dev libx11-xcb-dev libxkbcommon-dev libxkbcommon-x11-dev libgles2-mesa-dev libegl1-mesa-dev libffi-dev libxcursor-dev
+        run: |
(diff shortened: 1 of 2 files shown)


sbinet:
I don't understand why `lint` is failing (while trying to install cgo dependencies) whereas the proper `ci` builds don't.

kortschak:
The only difference I can see between the two is that the ci build does an apt update. This should make any difference, but maybe try adding it? I note that the failure seen in the linting is real; I get a 404 trying to hit that from my laptop.

kortschak (dismissed):

kortschak:
Well, disturbingly, that fixed it.

kortschak (approved):

sbinet:
yeah... I've also `s/apt install/apt-get install/`.

`¯\_(ツ)_/¯`

(ah! I can't believe github doesn't render very well "shruggie" in normal mode and I have to go "verbatim"...)

kortschak:
To shrug, you need to `\\` to get a single `\` and `\_` to get a `_`. Complicated? ¯\\\_(ツ)\_/¯ (In matrix, you can just `\shrug` to get it.

PR #699, opened 2021-06-02
font: make Cache goroutine safe

sbinet:
Please take a look.


sbinet:
PTAL
(GitHub was having a few issues today...)

kortschak (approved):

Issue #702, opened 2021-06-10
plot: Changing the default font for plots is unintuitive and/or not possible


stippi2:
### What are you trying to do?

Render a plot to SVG with a certain font as a default for title and axis

### What did you do?

```golang
plot.DefaultFont = font.Font{Typeface: "Arial"}
p := plot.New()
err := plotutil.AddLinePoints(p, ...)
...
err = p.Save(4*vg.Inch, 4*vg.Inch, "plot.svg")
...
```

### What did you expect to happen?

That all labels in the plot use the Arial font.

### What actually happened?

It is not clear at all, what fonts can even be used for `plot.DefaultFont`. Digging through the code reveals there was a built-in map, which is currently [being removed](https://github.com/gonum/plot/issues/696).

It is possible to load a custom font such as Arial, by extending the `vg.FontDirs` global. In addition, the default font cache needs to be manually extended with a mapping for "Arial" to a font object loaded via `vg.MakeFont()`.

However, this will still fail later in the SVG renderer, because it has a hard-coded mapping of typefaces to CSS style strings. This mapping cannot currently be extended. If a font is not part of this mapping, it will at least throw a proper error.

### What version of Go and Gonum/plot are you using?

From go.sum, it looks like I am using v0.9.0.

### Does this issue reproduce with the current master?

No idea.

stippi2:
This is the code I've had to use (on macOS) to get as far as the error in the SVG renderer:

```golang
vg.FontDirs = append(vg.FontDirs, "/System/Library/Fonts/Supplemental/")
vg.FontMap["Arial"] = "Arial"
arial, err := vg.MakeFont("Arial", 12)
if err != nil {
	return fmt.Errorf("failed to load Arial: %w", err)
}

font.DefaultCache.Add([]font.Face{{
	Font: font.Font{Typeface: "Arial"},
	Face: arial.Font(),
}})

plot.DefaultFont = font.Font{Typeface: "Arial"}
p := plot.New()
...
```

sbinet:
thanks for the report.

there are 2 issues in this issue:

- usability/intuitiveness of the new font system compared to the previous one
- ability to use 3rd-party fonts with the SVG backend.

## font

as for the former, I'd say it's a lack of examples (or discoverability of these examples, such as: [this one](https://pkg.go.dev/gonum.org/v1/plot@v0.9.0/vg#example-package-AddFont)) or familiarity with the new API.

as hinted in:
[...]

Issue #703, opened 2021-06-11
vg/vgsvg: consider adding the ability to embed fonts in SVG


sbinet:
as evocated in #702, we could perhaps add an option to `vg/vgsvg.Canvas` (like we did for `vg/vgpdf.Canvas`) to embed fonts inside the SVG document.

the default should probably be to *not* embed fonts.

ex:
```svg
<svg xmlns="http://www.w3.org/2000/svg" width="450" height="150" font-size="24" text-anchor="middle">
    <defs>
        <style>
            @font-face{
                font-family:"Roboto Condensed";
                src:url(data:font/ttf;charset=utf-8;base64,your_base64_encoded_long_string) format("ttf"); 
                font-weight:normal;font-style:normal;
            }
            @font-face{
                font-family:"Open Sans";
                src:url(data:font/ttf;charset=utf-8;base64,your_base64_encoded_long_string) format("ttf"); 
                font-weight:normal;font-style:normal;
            }
            @font-face{
                font-family:"Anonymous Pro";
                src:url(data:font/ttf;charset=utf-8;base64,your_base64_encoded_long_string) format("ttf"); 
                font-weight:normal;font-style:normal;
            }
        </style>
    </defs>
        <text font-family="Roboto Condensed" x="190" y="32.92">
        This is a Roboto Condensed font
    </text>
    <text font-family="Open Sans" x="190" y="82.92">
        This is a Open Sans font
    </text>
    <text font-family="Anonymous Pro" x="190" y="132.92">
        This is a Anonymous Pro font
    </text>
</svg>
```

kortschak:
SGTM

PR #704, opened 2021-06-14
vg/vgsvg: drop fontMap, use informations from font.Font

sbinet:
This CL drops the use of the internal fontMap that was associating some
pre-defined set of fonts with a set of SVG naming schemes (derived from
PostScript).
Instead, use the informations contained in plot/font.Font to derive the
expected SVG font-family (and friends) font style string.

Updates #702.

Please take a look.


sbinet:
PTAL @stippi2

vg/vgsvg/vgsvg.go
```diff
@@ -345,38 +351,87 @@ func (c *Canvas) DrawImage(rect vg.Rectangle, img image.Image) {
...
+	// https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/font-style
+	switch sty {
+	case stdfnt.StyleNormal:
+		return "normal"
+	case stdfnt.StyleItalic:
+		return "italic"
+	case stdfnt.StyleOblique:
+		return "oblique"
+	}
+	panic(fmt.Errorf("vgsvg: invalid font style %+v (v=%d)", sty, int(sty)))
```
kortschak, line 406: default?

vg/vgsvg/vgsvg.go
```diff
@@ -345,38 +351,87 @@ func (c *Canvas) DrawImage(rect vg.Rectangle, img image.Image) {
...
+	case stdfnt.WeightSemiBold:
+		return "600"
+	case stdfnt.WeightBold:
+		return "bold"
+	case stdfnt.WeightExtraBold:
+		return "800"
+	case stdfnt.WeightBlack:
+		return "900"
+	}
+	panic(fmt.Errorf("vgsvg: invalid font weight %+v (v=%d)", w, int(w)))
```
kortschak, line 433: default?

kortschak (commented):
This looks like it has broken text alignment; the width of text in the test golden values is greater and so now the text is not centred and extends beyond the bounds of the plot.

sbinet:
@kortschak what do you see?

here is what I get:
screenshot

could it be that the Liberation fonts are not available to your browser?

kortschak:
Screenshot from 2021-06-15 07-11-45

sbinet:
I get this with my browser:
screenshot

(`master` is left.)

and this in the github rich-diff:
screenshot

kortschak:
It's entirely possible that the installed fonts impacts on this. There should probably be some documentation explaining how to ensure that fonts are available for SVG-rendered plots with this change. We're bound to get issues otherwise.

sbinet:
you're probably right.

perhaps this could be addressed with embedding fonts?
_i.e.:_ https://github.com/gonum/plot/issues/703 and adding the proper documentation+example.

sbinet:
@kortschak 
I have a CL ready to add fonts embedding to SVG canvases.
should I add it to this PR or send another one (once that one is merged.)

kortschak:
Are they independent? If they are, then looking at the PR adding embedding and then rebasing this only that once merged would be the best I think.

sbinet:
they are independent.
I'll send a PR.

sbinet:
actually, it's a bit more intertwined than I remembered. (the coupling is in the naming taxonomy of fonts)
it's #705.

sbinet:
superseded by https://github.com/gonum/plot/pull/705

PR #705, opened 2021-06-18
vg/vgsvg: add switch to embed fonts to SVG plot

### LINKED ISSUE #703: vg/vgsvg: consider adding the ability to embed fonts in SVG

sbinet:
Fixes #703.


vg/vgsvg/vgsvg.go
```diff
@@ -17,8 +17,10 @@ import (
 	"image/png"
 	"io"
 	"math"
+	"strings"
 
 	svgo "github.com/ajstarks/svgo"
+	stdfnt "golang.org/x/image/font"
 
 	"gonum.org/v1/plot/font"
 	"gonum.org/v1/plot/vg"
```
kortschak, line 26: I think somewhere in this file we need documentation about how to ensure that the standard fonts are available on the system if they are not embedded.

sbinet: you mean adding documentation (or a mechanism) to be able to use the `Times/Helvetica/Courier` fonts when embedding isn't enabled?

kortschak: I think just documentation.

sbinet: I've added an example instead.

sbinet:
PTAL.

kortschak (commented):
We still need docs for the case when the fonts aren't embedded, otherwise we are bound to see issues from people complaining about the effect that I show in [this comment](https://github.com/gonum/plot/pull/704#issuecomment-861015532).

vg/vgsvg/vgsvg.go
```diff
@@ -408,6 +408,19 @@ func svgFamilyName(fnt font.Font) string {
 	if fnt.Variant != "" {
 		v += ", " + svgVariantName(fnt.Variant)
 	}
+
+	// add failover to "standard" fonts (Helvetica/Times/Courier)
```
kortschak, line 412: Initial capital and final period.

vg/vgsvg/font_example_test.go
```diff
@@ -0,0 +1,105 @@
...
+	"gonum.org/v1/plot"
+	"gonum.org/v1/plot/font"
+	"gonum.org/v1/plot/plotter"
+	"gonum.org/v1/plot/vg"
+	"gonum.org/v1/plot/vg/draw"
+	"gonum.org/v1/plot/vg/vgsvg"
+)
+
+func Example_embedFonts() {
+	// use Latin-Modern fonts.
```
kortschak, line 25: Initial capital.

kortschak (commented):
Sorry, I missed these, but the golden files need regeneration anyway.

sbinet:
PTAL.

I've added a simple mechanism that should preserve the old behaviour when relying on the `Liberation <==> Times/Helvetica/Courier` association map.
could you confirm?

kortschak:
No, that doesn't fix the issue.

sbinet:
(sorry, comments/PRs crossed each others... PS4 should address your latest comments)

sbinet:
ok. I'll remove the mechanism and add a comment linking to the example, then.

where should I hang it?
on `vgsvg.Canvas`, `vgsvg.New` or at the package level?

kortschak:
I would suggest that there also be documentation for installation of the liberation fonts, both in the package-level docs and in a README for the package.

sbinet:
installing fonts is widely OS (flavour) dependent... (and I actually don't know how to install fonts on darwin nor Windows nor *BSDs...)
I've just redirected to the wikipedia page for now.

(alternatively, we could switch to "embed-fonts=true", like for PDFs: this would make `vgsvg` do the "right thing" by default.)

kortschak:
Yeah, I think just embedding it probably the right thing to do. I just checked my install and I do have them installed, so even then we can't guaranteed the correct behaviour.

sbinet:
PTAL

kortschak (approved):
The fonts in the originally incorrectly rendering fonts are still incorrectly rendering, but this has gone on long enough, so I'll approve.

sbinet:
ok. thanks.

out of curiosity, what's your setup? (browser, OS)

sbinet:
(the force-push is just to lump the `PSxxx` commits together)

kortschak:
ff89.0.1 on ubuntu 18.04.4.

kortschak (approved):

sbinet:
ah!
I can reproduce with ff89 on my archlinux system.

it seems that the culprit is specifying `font-family: Liberation, serif;` or `font-family: Liberation, serif, Times;` instead of, say: `font-family: Liberation, Times;`

let me see if there isn't some deeper flaw at work...

sbinet:
aha!
for firefox (but it also works on my Chromium browser), it should read:
`font-family: Liberation Serif;`
(or all lower case: `font-family: liberation serif;`)
it would otherwise switch to its default font "DejaVu" which obviously doesn't metrically match with Liberation/Times.

I'll fix that.
but perhaps we could then switch back the default to "embed=false" ?

kortschak:
> but perhaps we could then switch back the default to "embed=false" ?

If it work with that, then SGTM

sbinet:
PTAL.

now `svgFontDescr` will generate the correct `font-family` name directly from the underlying raw SFNT data (using `sfnt.NameIDFamily`) and handle the mismatch of meaning for `font-variant` between SVG and gonum/plot/font.

sbinet:
PTAL (just rebased off the new `master`)

kortschak (approved):
:tada:

PR #711, opened 2021-06-30
plotter: fix handling of x/y-offsets in Labels glyph boxes

### LINKED ISSUE #710: plotter: Labels x/y-offsets are in vg.Length and it's inconvenient
while working on the glyphboxes issue, I've noticed:

```go
package plotter // import "gonum.org/v1/plot/plotter"

// Labels implements the Plotter interface,
// drawing a set of labels at specified points.
type Labels struct {
        XYs

        // Labels is the set of labels corresponding
        // to each point.
        Labels []string

        // TextStyle is the style of the label text. Each label
        // can have a different text style.
        TextStyle []text.Style

        // XOffset and YOffset are added directly to the final
        // label X and Y location respectively.
        XOffset, YOffset vg.Length
}
```

the `{X,Y}Offset` fields are in `vg.Length` (so, centimeters and what not).
it's convenient when drawing the labels:

```go
// Plot implements the Plotter interface, drawing labels.
func (l *Labels) Plot(c draw.Canvas, p *plot.Plot) {
        trX, trY := p.Transforms(&c)
        for i, label := range l.Labels {
                pt := vg.Point{X: trX(l.XYs[i].X), Y: trY(l.XYs[i].Y)}
                if !c.Contains(pt) {
                        continue
                }
                pt.X += l.XOffset
                pt.Y += l.YOffset
                c.FillText(l.TextStyle[i], pt, label)
        }
}
```

but rather inconvenient when trying to compute the glyphbox around each label:
```go
// GlyphBoxes returns a slice of GlyphBoxes,
// one for each of the labels, implementing the
// plot.GlyphBoxer interface.
func (l *Labels) GlyphBoxes(p *plot.Plot) []plot.GlyphBox {
        bs := make([]plot.GlyphBox, len(l.Labels))
        for i, label := range l.Labels {
                pt := l.XYs[i]
                pt.X += l.XOffset // error (cannot add float64 and vg.Length)
                pt.Y += l.YOffset // ditto
                bs[i].X = p.X.Norm(pt.X)
                bs[i].Y = p.Y.Norm(pt.Y)
                sty := l.TextStyle[i]
                bs[i].Rectangle = sty.Rectangle(label)
        }
        return bs
}
```
(the original code of `Labels.GlyphBoxes(...)` was missing the x/y-offsets)

there's no way (at least, I couldn't find one) to go from `vg` units back to "canvas/user" units.
(even less so when one has only a `plot.Plot` in hands, w/o the corresponding `draw.Canvas` as one has in `Labels.Plot(c,p)`)

I see two avenues to solve this:

- change `{X,Y}Offset` to data units (and document)
- modify the `GlyphBoxes` signature (and the according `plot.GlyphBoxer` interface) to take an additional `draw.Canvas` argument, add a way (to `plot.Plot`?) to go from `vg.Length` units back to "data units".

option a) is probably less wide-ranging a change (only a couple of plotters in `gonum/plot` are using these fields)

(in any event, adding a method to go from `vg.Length` units back to "data units" may very well be quite convenient, though. I had the use for such a thing for `go-hep/hplot` [here](https://github.com/go-hep/hep/blob/4b45b719da89818515cf72823748522e4bd57683/hplot/label.go#L26) when trying to set labels in "normalized" coordinates.)

thoughts?

sbinet:
This CL fixes the handling of the X/Y-offsets of plotter.Labels: they
were previously ignored in the glyph box computation.
This CL also adds a test that exercizes offsetted labels and their glyph box.

Fixes #710.

Please take a look.


sbinet:
this is a spinoff of #708 and #710.
(I'll rebase #708 off that one once/when/if it's merged)

sbinet:
note that we're still not out of the woods: the `Aq` and `Bg` labels in red should align correctly on the (resp.) blue and green horizontal lines (they should be the baselines.)

plotter/labels.go
```diff
@@ -96,11 +96,16 @@ func (l *Labels) DataRange() (xmin, xmax, ymin, ymax float64) {
 // plot.GlyphBoxer interface.
 func (l *Labels) GlyphBoxes(p *plot.Plot) []plot.GlyphBox {
 	bs := make([]plot.GlyphBox, len(l.Labels))
+	offset := vg.Point{X: l.XOffset, Y: l.YOffset}
```
kortschak, line 99: This sort of suggests that `Labels.[XY]Offset` should have been a `vg.Point`. Maybe later?

plotter/labels.go
```diff
@@ -96,11 +96,16 @@ func (l *Labels) DataRange() (xmin, xmax, ymin, ymax float64) {
...
 	for i, label := range l.Labels {
-		bs[i].X = p.X.Norm(l.XYs[i].X)
-		bs[i].Y = p.Y.Norm(l.XYs[i].Y)
-		sty := l.TextStyle[i]
-		bs[i].Rectangle = sty.Rectangle(label)
+		rect := l.TextStyle[i].Rectangle(label)
```
kortschak, line 101: Do we want to add a `Rectangle.Add(vg.Point)`? There is precedent in image.

plotter/labels_test.go
```diff
@@ -102,6 +102,46 @@ func TestLabelsWithGlyphBoxes(t *testing.T) {
 				sty.YAlign = draw.YBottom
 			}
 
+			// label with positive x/y-offsets
+			loffp, err := plotter.NewLabels(plotter.XYLabels{
+				XYs: []plotter.XY{
+					{X: +0.0 + left, Y: +0},
```
kortschak, line 108: Are the unary `+`s and decimal points necessary?

plotter/labels_test.go
```diff
@@ -102,6 +102,46 @@ func TestLabelsWithGlyphBoxes(t *testing.T) {
...
+			if err != nil {
+				t.Fatalf("could not creates labels plotter: %+v", err)
+			}
+			for i := range loffp.TextStyle {
+				sty := &loffp.TextStyle[i]
+				sty.Font.Size = vg.Length(fontSize)
+				sty.Color = color.RGBA{G: 255, A: 255}
+			}
+			loffp.XOffset = vg.Points(75)
+			loffp.YOffset = vg.Points(75)
```
kortschak, line 123: These conversions are not necessary (and below).

plotter/labels.go
sbinet: sure.
I'll try to insert this before this commit and rebase.
propagating to image will probably another PR.

plotter/labels.go
sbinet: I'll see if I can swizzle that in front of this commit.

plotter/labels.go
kortschak: Sorry, I meant that stdlib's `image.Rectangle` has an `Add` method.

plotter/labels.go
```diff
@@ -80,8 +79,7 @@ func (l *Labels) Plot(c draw.Canvas, p *plot.Plot) {
 		if !c.Contains(pt) {
 			continue
 		}
-		pt.X += l.XOffset
-		pt.Y += l.YOffset
+		pt = pt.Add(l.Offset)
```
sbinet, line 82: not sure whether we shouldn't move the offset fixup *before* the `c.Contains(pt)` condition...

sbinet:
PTAL

plotter/labels.go
```diff
@@ -97,10 +95,11 @@ func (l *Labels) DataRange() (xmin, xmax, ymin, ymax float64) {
...
-		sty := l.TextStyle[i]
-		bs[i].Rectangle = sty.Rectangle(label)
+		bs[i] = plot.GlyphBox{
+			X:         p.X.Norm(l.XYs[i].X),
+			Y:         p.Y.Norm(l.XYs[i].Y),
+			Rectangle: l.TextStyle[i].Rectangle(label).Add(l.Offset),
+		}
 	}
 	return bs
 }
```
kortschak, line 105: This is really nice now.

kortschak (commented):
One unresolved nit in labels_test.go.

plotter/labels.go
sbinet: you made it happen :)

kortschak: Team work; I wouldn't have looked without prompting.

sbinet:
sorry, fixed `loffp` and not the other.

PTAL.
(rebased+merged preemptively)

kortschak (approved):

PR #718, opened 2021-08-03
cmpimg: use rsc.io/pdf to properly compare PDFs

### LINKED ISSUE #717: cmpimg: invalid PDF comparison
`cmpimg` fails to provide a meaningful comparison for PDF documents:

```go
func cmpPdf(pdf1, pdf2 []byte) bool {
	return fpdf.CompareBytes(pdf1, pdf1, false) == nil
}
```
(it compares `pdf1` to itself)

unfortunately, `fpdf.CompareBytes` just compares bytes (it's a convoluted `bytes.Equal`) so all tests fails when fixing the above bug.

I'd vote for going back to `rsc.io/pdf` to perform the comparison.
yes, it's archived, but the PDF standard won't change anytime soon.
and if we find a bug in `rsc.io/pdf`, we can fork.

(I haven't looked for a replacement of `rsc.io/pdf`.)

sbinet:
Fixes #717.

Also improve test coverage of EqualApprox.


cmpimg/cmpimg_test.go
```diff
@@ -101,3 +103,155 @@ func TestEqualApprox(t *testing.T) {
...
+			img1:  []byte("<svg></svg>"),
+			img2:  []byte("<svg></svg>"),
+			delta: -10,
+			want:  true,
+		},
+		{
+			name:  "svg-diff",
+			img1:  []byte("<svg></svg>"),
+			img2:  []byte("<svg>1</svg>"),
+			delta: +10,
```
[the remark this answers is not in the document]
sbinet: to test for clipping:
https://github.com/gonum/plot/blob/30a9c6daafeb5e82f28a89752e5f2139f33fdcde/cmpimg/cmpimg.go#L44..L49

cmpimg/cmpimg_test.go
```diff
@@ -101,3 +103,155 @@ func TestEqualApprox(t *testing.T) {
...
+			name: "pdf-ok",
+			img1: read("../vg/vgpdf/testdata/arc_golden.pdf"),
+			img2: read("../vg/vgpdf/testdata/arc_golden.pdf"),
+			want: true,
+		},
+		{
+			name:  "pdf-diff",
+			img1:  read("../vg/vgpdf/testdata/arc_golden.pdf"),
+			img2:  read("../vg/vgpdf/testdata/issue540_golden.pdf"),
+			delta: +10,
```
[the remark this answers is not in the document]
sbinet: I'll add some documentation.

sbinet:
arf... GHA is finicky and seems to randomly fail to initialize correctly xvfb...
```
=== RUN   TestCanvas
vggio_test.go:84: could not create screenshot: vggio: could not create headless window: eglInitialize failed: 0x3001
```

sbinet:
PTAL.

vg/vggio/vggio_test.go
```diff
@@ -24,6 +26,45 @@ import (
 
 const deltaGio = 0.05 // empirical value from experimentation.
 
+func init() {
```
kortschak, line 29: This is to check that vggio works without timing out correct? If that's the case, I'd prefer that it be in a real test; it can be made a subtest that the other tests are also in to ensure that it runs before them.

Is there any clue why this fails on GHA?

vg/vggio/vggio_test.go
```diff
@@ -24,6 +26,45 @@ import (
...
+		}
+		_, err := New(gtx, w, h, UseDPI(dpi)).Screenshot()
+		return err
+	}
+
+	tck := time.NewTimer(10 * time.Second)
+	defer tck.Stop()
+
+	var err error
+	for {
```
kortschak, line 55: Why is the loop necesary?

vg/vggio/vggio_test.go
```diff
@@ -24,6 +26,45 @@ import (
...
+		select {
+		case <-tck.C:
+			panic(fmt.Errorf("vg/vggio_test timeout: %+v", err))
+		default:
+			err = test()
+			if err == nil {
```
kortschak, line 61: `err := test()`

vg/vggio/vggio_test.go
```diff
@@ -24,6 +26,45 @@ import (
...
+			Constraints: layout.Exact(image.Pt(
+				int(w.Dots(dpi)),
+				int(h.Dots(dpi)),
+			)),
+		}
+		_, err := New(gtx, w, h, UseDPI(dpi)).Screenshot()
+		return err
+	}
+
+	tck := time.NewTimer(10 * time.Second)
```
kortschak, line 51: s/tck/timer/g

vg/vggio/vggio_test.go
sbinet: my working hypothesis was first that Xvfb needed some time to boot up correctly (so I added a `sleep 30s` in the corresponding yaml file), but that wasn't it.

my current working hypothesis is that Xvfb sets up the headless display "on demand" and there might be some race or delay between its availability and the first time it is requested.

thus the `init` function which would need to be executed *before* all other `TestXYZ` funcs.

I guess one could simplify that `init` func and just execute `test` twice and return.

vg/vggio/vggio_test.go
sbinet: I'll just run `test` twice.

sbinet:
PTAL

vg/vggio/vggio_test.go
kortschak: Does this work? It has less boiler plate.
```
func init() {
	if runtime.GOOS != "linux" {
		return
	}

	const (
		w   = 20 * vg.Centimeter
		h   = 15 * vg.Centimeter
		dpi = 96
	)
	gtx := layout.Context{
		Ops: new(op.Ops),
		Constraints: layout.Exact(image.Pt(
			int(w.Dots(dpi)),
			int(h.Dots(dpi)),
		)),
	}

	for try := 0; try < 2; try++ {
		_, err := New(gtx, w, h, UseDPI(dpi)).Screenshot()
		if err == nil {
			return
		}
	}
	panic(fmt.Errorf("vg/vggio_test: could not setup headless display: %+v", err))
}
```

sbinet: it does (with a minimal fix for the dangling `err` variable)

kortschak (approved):
Thanks

PR #719, opened 2021-08-03
ci: enable GHA+Windows

sbinet:
Please take a look.


kortschak:
It looks like the issues that you were facing are related to this code that we no longer run, https://github.com/gonum/gonum/blob/89d1544a09ed861d5a2b1745fde6a95c6b45b545/.travis.yml#L68-L83.

sbinet:
yeah, thanks.
I couldn't find it back but ended up "rediscovering it".
although it didn't work: the `mv` could work because the directory being moved (`gonum.org/v1/gonum`) was used by another process.

oh well, hopefully https://github.com/actions/checkout/issues/135#issuecomment-890252324 will be addressed at some point (ideally, *before* we are fed up with AppVeyor)

PR #730, opened 2021-08-17
Gha windows

sbinet:
Please take a look.


sbinet:
it's really stranged.
I applied the same modification on Go-HEP: https://github.com/go-hep/hep/pull/866
it worked there.
but not here.

no clue as to why.

PR #738, opened 2021-09-26
ci: make cache works by removing comments

ylz-at:
The path is a multiline string, which includes comments in its value. This
means that the caching was not working at all, because it would include
the comment in the path. Here I'm moving the comments, which fixes the
caching so it works.


kortschak:
Thank you.

kortschak (dismissed):
LGTM but waiting for A+C PR at gonum/gonum

kortschak (approved):
Thanks

PR #741, opened 2022-01-18
all: bump to git.sr.ht/~sbinet/gg@v0.3.1

### LINKED ISSUE #559: Strange NewLine behaviour
I am trying to plot some data. Mostly "0" and "1". I am doing so with more than 200k values. 

### What did you do?
The figure didn't make sense so I just changed from NewLine to NewScatter.

To have similar figures.

### What actually happened?
The NewLine figure doesn't makes sense to me.

Go version `go version go1.12.4 windows/amd64`.

Gonum plot version: `9aa86143727f898329b04f9a16882237caa96ddc`

The only relevant change in the code that I made:
```go
gen, _ := plotter.NewScatter(pts)
//gen, _ := plotter.NewLine(pts)
```

I am sorry that I cannot attach the data itself, but it is just a binary signal which changes from 0 to 1.

The NewScatter figure (OK):

scatter

The NewLine figure (NOK):

line

### LINKED ISSUE #645: plot: saved file missing data points
Save plot as PNG

### What did you do?

I created a simple line plot.

The is was I would expect the PNG to show. But it only works in an SVG or PDF format.
image

### What actually happened?

When I try to output the file as PNG, it doesn't draw the full dataset
image

go version go1.14.4 windows/amd64

gonum.org/v1/plot v0.0.0-20190515093506-e2840ee46a6b/go.mod h1:Wt8AAjI+ypCyYX3nZBvf6cAIx93T+c/OS2HFAYskSZc=
gonum.org/v1/plot v0.8.1 h1:1oWyfw7tIDDtKb+t+SbR9RFruMmNJlsKiZUolHdys2I=
gonum.org/v1/plot v0.8.1/go.mod h1:3GH8dTfoceRTELDnv+4HNwbvM/eMfdDUGHFG2bo3NeE=

haven't tried

### LINKED ISSUE #687: vg/vgimg: fails to draw horizontal lines with many line segments

Plot a long data series.

### What did you do?

This won't run in playground due to timeout, but change the playground constant to false and it will run locally.

https://play.golang.org/p/gSINhzskCHi

Plot something equivalent to this (svg output).
Screenshot from 2021-04-21 09-24-46

### What actually happened?

bug

The plot is correct with svg, eps and pdf formats, but missing the horizontal segment with png, tiff and jpg, so it appears to be due to vg/vgimg. The `*Canvas.Stroke` method is a thin wrapper around `*"github.com/fogleman/gg".Context.Stroke`, so the error may there. Note also that the effect is not noticeably visible for `len(data)` values lower than ~4000 in the bug.go reproducer.

v0.9.0 caab7788e04dfd7a74cec4ee80f55b7d1f9a2bfa

Yes.

sbinet:
Please take a look.

Fixes #687.
Fixes #645.
Fixes #559.


kortschak:
Looks like some module clean up needs to happen.

sbinet:
it's probably coming from `x/image` itself.

sbinet:
PTAL

kortschak:
Do we have a test case for the bug that this fixes?

sbinet:
in `~sbinet/gg`, yes:
- https://git.sr.ht/~sbinet/gg/commit/54cd41f7498eddd8bd578a51d0ce6cbdd693a9aa

but you're right that one of the referenced `gonum/plot` issues could be turned into a proper test.
I'll do that tomorrow (Paris' time)

kortschak:
Thanks. In that commit, I still don't understand the need for the slackness in the pixel distance; I was unable to get any of the test suite of gg to fail by changing it to `dx+dy != 0` and the commit that introduced it made no comment about why it was chosen.

sbinet:
PTAL.

as for your https://github.com/gonum/plot/pull/741#issuecomment-1015894185, I haven't put too much intelligence in `~sbinet/gg` yet (?), aiming for just "same codebase, same bugs, save for the one at hand".

there's probably some amount of low hanging fruits to address in this new repo
(_e.g._ reducing the API surface, reusing [go-text](https://github.com/go-text) stuff, ...)

kortschak (approved):

PR #744, opened 2022-03-16
all: bump gonum/gonum@v0.11.0

sbinet:

Please take a look.


sbinet:
hum... @kortschak now it seems the builds are stuck :)

kortschak:
Also... note that it's not crucial to get this change in since MVS doesn't require it as plot doesn't make use of any v0.11 feature and users will pull what they need.

sbinet:
github was probably having a case of "Monday on Wednesday"...

PR #745, opened 2022-05-10
Staticcheck

sbinet:
Please take a look.


sbinet:
@kortschak PTAL.

(could you also remove the `lint` check as being required?)

.github/workflows/ci.yml
```diff
@@ -92,6 +93,14 @@ jobs:
       run: |
         go test -v ./...
 
+    - name: static-check
```
kortschak, line 96: I'd like to have this earlier, or potentially running in parallel in it's own job.

If you rename it to be "lint" to match the golangci-lint job, then the branch protection rule doesn't need to be updated.

sbinet:
PTAL

kortschak (approved):
:tada:

PR #748, opened 2022-07-14
Fix off by one indexing bug in plotutil.MedianAndMinMax

skiesel:
Fix off by one indexing bug in plotutil.MedianAndMinMax when determining the mean of the two inner-most elements in the slice.

Please take a look.


kortschak:
Thank you. Would you please open a PR at gonum/gonum with the commit message "A+C: add Scott Kiesel" adding your details to the AUTHORS and CONTRIBUTORS files. When you have done that I'll approve and merge.

kortschak (approved):

PR #751, opened 2022-08-17
all: bump to Go-1.19

sbinet:
Please take a look.

### DIFF
--- a/plotter/boxplot.go
+++ b/plotter/boxplot.go
@@ -87,7 +87,7 @@ type BoxPlot struct {
 // NewBoxPlot returns a new BoxPlot that represents
 // the distribution of the given values.  The style of
 // the box plot is that used for Tukey's schematic
-// plots is ``Exploratory Data Analysis.''
+// plots in “Exploratory Data Analysis.”
 //
 // An error is returned if the boxplot is created with
 // no values.
--- a/.ci/check-formatting.sh
+++ b/.ci/check-formatting.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
 test -z "$(goimports -d .)"
 if [[ -n "$(gofmt -s -l .)" ]]; then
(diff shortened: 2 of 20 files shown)


sbinet:
@kortschak, do you understand why Go-1.19's test fails with `Process completed with exit code 1.` while there are no error being reported by `go test ./...` ?

(also: could you please remove the expected flag for `1.17.x` ? thanks)

kortschak:
I have no idea why that is failing. It passes locally.

kortschak:
Maybe try running the tests if `-v` to see if that gives a hint. Otherwise, bisecting the packages to see which package is causing the problem might help.

sbinet:
aha! that was actually `test -z "$(gofmt -d .)"` (GitHub's CI UI was greatly misleading...)

sbinet:
PTAL

sbinet:
I guess it doesn't handle well shell builtin functions ?
(anyways, there was also a bug in `.ci/check-formatting.sh` which should have made the initial error easier to spot)

plotter/boxplot.go
```diff
@@ -87,7 +87,7 @@ type BoxPlot struct {
 // NewBoxPlot returns a new BoxPlot that represents
 // the distribution of the given values.  The style of
 // the box plot is that used for Tukey's schematic
-// plots is ``Exploratory Data Analysis.''
+// plots is “Exploratory Data Analysis.”
```
kortschak, line 90:
```suggestion
// plots in “Exploratory Data Analysis.”
```

kortschak (commented):
There are cases where we are using \<pre\> blocks to do lists. I think those can be done with the godoc mark up now.

sbinet:
PTAL

kortschak (approved):

sbinet:
(force-push is only to clean up git history, no code change)

PR #753, opened 2022-08-19
plotter: add special handling for image.Uniform

### LINKED ISSUE #752: plotter: NewImage does not handle image.Uniform
creating a `plotter.Image` from an `image.Uniform` results in `gonum/plot` taking a very veeeery long time to produce the plot.
(that's because the `image.Uniform` has very large dimensions (`1e9`))

`vg/vgimg.Canvas.DrawImage` (and possibly other `vg.Canvas` implementations as well) will in turn fail to handle that `image.Uniform` (and "choke" on the very large dimensions).

sbinet:
Fixes #752.

Please take a look.


plotter/image.go
```diff
@@ -27,6 +28,11 @@ type Image struct {
 // (xmin, ymin) and (xmax, ymax) points given in the data space.
 // The img will be scaled to fit inside the rectangle.
 func NewImage(img image.Image, xmin, ymin, xmax, ymax float64) *Image {
+	if src, ok := img.(*image.Uniform); ok {
+		uni := image.NewRGBA(image.Rect(0, 0, int(xmax-xmin+0.5), int(ymax-ymin+0.5)))
+		draw.Draw(uni, uni.Bounds(), src, image.Point{}, draw.Src)
+		img = uni
+	}
```
kortschak, line 35:
```suggestion
	if src, ok := img.(*image.Uniform); ok {
		img = cropped{src, image.Rect(0, 0, int(xmax-xmin+0.5), int(ymax-ymin+0.5))}
	}
```
with this after the `NewImage` body means we don't need to do any copying,
```
type cropped struct {
	*image.Uniform
	rect image.Rectangle
}

func (c cropped) Bounds() image.Rectangle { return c.rect }
```
It also means the confusing rename goes away.

plotter/image.go
```diff
@@ -13,6 +13,16 @@ import (
 	"gonum.org/v1/plot/vg/draw"
 )
 
+// uniform is a cropped uniform image.
```
kortschak, line 16: I'd usually put helpers after their use; this isn't a focus, just an incidental item.

kortschak (approved):

kortschak (approved):

PR #760, opened 2023-02-21
all: bump to Go-1.20, drop Go-1.18

sbinet:
Please take a look.


sbinet:
PTAL.

@kortschak it seems the `1.18.x` builds for darwin and linux are still referenced (and marked required.)

kortschak (approved):

PR #762, opened 2023-04-04
plotter: implement GlyphBoxer for plotter.Line

### LINKED ISSUE #761: plot: Plot renders outside of image bounds

Just trying to make a simple line plot. However, depending on the axis limits, the rendered plot does not fit onto the image and exceeds its bounds.

### What did you do?

Minimal reproducing example:

```go
package main

import (
	"math/rand"

	"gonum.org/v1/plot"
	"gonum.org/v1/plot/plotter"
)

func main() {
	p := plot.New()

	p.Legend = plot.NewLegend()

	length := 70
	xy := make(plotter.XYs, length)
	for i := 0; i < length; i++ {
		xy[i] = plotter.XY{X: float64(i), Y: rand.Float64()}
	}

	lines, err := plotter.NewLine(xy)
	if err != nil {
		panic(err)
	}

	p.Add(lines)
	p.Legend.Add("Legend entry", lines)

	p.Save(600, 400, "test.png")
}
```

The plot nicely fills the image.

### What actually happened?

The plot is partially outside the image bounds:

test

go version go1.20 windows/amd64
plot vrsion v0.12.0

Yes.

sbinet:
Fixes #761.

Please take a look.


sbinet:
one existential question: should we create one gigantic glyphbox (or rather 4, one around each of the `xmin`, `xmax`, `ymin`, and `ymax` points), or one for each of line's points (as is done in this PR) ?

plotter/testdata/step_golden.png
kortschak, line 1: It's worth taking a look at how this affects this file when zoomed in.

plotter/line.go
```diff
@@ -144,6 +144,32 @@ func (pts *Line) DataRange() (xmin, xmax, ymin, ymax float64) {
...
+	bs := make([]plot.GlyphBox, pts.XYs.Len())
+	for i := range bs {
+		x, y := pts.XY(i)
+		bs[i] = plot.GlyphBox{
+			X:         plt.X.Norm(x),
+			Y:         plt.Y.Norm(y),
+			Rectangle: rect,
+		}
+	}
+	return bs
```
kortschak, line 170: What about this?
```suggestion
	r := 0.5 * pts.LineStyle.Width
	rect := vg.Rectangle{
		Min: vg.Point{
			X: -r,
			Y: -r,
		},
		Max: vg.Point{
			X: +r,
			Y: +r,
		},
	}
	xmin, xmax, ymin, ymax := pts.DataRange()
	return []plot.GlyphBox{
		{X: plt.X.Norm(xmin), Y: plt.Y.Norm(ymin), Rectangle: rect},
		{X: plt.X.Norm(xmax), Y: plt.Y.Norm(ymax), Rectangle: rect},
	}
```

kortschak:
I think what you have is probably the best approach. I am concerned about the impact on the step plot though. It does weird things to the alignment in ways that I would not expect.

plotter/testdata/step_golden.png
sbinet: it seems like, for the green "pre-" step, the outer limit of the steps is now the outer shell of the black line.

sbinet: I think I've convinced myself it's coming from an hysteresis effect stemming from the original width of the (vertical) line.
it only shows for vertical lines because, in the example, the `p.Y.{Min,Max}` are fixed and the glyph boxes won't change the data view range in `y` (as `ymax=20` is quite larger than the biggest `y` value).

the effect is then coming from clipping the x/y-ranges (via `padX` in `plot.go`) + (probably) some floating point rounding errors.

sbinet: _ping_ ?

kortschak: Yeah, OK.

kortschak (approved):

PR #764, opened 2023-04-27
Gio up

sbinet:
Please take a look.


sbinet:
this isn't the final answer: the new Gio API doesn't lend itself very nicely with our `Push/Pop` semantics.
see [1] for more details.

(but we don't use explicitly `Push/Pop` in the `vg/vggio` tests. I'd expect `plot/plotter.Field.Plot` to fail)

also, it seems the latest Gio version has a somewhat different kerning algorithm.

[1]: https://lists.sr.ht/~eliasnaur/gio/%3CCS7L02OIS1NL.1YSO72ZCFPMPE%40clrinfopc42%3E

kortschak (approved):

sbinet:
this needs https://lists.sr.ht/~eliasnaur/gio-patches/patches/40703

I haven't understood (yet?) why the embedded image is so blurry.

sbinet:
so the blur comes from a different filtering algorithm used deep inside Gio.
I guess we'll have to do the rescaling by hand CPU-side, with `image/draw`.

kortschak (approved):
TBH (maybe my eyes are too blurry already) I can't really see the difference.

PR #780, opened 2024-06-12
all: remove descent padding from horizontal axis

### LINKED ISSUE #779: Extra X-axis padding when there is a label

Create a graph with 0-padded axis and axis names

### What did you do?
```
package main

import (
	"gonum.org/v1/plot"
	"gonum.org/v1/plot/plotter"
)

type (
	benchmarkSequence struct {
	}
)

func main() {
	p := plot.New()
	p.X.Padding = 0
	p.Y.Padding = 0
	p.X.Label.Text = "P (Passingrate)" // comment this line to get expected result

	line, err := plotter.NewLine(benchmarkSequence{})
	if err != nil {
		panic(err)
	}
	p.Add(line)
	err = p.Save(500, 500, "plotLogo.png")
	if err != nil {
		panic(err)
	}
}

// Implement plot.XYer for benchmarkSequence
func (s benchmarkSequence) Len() int {
	return 99
}

func (s benchmarkSequence) XY(i int) (float64, float64) {
	return float64(i) / 100, float64(i) / 100
}
```
plotLogo

### What actually happened?
plotLogo

I just fetched gonum 0.13, and im using go 1.22.3 (linux/amd64)

Did not try, but there does not seem to be any real changes in master

kortschak:
Please take a look.

Fixes #779


sbinet (approved):

kortschak:
There's a bunch of stuff that I'm working through fixing because we were assuming all testing on amd64.

kortschak:
@sbinet vg/vggio.TestCanvas appears to be failing in an OS version-dependent manner.

kortschak:
Now it's only appveyor. I think I'm OK with that.

palette/moreland/convert.go
```diff
@@ -17,9 +17,9 @@ type rgb struct {
 // cieXYZ returns a CIE XYZ color representation of the receiver.
 func (c rgb) cieXYZ() cieXYZ {
 	return cieXYZ{
-		X: 0.4124*c.R + 0.3576*c.G + 0.1805*c.B,
-		Y: 0.2126*c.R + 0.7152*c.G + 0.0722*c.B,
-		Z: 0.0193*c.R + 0.1192*c.G + 0.9505*c.B,
+		X: float64(0.4124*c.R) + float64(0.3576*c.G) + float64(.1805*c.B),
```
vladimir-ch, line 20: Maybe add back the omitted leading 0.

kortschak: The joy of working remotely on a rpi.

sbinet (approved):

PR #784, opened 2024-10-22
all: bump Go-1.23, drop Go-1.21

sbinet:
Please take a look.


sbinet:
the Windows fail is known (and we don't care much about it).
perhaps it's time to drop appveyor and test Windows on GH ? (but let's try to do that in another PR)

kortschak (approved):

PR #796, opened 2025-03-21
all: use built-in min and max functions

Juneezee:
We can use the built-in `min` and `max` functions since Go 1.21.

Reference: https://go.dev/ref/spec#Min_and_max

### DIFF
--- a/axis.go
+++ b/axis.go
@@ -490,14 +490,14 @@ type DefaultTicks struct{}
 var _ Ticker = DefaultTicks{}
 
 // Ticks returns Ticks in the specified range.
-func (DefaultTicks) Ticks(minTick, maxTick float64) []Tick {
-	if maxTick <= minTick {
+func (DefaultTicks) Ticks(min, max float64) []Tick {
+	if max <= min {
 		panic("illegal range")
 	}
 
 	const suggestedTicks = 3
 
-	labels, step, q, mag := talbotLinHanrahan(minTick, maxTick, suggestedTicks, withinData, nil, nil, nil)
+	labels, step, q, mag := talbotLinHanrahan(min, max, suggestedTicks, withinData, nil, nil, nil)
 	majorDelta := step * math.Pow10(mag)
 	if q == 0 {
 		// Simple fall back was chosen, so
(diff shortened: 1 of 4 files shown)


kortschak (commented):
I'm not convinced this is a win.

Juneezee:
> I'm not convinced this is a win.

Thank you for the review. I'm eager to make this PR beneficial for the project. May I know on what aspects you find it problematic?

kortschak:
In order to use the builtin, we need to reduce the simplicity of the documentation by changing the return parameter names.

axis.go
```diff
@@ -490,14 +490,14 @@ type DefaultTicks struct{}
 var _ Ticker = DefaultTicks{}
 
 // Ticks returns Ticks in the specified range.
-func (DefaultTicks) Ticks(min, max float64) []Tick {
-	if max <= min {
+func (DefaultTicks) Ticks(minTick, maxTick float64) []Tick {
```
Juneezee, line 493: > In order to use the builtin, we need to reduce the simplicity of the documentation by changing the return parameter names.

I assume you are referring to the change here? I’d love to understand your perspective better because, from my view, this change doesn’t reduce simplicity, it only modifies the parameter names, which are internal details and do not impact the API or how users interact with the function. `DefaultTicks` still fulfils the `Ticker` interface.

If there's a particular case where you think it might cause confusion, I'd be happy to explore alternatives.

kortschak: > I assume you are referring to the change here?

Yes.

> I’d love to understand your perspective better because, from my view, this change doesn’t reduce simplicity, it only modifies the parameter names,

This is the issue. It changes the names of the returned values from the consistently used names. This is a reduction in simplicity.

*which are internal details* and do not impact the API or how users interact with the function.

This is definitively not true. These names are published in the public godoc and so impact on the client user's interaction with the API via that surface.

This change is not necessary. Let's drop it please.

Juneezee: Thanks for the feedback. I have reverted the change in this file in commit e0aeedfeb94ca850e3aefdc62dbeff82fa87b113.

sbinet (approved):
LGTM

(the gain is somewhat marginal, but I guess it's in line with the "modernize" movement)

kortschak (approved):

Issue #798, opened 2025-03-21
palette/moreland: possible panic for Palette method


Juneezee:
Minimal reproducible example:

```go
package plotter_test

import (
	"testing"

	"gonum.org/v1/plot/palette/moreland"
)

func TestPalettePanic(t *testing.T) {
	minVal := 0.3402859786606234
	maxVal := 15.322841335211892
	n := 15

	t.Run("moreland.SmoothBlueRed panic", func(t *testing.T) {
		colors := moreland.SmoothBlueRed()
		colors.SetMin(minVal)
		colors.SetMax(maxVal)
		colors.Palette(n)
	})

	t.Run("moreland.Kindlmann panic", func(t *testing.T) {
		colors := moreland.Kindlmann()
		colors.SetMin(minVal)
		colors.SetMax(maxVal)
		colors.Palette(n)
	})
}
```

**`moreland.SmoothBlueRed`**:

```
=== RUN   TestPalettePanic
=== RUN   TestPalettePanic/moreland.SmoothBlueRed_panic
--- FAIL: TestPalettePanic/moreland.SmoothBlueRed_panic (0.00s)

--- FAIL: TestPalettePanic (0.00s)
panic: palette: specified value > maximum [recovered]
	panic: palette: specified value > maximum

goroutine 7 [running]:
testing.tRunner.func1.2({0x6d8d20, 0xe61110})
	/usr/local/go/src/testing/testing.go:1734 +0x21c
testing.tRunner.func1()
	/usr/local/go/src/testing/testing.go:1737 +0x35e
panic({0x6d8d20?, 0xe61110?})
	/usr/local/go/src/runtime/panic.go:792 +0x132
gonum.org/v1/plot/palette/moreland.smoothDiverging.Palette({{0x4054000000000000, 0x3ff147ae147ae148, 0xbff199999999999a}, {0x4054000000000000, 0x3ff147ae147ae148, 0x3fe0000000000000}, 0x4056000000000000, 0x3ff0000000000000, 0x3fd5c73ed7689f70, 0x402ea54b75a11069, ...}, ...)
	/home/jun/Desktop/github/waiting-for-pr/plot/palette/moreland/smooth.go:180 +0x1ab
gonum.org/v1/plot/plotter_test.TestPalettePanic.func1(0xc0002061c0?)
	/home/jun/Desktop/github/waiting-for-pr/plot/plotter/panic_test.go:18 +0x1b2
[...]

PR #799, opened 2025-03-22
palette/moreland: fix floating-point arithmetic error in Palette

Juneezee:
Fixes https://github.com/gonum/plot/issues/798.

This PR fixes a panic in the `Palette` method caused by a floating-point precision error, where the computed value `v` slightly exceeds the upper bound `l.max`.

### Problem

As shown in the above screenshot, we have the following values:

- `l.min = 0.3402859786606234`
- `l.max = 15.322841335211892`
- `n = 15`
- `delta = 1.0701825254679478`
- `i = 14`
- `v = 15.322841335211894` (note here `v` is `0.000000000000002` greater than `l.max`)

The expected calculation of `delta`:

```
delta = (l.max - l.min) / (n - 1)
      = (15.322841335211892 - 0.3402859786606234) / (15 - 1)
      = 1.0701825254679478
```

This confirms that `delta` is correct and has no floating-point error.

Next, the expected calculation of `v`:

```
v = l.min + (delta  * i)
  = 0.3402859786606234 + (1.0701825254679478 * 14)
  = 15.322841335211892
```

Ideally, `v` should be exactly equal to `l.max`. However, due to a small floating-point precision error of `0.000000000000002`, `v` becomes slightly greater than `l.max`, triggering the panic in `checkRange` function:

https://github.com/gonum/plot/blob/2815ea166a2a2c6b1f31037ebaa26616029a2929/palette/moreland/luminance.go#L116-L118

### Solution

Use the built-in `min` function to ensure `v` does not exceed `l.max`. If `v` becomes greater than `l.max` due to floating-point precision errors, it will be capped at `l.max`.

### DIFF
--- a/palette/moreland/luminance_test.go
+++ b/palette/moreland/luminance_test.go
@@ -162,7 +162,7 @@ func BenchmarkLuminance_At(b *testing.B) {
 	}
 }
 
-func Test_luminance_Palette(t *testing.T) {
+func TestIssue798Kindlmann(t *testing.T) {
 	// https://github.com/gonum/plot/issues/798
 	colors := Kindlmann()
 	colors.SetMin(0.3402859786606234)
--- a/palette/moreland/luminance_test.go
+++ b/palette/moreland/luminance_test.go
@@ -161,3 +161,11 @@ func BenchmarkLuminance_At(b *testing.B) {
 		})
 	}
 }
+
+func Test_luminance_Palette(t *testing.T) {
+	// https://github.com/gonum/plot/issues/798
+	colors := Kindlmann()
+	colors.SetMin(0.3402859786606234)
+	colors.SetMax(15.322841335211892)
+	colors.Palette(15)
+}
--- a/palette/moreland/luminance_test.go
+++ b/palette/moreland/luminance_test.go
@@ -162,10 +162,46 @@ func BenchmarkLuminance_At(b *testing.B) {
 	}
 }
 
+// See https://github.com/gonum/plot/issues/798
 func TestIssue798Kindlmann(t *testing.T) {
-	// https://github.com/gonum/plot/issues/798
-	colors := Kindlmann()
-	colors.SetMin(0.3402859786606234)
-	colors.SetMax(15.322841335211892)
-	colors.Palette(15)
+	for _, test := range []struct {
+		n        int
+		min, max float64
+	}{
+		0: {n: 2, min: 0, max: 1},
+		1: {n: 15, min: 0.3402859786606234, max: 15.322841335211892},
+	} {
+		t.Run("", func(t *testing.T) {
+			defer func() {
+				r := recover()
+				if r != nil {
+					t.Errorf("unexpected panic with n=%d min=%f max=%f: %v", test.n, test.min, test.max, r)
+				}
+			}()
+			colors := Kindlmann()
+			colors.SetMin(test.min)
+			colors.SetMax(test.max)
+			col := colors.Palette(test.n).Colors()
+			min, err := colors.At(test.min)
+			if err != nil {
+				t.Fatalf("unexpected error calling colors.At(min): %v", err)
+			}
+			if !sameColor(min, col[0]) {
+				t.Errorf("unexpected min color %#v != %#v", min, col[0])
+			}
+			max, err := colors.At(test.max)
+			if err != nil {
+				t.Fatalf("unexpected error calling colors.At(max): %v", err)
+			}
+			if !sameColor(max, col[len(col)-1]) {
+				t.Errorf("unexpected max color %#v != %#v", max, col[len(col)-1])
+			}
+		})
+	}
+}
+
+func sameColor(a, b color.Color) bool {
+	ar, ag, ab, aa := a.RGBA()
+	br, bg, bb, ba := b.RGBA()
+	return ar == br && ag == bg && ab == bb && aa == ba
 }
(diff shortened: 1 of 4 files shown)


palette/moreland/luminance_test.go
```diff
@@ -161,3 +161,11 @@ func BenchmarkLuminance_At(b *testing.B) {
 		})
 	}
 }
+
+func Test_luminance_Palette(t *testing.T) {
```
sbinet, line 165:
```suggestion
func TestIssue798Kindlmann(t *testing.T) {
```

palette/moreland/smooth_test.go
```diff
@@ -246,3 +246,11 @@ func similar(a, b color.Color, tolerance float64) bool {
 	}
 	return true
 }
+
+func Test_smoothDiverging_Palette(t *testing.T) {
```
sbinet, line 250:
```suggestion
func TestIssue798SmoothBlueRed(t *testing.T) {
```

sbinet (commented):
Thanks for the PR.
I just have minor requests.

See below:

palette/moreland/luminance_test.go
Juneezee: Updated in 55662b36b4613f9934453412090318e259a40308.

palette/moreland/smooth_test.go
Juneezee: Updated in 55662b36b4613f9934453412090318e259a40308.

sbinet (dismissed):
LGTM

thanks

palette/moreland/luminance_test.go
```diff
@@ -161,3 +161,11 @@ func BenchmarkLuminance_At(b *testing.B) {
 		})
 	}
 }
+
+func TestIssue798Kindlmann(t *testing.T) {
+	// https://github.com/gonum/plot/issues/798
```
kortschak, line 166:
```suggestion
// See https://github.com/gonum/plot/issues/798
func TestIssue798Kindlmann(t *testing.T) {
```

palette/moreland/smooth.go
```diff
@@ -173,7 +173,7 @@ func (p smoothDiverging) Palette(n int) palette.Palette {
 	delta := (p.max - p.min) / float64(n-1)
 	c := make([]color.Color, n)
 	for i := range c {
-		v := p.min + float64(delta*float64(i))
+		v := min(p.min+float64(delta*float64(i)), p.max)
```
kortschak, line 176: Same as above.

palette/moreland/luminance_test.go
```diff
@@ -161,3 +161,11 @@ func BenchmarkLuminance_At(b *testing.B) {
...
+	colors := Kindlmann()
+	colors.SetMin(0.3402859786606234)
+	colors.SetMax(15.322841335211892)
+	colors.Palette(15)
```
kortschak, line 170: This should not just panic on failure, we should capture and recover to present a test failure. I think we should also confirm that the min and max palette values are what we expect.

```
// See https://github.com/gonum/plot/issues/798
func TestIssue798Kindlmann(t *testing.T) {
	for _, test := range []struct {
		n        int
		min, max float64
	}{
		0: {n: 2, min: 0, max: 1},
		1: {n: 15, min: 0.3402859786606234, max: 15.322841335211892},
	} {
		t.Run("", func(t *testing.T) {
			defer func() {
				r := recover()
				if r != nil {
					t.Errorf("unexpected panic with n=%d min=%f max=%f: %v", test.n, test.min, test.max, r)
				}
			}()
			colors := Kindlmann()
			colors.SetMin(test.min)
			colors.SetMax(test.max)
			col := colors.Palette(test.n).Colors()
			min, err := colors.At(test.min)
			if err != nil {
				t.Fatalf("unexpected error calling colors.At(min): %v", err)
			}
			if !sameColor(min, col[0]) {
				t.Errorf("unexpected min color %#v != %#v", min, col[0])
			}
			max, err := colors.At(test.max)
			if err != nil {
				t.Fatalf("unexpected error calling colors.At(max): %v", err)
			}
			if !sameColor(max, col[len(col)-1]) {
				t.Errorf("unexpected max color %#v != %#v", max, col[len(col)-1])
			}
		})
	}
}

func sameColor(a, b color.Color) bool {
	ar, ag, ab, aa := a.RGBA()
	br, bg, bb, ba := b.RGBA()
	return ar == br && ag == bg && ab == bb && aa == ba
}
```

kortschak (changes requested):

palette/moreland/luminance_test.go
Juneezee: Updated in commit 1434e666b2dbe032a84df62e813be71c9f6b8831.

kortschak (approved):
Thanks.

Please send a change to gonum/gonum with the commit message "A+C: add Eng Zer Jun" (please correct this if I got your name wrong) adding yourself to AUTHORS and CONTRIBUTORS.

Juneezee:
> Thanks.
> 
> Please send a change to gonum/gonum with the commit message "A+C: add Eng Zer Jun" (please correct this if I got your name wrong) adding yourself to AUTHORS and CONTRIBUTORS.

@kortschak My name is already in the AUTHORS and CONTRIBUTORS file :blush: 

- https://github.com/gonum/gonum/blob/7826ba4358a76646a8bd1f852201c4fed2c81c38/AUTHORS#L43
- https://github.com/gonum/gonum/blob/7826ba4358a76646a8bd1f852201c4fed2c81c38/CONTRIBUTORS#L50

kortschak:
@Juneezee Ah, thanks. You show up as first time contributor here, but this is because the repo is different. Thanks for the change.

Issue #797, opened 2025-03-21
Replace `golang.org/x/exp/rand` with `math/rand/v2`


Juneezee:
This is similar to https://github.com/gonum/gonum/issues/1906. We have already upgraded our minimum Go version to 1.23 in PR https://github.com/gonum/plot/pull/790. `golang.org/x/exp/rand` has been deprecated is scheduled for removal (https://github.com/golang/exp/commit/f9890c6ad9f380fd3cdb66a33843f522d4790e03).

Our usage of `golang.org/x/exp/rand` is limited to tests and examples, so replacing it should not introduce any breaking changes.

```less
$ grep --recursive --line-number --include='*.go' 'golang.org/x/exp'

plotter/contour_example_test.go:13:	"golang.org/x/exp/rand"
plotter/boxplot_example_test.go:12:	"golang.org/x/exp/rand"
plotter/barchart_example_test.go:12:	"golang.org/x/exp/rand"
plotter/bubbles_example_test.go:12:	"golang.org/x/exp/rand"
plotter/histogram_example_test.go:12:	"golang.org/x/exp/rand"
plotter/contour_test.go:16:	"golang.org/x/exp/rand"
plotter/filledLine_test.go:12:	"golang.org/x/exp/rand"
plotter/step_example_test.go:11:	"golang.org/x/exp/rand"
plotter/errbars_example_test.go:10:	"golang.org/x/exp/rand"
plotter/quartile_example_test.go:10:	"golang.org/x/exp/rand"
plotter/scatter_example_test.go:11:	"golang.org/x/exp/rand"
plotter/filledLine_example_test.go:11:	"golang.org/x/exp/rand"
plotter/scatterColor_example_test.go:13:	"golang.org/x/exp/rand"
plotter/timeseries_example_test.go:12:	"golang.org/x/exp/rand"
gob/gob_test.go:14:	"golang.org/x/exp/rand"
plotutil/main.go:11:	"golang.org/x/exp/rand"
plotutil/errorpoints_example_test.go:8:	"golang.org/x/exp/rand"
palette/moreland/luminance_test.go:12:	"golang.org/x/exp/rand"
palette/moreland/smooth_test.go:13:	"golang.org/x/exp/rand"
```

If this proposal is accepted, I am happy to open a PR.

sbinet:
yes, please.
(but perhaps a fix for #798 should first be submitted ? as #797 relies on it to correctly run)

Juneezee:
[...]

PR #801, opened 2025-03-23
all: migrate to math/rand/v2

Juneezee:
Closes https://github.com/gonum/plot/issues/797.

Please take a look.


Juneezee (commented):
For commit 20fe21897a144fe364ac1274ca390770f4144659, although the differences between the AMD64 and ARM64 golden images for the Contour and Filled Line tests are not distinguishable to the human eye, there appears to be a slight difference in floating-point calculations between AMD64 and ARM64.

https://github.com/golang/go/issues/36536

plotter/scatterColor_example_test.go
```diff
@@ -75,7 +74,8 @@ func ExampleScatter_color() {
 		_, _, z := scatterData.XYZ(i)
 		d := (z - minZ) / (maxZ - minZ)
 		rng := maxZ - minZ
-		k := d*rng + minZ
+		// Possible floating point error, similar to https://github.com/gonum/plot/issues/798,
+		k := min(d*rng+minZ, maxZ)
```
sbinet, line 78: indeed, it seems to possibly be prey to the same bug than #798.
I'll leave it to @kortschak to decide whether he'd prefer to deal with it in the same way.
(I'm fine with clamping the value like that)

sbinet (dismissed):

plotter/boxplot_example_test.go
```diff
@@ -8,16 +8,15 @@ import (
 	"fmt"
 	"image/color"
 	"log"
-
-	"golang.org/x/exp/rand"
+	"math/rand/v2"
```
kortschak, line 11: This should be in the std stanza, throughout.

plotter/scatterColor_example_test.go
kortschak: We don't need the comment, at least, we don't need a reference to the issue. There can be an explanatory not that this is required due to float error.

However, I'd put it in the call, `c, err := colors.At(min(k, colors.Max()))` and leave the original `k := d*rng + minZ`.

kortschak (changes requested):

sbinet (dismissed):
LGTM

plotter/scatterColor_example_test.go
```diff
@@ -75,8 +74,9 @@ func ExampleScatter_color() {
 		_, _, z := scatterData.XYZ(i)
 		d := (z - minZ) / (maxZ - minZ)
 		rng := maxZ - minZ
-		k := d*rng + minZ
-		c, err := colors.At(k)
+		k := min(d*rng+minZ, maxZ)
+		// Clamp k to avoid potential overflow due to floating point error
```
kortschak, line 78:
```suggestion
		// Clamp k to avoid potential overflow due to floating point error.
```

plotter/scatterColor_example_test.go
kortschak: This is not done; use `k := d*rng + minZ` rather than calling `min` twice.

kortschak (approved):
Thanks

